PR #2, opened 2020-06-19
Set hostname to user configurable device name if possible

nirev:
Attempt to fix https://github.com/tailscale/tailscale/issues/482

According to [1] it may not be possible to get the device nickname.
In this PR I tailscale tries to set hostname in the following order:
- Settings.System "bluetooth_name"
- Settings.Secure "bluetooth_name"
- Settings.System "device_name"
- Manufactor + Model

The first one available wins.
In my device, a Samsung S9, this is what I get from logcat:
```
06-19 19:27:00.517 28825 28937 D com.tailscale.ipn.App: Device name from System Bluetooth: null
06-19 19:27:00.517 28825 28937 D com.tailscale.ipn.App: Device name from Secure Bluetooth: destruction
06-19 19:27:00.517 28825 28937 D com.tailscale.ipn.App: Device name from System Device: null
```

I left the debug messages as I thought it would be useful, given how every device is a bit different.
Let me know what you think.

[1] https://medium.com/capital-one-tech/how-to-get-an-android-device-nickname-d5eab12f4ced

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.java
+++ b/android/src/main/java/com/tailscale/ipn/App.java
@@ -115,7 +115,8 @@ public class App extends Application {
 
 		if (!isEmpty(nameFromSystemBluetooth)) return nameFromSystemBluetooth;
 		if (!isEmpty(nameFromSecureBluetooth)) return nameFromSecureBluetooth;
-		return nameFromSystemDevice;
+		if (!isEmpty(nameFromSystemDevice)) return nameFromSystemDevice;
+		return null;
 	}
 
 	private static boolean isEmpty(String str) {
--- a/android/src/main/java/com/tailscale/ipn/App.java
+++ b/android/src/main/java/com/tailscale/ipn/App.java
@@ -86,7 +86,7 @@ public class App extends Application {
 
 	String getHostname() {
 		String userConfiguredDeviceName = getUserConfiguredDeviceName();
-		if (notEmpty(userConfiguredDeviceName)) return userConfiguredDeviceName;
+		if (!isEmpty(userConfiguredDeviceName)) return userConfiguredDeviceName;
 
 		return getModelName();
 	}
@@ -113,13 +113,13 @@ public class App extends Application {
 		Log.d("com.tailscale.ipn.App", "Device name from Secure Bluetooth: " + nameFromSecureBluetooth);
 		Log.d("com.tailscale.ipn.App", "Device name from System Device: " + nameFromSystemDevice);
 
-		if (notEmpty(nameFromSystemBluetooth)) return nameFromSystemBluetooth;
-		if (notEmpty(nameFromSecureBluetooth)) return nameFromSecureBluetooth;
+		if (!isEmpty(nameFromSystemBluetooth)) return nameFromSystemBluetooth;
+		if (!isEmpty(nameFromSecureBluetooth)) return nameFromSecureBluetooth;
 		return nameFromSystemDevice;
 	}
 
-	private static boolean notEmpty(String str) {
-		return !(str == null || str.length() == 0);
+	private static boolean isEmpty(String str) {
+		return str == null || str.length() == 0;
 	}
 
 	// Tracklifecycle adds a Peer fragment for tracking the Activity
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -94,6 +103,25 @@ String getHostname() {
...
+		Log.d("com.tailscale.ipn.App", "Device name from System Bluetooth: " + nameFromSystemBluetooth);
+		Log.d("com.tailscale.ipn.App", "Device name from Secure Bluetooth: " + nameFromSecureBluetooth);
+		Log.d("com.tailscale.ipn.App", "Device name from System Device: " + nameFromSystemDevice);
+
+		if (notEmpty(nameFromSystemBluetooth)) return nameFromSystemBluetooth;
+		if (notEmpty(nameFromSecureBluetooth)) return nameFromSecureBluetooth;
+		return nameFromSystemDevice;
+	}
+
+	private static boolean notEmpty(String str) {
```
bradfitz, line 121: negatively named boolean functions inevitably get double negatived and read confusingly later.

better to just name it `emptyStr` and then write `!emptyStr` at the call sites.

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -94,6 +103,25 @@ String getHostname() {
...
+		String nameFromSecureBluetooth = Settings.Secure.getString(getContentResolver(), "bluetooth_name");
+		String nameFromSystemDevice = Settings.Secure.getString(getContentResolver(), "device_name");
+
+		Log.d("com.tailscale.ipn.App", "Device name from System Bluetooth: " + nameFromSystemBluetooth);
+		Log.d("com.tailscale.ipn.App", "Device name from Secure Bluetooth: " + nameFromSecureBluetooth);
+		Log.d("com.tailscale.ipn.App", "Device name from System Device: " + nameFromSystemDevice);
+
+		if (notEmpty(nameFromSystemBluetooth)) return nameFromSystemBluetooth;
+		if (notEmpty(nameFromSecureBluetooth)) return nameFromSecureBluetooth;
+		return nameFromSystemDevice;
```
bradfitz, line 118: You check all the rest for being either null or length 0, but not this one. Which means the func overall can't make any promises on what it does on not finding a match.

I'd use the same `!emptyStr` check for all three, then return `null` on failure to match, and document that on `getUserConfiguredDeviceName`

android/src/main/java/com/tailscale/ipn/App.java
nirev: absolutely right. my bad.

nirev:
Thank you, @bradfitz! 
It was super easy to contribute to and test it by using the provided Makefile and Dockerfile :)

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -94,6 +103,27 @@ String getHostname() {
...
+	// get user defined nickname from Settings
+	// returns null if not available
+	private String getUserConfiguredDeviceName() {
+		String nameFromSystemBluetooth = Settings.System.getString(getContentResolver(), "bluetooth_name");
+		String nameFromSecureBluetooth = Settings.Secure.getString(getContentResolver(), "bluetooth_name");
+		String nameFromSystemDevice = Settings.Secure.getString(getContentResolver(), "device_name");
+
+		Log.d("com.tailscale.ipn.App", "Device name from System Bluetooth: " + nameFromSystemBluetooth);
+		Log.d("com.tailscale.ipn.App", "Device name from Secure Bluetooth: " + nameFromSecureBluetooth);
+		Log.d("com.tailscale.ipn.App", "Device name from System Device: " + nameFromSystemDevice);
```
eliasnaur, line 115: Let's drop the logging until they're needed. Also, only the Go side of the app is logging right now, avoiding the need for updating the log tag in multiple places.

nirev: Sounds good. Removed!

eliasnaur:
Thanks! Merged.

PR #3, opened 2020-10-31
version,Makefile: implement new Tailscale versioning scheme

eliasnaur:
The strategy is to ask `go list` for the tailscale.com module version, and then
use that as a tag in a `git ls-remote` query.

If the module version is a pseudo-version such as
"v1.1.1-0.20201030135043-eab6e9ea4e45", use the abbreviated commit directly
(git ls-remote only list remote refs, not commits).

Fixes tailscale/tailscale#883

Signed-off-by: Elias Naur <mail@eliasnaur.com>


eliasnaur:
@bradfitz WDYT? I'm still not thrilled that we have to basically second-guess the Go module system to dig out the tailcale.com module commit in a way that can, at least in theory, differ from the version checked out by go get or the module proxy.

PR #5, opened 2020-12-20
Exclude com.tailscale.ipn from its own VPN.

danderson:
This means that the Tailscale app's traffic will never use
the VPN that it sets up, which avoids routing loops in
scenarios like publishing a default route over Tailscale.

Signed-off-by: David Anderson <dave@natulte.net>

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/IPNService.java
+++ b/android/src/main/java/com/tailscale/ipn/IPNService.java
@@ -8,6 +8,7 @@ import android.os.Build;
 import android.app.PendingIntent;
 import android.app.NotificationChannel;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.net.VpnService;
 import android.system.OsConstants;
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/IPNService.java
```diff
@@ -61,6 +62,12 @@ private PendingIntent configIntent() {
 			.setConfigureIntent(configIntent())
 			.allowFamily(OsConstants.AF_INET)
 			.allowFamily(OsConstants.AF_INET6);
+		try {
+			b.addDisallowedApplication("com.tailscale.ipn");
```
bradfitz, line 66: I'd rather we look this up from ourselves rather than hard code it where it might fall out of sync in a future fork or rename.

bradfitz: BuildConfig.APPLICATION_ID

bradfitz: @eliasnaur, can you finish this change up and submit it? @danderson is out until the new year.

eliasnaur:
Merged as https://github.com/tailscale/tailscale-android/commit/8daee9c431264c7780a39802e6213a1b13cab7f8.

PR #7, opened 2021-03-04
Implement Tailscale 1.6 relay setting

eliasnaur:
As mentioned in https://github.com/tailscale/tailscale/issues/1401 I haven't got relaying to actually work yet, but the UI should be complete according to @rosszurowski's design.


rosszurowski:
Parsing through the code, it seems great. I don't have my devices set up to build the Android app, but I have time with @soniaappasamy later today to review. I'll leave any comments then.

cmd/tailscale/ui.go
```diff
@@ -338,6 +404,51 @@ func (d *Dismiss) Dismissed(gtx layout.Context) bool {
...
+			Top:    unit.Dp(12),
+			Bottom: unit.Dp(12),
+			Right:  unit.Dp(24),
+			Left:   unit.Dp(24),
+		}.Layout(gtx, func(gtx C) D {
+			return layout.Flex{Alignment: layout.Middle}.Layout(gtx,
+				layout.Flexed(1, func(gtx C) D {
+					return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+						layout.Rigid(func(gtx C) D {
+							lbl := material.Caption(ui.theme, text)
```
rosszurowski, line 433: Lets use the same type size at the IP addresses for both of these labels. It looks like they're a new, smaller type size than we use elsewhere.

cmd/tailscale/ui.go
```diff
@@ -509,86 +620,191 @@ func (ui *UI) menuClicked(btn *widget.Clickable) bool {
...
-						children = append(children, layout.Rigid(func(gtx C) D {
-							return material.Clickable(gtx, it.btn, func(gtx C) D {
-								return layout.UniformInset(unit.Dp(16)).Layout(gtx, func(gtx C) D {
-									gtx.Constraints.Min.X = minWidth
-									dims := material.Body1(ui.theme, it.title).Layout(gtx)
-									if w := dims.Size.X; w > maxWidth {
-										maxWidth = w
-									}
-									return dims
+								return material.Clickable(gtx, &d.done, func(gtx C) D {
```
rosszurowski, line 698: Let's update this dialog so tap an option immediately selects it. This also means we can do away with the "Done" button. It looks like Android handles overflow scrolling on these dialogs well by default, but if we need space to close the dialog we can add an explicit "Cancel" button too.

Here's a reference from the stock messages app on my Motorola G5.

https://user-images.githubusercontent.com/303731/110039551-68226c80-7d0f-11eb-93fd-aec5859d28cd.mp4

Open to other approaches if you think there's a more natural Android pattern here.

cmd/tailscale/ui.go
```diff
@@ -256,18 +304,24 @@ func (ui *UI) layout(gtx layout.Context, sysIns system.Insets, state *clientStat
 				if netmap == nil || state.backend.State < ipn.Stopped {
 					return D{}
 				}
+				for ui.self.Clicked() {
+					events = append(events, CopyEvent{Text: localAddr})
+					ui.showCopied(gtx, localAddr)
+				}
 				return ui.layoutLocal(gtx, sysIns, localName, localAddr)
 			case 2:
+				return ui.layoutExitStatus(gtx, &state.backend)
```
rosszurowski, line 313: I noticed in my testing that switching from a router back to "None" caused this panel to stick around for a few seconds more. In one case it wouldn't close at all.

Not sure if this is an IPN state issue, but calling it out in case you have thoughts.

rosszurowski (commented):
A few small nits about the UI but generally looks good. I had trouble using the 
default route feature. I think @danderson is planning on following up in the other thread about this.

cmd/tailscale/main.go
```diff
@@ -254,6 +282,11 @@ func (a *App) runBackend() error {
 			}
 			if m := n.NetMap; m != nil {
 				state.NetworkMap = m
+				exitWasOnline := state.ExitState == ExitOnline
+				state.updateExitNodes()
+				if exitWasOnline && state.ExitState == ExitOffline && service != 0 {
+					a.pushNotify(service, "Connection Lost", "Your traffic router is no longer available. Disable traffic routing or contact your local network admin for help.")
```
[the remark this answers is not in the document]
eliasnaur: Done. Sorry for missing this one.

cmd/tailscale/ui.go
eliasnaur: Fixed. I had mistakenly forgot to update the relay state when the preference change but not the netmap. For some reason, my test device always got a netmap update immediately after the preference update.

DentonGentry:
Is this ready? We'd like this to be in the 1.6 release, and we're almost out of time.

eliasnaur:
I believe so. I need to go get tailscale@v1.6.0 whenever that's done and then release.

cmd/tailscale/main.go
```diff
@@ -249,11 +279,16 @@ func (a *App) runBackend() error {
...
 				state.NetworkMap = m
+				state.updateExitNodes()
 				a.notify(state)
 				if service != 0 {
 					alarm(a.notifyExpiry(service, m.Expiry))
 				}
 			}
+			// Notify if a previously online exit is not longer online (or missing).
+			if service != 0 && exitWasOnline && state.ExitStatus == ExitOffline {
+				a.pushNotify(service, "Connection Lost", "Your traffic relay is offline. Disable your traffic relay or contact your network admin for help.")
```
rosszurowski, line 290: > Your exit node is offline. Disable your exit node or contact your network admin for help.

cmd/tailscale/ui.go
```diff
@@ -338,6 +406,51 @@ func (d *Dismiss) Dismissed(gtx layout.Context) bool {
...
 }
 
+func (ui *UI) layoutExitStatus(gtx layout.Context, state *BackendState) layout.Dimensions {
+	var bg color.NRGBA
+	var text string
+	switch state.ExitStatus {
+	case ExitNone:
+		return D{}
+	case ExitOffline:
+		text = "Traffic relay offline"
```
rosszurowski, line 416: > Exit node offline

cmd/tailscale/ui.go
```diff
@@ -338,6 +406,51 @@ func (d *Dismiss) Dismissed(gtx layout.Context) bool {
...
+		bg = rgb(0xc65835)
+	case ExitOnline:
+		text = "Relaying traffic via"
```
rosszurowski, line 419: > Using exit node

cmd/tailscale/ui.go
```diff
@@ -509,9 +622,134 @@ func (ui *UI) menuClicked(btn *widget.Clickable) bool {
...
+				return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+					layout.Rigid(func(gtx C) D {
+						// Header.
+						return layout.Inset{
+							Top:    unit.Dp(16),
+							Right:  unit.Dp(20),
+							Left:   unit.Dp(20),
+							Bottom: unit.Dp(16),
+						}.Layout(gtx, func(gtx C) D {
+							l := material.Body1(ui.theme, "Route traffic via...")
```
rosszurowski, line 655: > Use exit node…

cmd/tailscale/ui.go
```diff
@@ -520,75 +758,31 @@ func (ui *UI) layoutMenu(gtx layout.Context, sysIns system.Insets, expiry time.T
...
-					// Second pass: layout items with equal width.
-					minWidth = maxWidth
-					return f.Layout(gtx, children...)
-				})
+			menu := &ui.menu
+			items := []menuItem{
+				{title: "Copy My IP Address", btn: &menu.copy},
+			}
+			if showExits {
+				items = append(items, menuItem{title: "Route traffic via...", btn: &menu.exits})
```
rosszurowski, line 766: > Use exit node…

rosszurowski (commented):
One small tweak: we've been flip-flopping on the term we were going to use for this feature: are they "default routers" or "traffic relays" or "gateways"? After a bit of research and conversation, we're settling on "exit node."

Mind updating this copy before making the final release? Thanks! 🙏

eliasnaur:
> Mind updating this copy before making the final release? Thanks! pray

Done!

danderson:
I tested this on my Android phone, and it works (and looks) great. I checked that both IPv4 and IPv6 relaying are working correctly, using adb shell and various test websites. https://github.com/tailscale/tailscale/issues/1486#issuecomment-797822013 records my findings, but TL;DR: everything works as expected, except that Chrome is doing slightly surprising (but not broken) behavior. So, I think this is good to submit, and then release next week when we pull the trigger on 1.6.0.

danderson (approved):

PR #8, opened 2021-04-18
go.*,cmd/tailscale: upgrade to latest tailscale.com

eliasnaur:


eliasnaur:
As expected, Android doesn't allow you to read `/etc/resolv.conf`:

`health("dns"): error: open /etc/resolv.conf: no such file or directory`

There used to be a set of system properties for reading the DNS servers, but [Android 8](https://developer.android.com/about/versions/oreo/android-8.0-changes.html#o-pri) removed them. The replacement is [ConnectivityManager.registerNetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager#registerNetworkCallback(android.net.NetworkRequest,%20android.net.ConnectivityManager.NetworkCallback,%20android.os.Handler)), but that's asynchronous and may update DNS servers over time.

@danderson, how do you suggest I plumb the Android network callbacks through to Tailscale?

danderson:
Thanks for investigating how android does DNS. It's annoying that they're hiding the base DNS config from us :/

Using the network callbacks, can we figure out which servers are supposed to be primary (e.g. if you're connected to wifi + LTE) ? From the docs I can't immediately tell if we get enough information for that.

Another alternative would be to add support for callback-based resolvers in net/dns/resolver, so we can set a default resolver that calls back into tailscale-android and does a DNS lookup via JNI, using android's DNS resolver classes. AFAICT, those classes might do the right thing, assuming the tailscale app is exempt from its own DNS settings when it does lookups... Do you know if that's true?

In any case, for right now, we needed a workaround for macOS as well, so I implemented fallback resolvers for android in 1.8, if you sync the repo again. The android OSConfigurator can safely return not-implemented for `GetBaseConfig()` while we figure out how to make things work.

danderson:
Oh, and if we do implement stuff with the network callbacks: I'm going to have to add some kind of "DNS monitor" to trigger reconfig on changes for other OSes as well, so android can use that as well. Not done yet though.

eliasnaur:
> Thanks for investigating how android does DNS. It's annoying that they're hiding the base DNS config from us :/
> 

Not hiding so much as needing to call ConnectivityManager.getAllNetworks, and then ConnectivityManager.getLinkProperties on each network, and for each network call [getDnsServers](https://developer.android.com/reference/android/net/LinkProperties#getDnsServers()). I mentioned the callbacks so we don't forget to handle network switching.

> Using the network callbacks, can we figure out which servers are supposed to be primary (e.g. if you're connected to wifi + LTE) ? From the docs I can't immediately tell if we get enough information for that.
> 

What do you mean by primary DNS server? getAllNetworks may return multiple networks, and each network have a list of DNS servers.

We may need to filter the networks and/or resolvers. For example, [this SO answer](https://stackoverflow.com/a/48973823) filter networks on them having a default route.

> Another alternative would be to add support for callback-based resolvers in net/dns/resolver, so we can set a default resolver that calls back into tailscale-android and does a DNS lookup via JNI, using android's DNS resolver classes. AFAICT, those classes might do the right thing, assuming the tailscale app is exempt from its own DNS settings when it does lookups... Do you know if that's true?
> 

I don't think this is necessary, given the above.

> In any case, for right now, we needed a workaround for macOS as well, so I implemented fallback resolvers for android in 1.8, if you sync the repo again. The android OSConfigurator can safely return not-implemented for `GetBaseConfig()` while we figure out how to make things work.

I've pushed an update to latest tailscale.com. Do you want me to implement DNS server plumbing now or wait for support DNS server change callbacks in tailscale?

apenwarr:
I can't tell if this is obsoleted by @danderson's and @DentonGentry's earlier Android work for DNS support. I'll close for now assuming there's nothing more to do here, but feel free to reopen. cc: @eliasnaur

PR #9, opened 2021-05-27
backend: use dns.NewNoopManager.

DentonGentry:
Android updates its DNS config in updateTUN() when in response
to several different channels from the backend.

There is not an Android-specific NewOSConfigurator, we end
up pulling in the Linux NewOSConfigurator:
https://github.com/tailscale/tailscale/blob/main/net/dns/manager_linux.go

The Linux DNS manager expects to be able to write to /etc/resolv.conf,
which does not work on Android and causes errors in updating DNS config.

Instead, allocate dns.NewNoopManager to disable the DNS manager, and
rely on the updateTUN() code to handle DNS.

Fixes https://github.com/tailscale/tailscale/issues/1956


bradfitz:
LGTM

PR #10, opened 2021-05-28
backend: Send DNS config through CallbackRouter.

DentonGentry:
Using NewNoopManager avoided the errors from trying to overwrite
/etc/resolv.conf, but still didn't fully work. Route DNS
config through the CallbackRouter.

Fixes https://github.com/tailscale/tailscale/issues/1956

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -87,16 +87,14 @@ func newBackend(dataDir string, jvm *jni.JVM, store *stateStore, settings settin
 		logID.UnmarshalText([]byte(storedLogID))
 	}
 	b.SetupLogs(dataDir, logID)
-	d, err := dns.NewNoopManager()
-	if err != nil {
-		return nil, err
+	cb := &router.CallbackRouter{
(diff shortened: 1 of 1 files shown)


cmd/tailscale/backend.go
```diff
@@ -87,16 +87,14 @@ func newBackend(dataDir string, jvm *jni.JVM, store *stateStore, settings settin
 		logID.UnmarshalText([]byte(storedLogID))
 	}
 	b.SetupLogs(dataDir, logID)
-	d, err := dns.NewNoopManager()
-	if err != nil {
-		return nil, err
+	cb := &router.CallbackRouter{
+		SetBoth:  b.setCfg,
+		SplitDNS: false, // TODO: make true
```
bradfitz, line 92: reference a bug in the TODO with more detail?

bradfitz (approved):

cmd/tailscale/backend.go
DentonGentry: Done.

danderson (approved):

PR #11, opened 2021-06-01
ui: show IPv6 address if there's no IPv4 address

DentonGentry:
There are several situations where a peer may only have IPv6:
+ it was authorized using an ephemeral node key
  https://tailscale.com/kb/1111/ephemeral-nodes/
+ the Tailnet has IPv4 disabled
  https://twitter.com/bradfitz/status/1398380415124082690

So:
+ if a peer has an IPv4 address, display that.
+ if a peer does not have an IPv4 address, display IPv6.
+ If a peer does not have either IPv4 or IPv6 then leave
  the address blank and pretend nothing happened.

### DIFF
--- a/cmd/tailscale/ui.go
+++ b/cmd/tailscale/ui.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"image"
 	"image/color"
-	"strings"
 	"time"
 
 	"gioui.org/f32"
@@ -25,6 +24,7 @@ import (
 	"gioui.org/widget"
 	"gioui.org/widget/material"
 	"golang.org/x/exp/shiny/materialdesign/icons"
+	"inet.af/netaddr"
 	"tailscale.com/ipn"
 	"tailscale.com/tailcfg"
 
@@ -839,15 +839,13 @@ func (ui *UI) layoutPeer(gtx layout.Context, sysIns system.Insets, p *UIPeer, us
 					})
 				}),
 				layout.Rigid(func(gtx C) D {
-					var addrs []string
+					var bestIP netaddr.IP // IP to show; first IPv4, or first IPv6 if no IPv4
 					for _, addr := range p.Peer.Addresses {
-						if addr.IP().Is6() {
-							// Don't surface IPv6 addresses.
-							continue
+						if ip := addr.IP(); bestIP.IsZero() || bestIP.Is6() && ip.Is4() {
+							bestIP = ip
 						}
-						addrs = append(addrs, addr.IP().String())
 					}
-					l := material.Body2(ui.theme, strings.Join(addrs, ","))
+					l := material.Body2(ui.theme, bestIP.String())
 					l.Color = rgb(0x434343)
 					return l.Layout(gtx)
 				}),


cmd/tailscale/ui.go
```diff
@@ -839,12 +839,22 @@ func (ui *UI) layoutPeer(gtx layout.Context, sysIns system.Insets, p *UIPeer, us
...
 					for _, addr := range p.Peer.Addresses {
 						if addr.IP.Is6() {
-							// Don't surface IPv6 addresses.
-							continue
+							if ip6 == "" {
+								ip6 = addr.IP.String()
+							}
+						} else {
+							if ip4 == "" {
+								ip4 = addr.IP.String()
```
eliasnaur, line 850: This code will select the last address for display. I don't know whether the ordering of Peer.Addresses is significant for that to matter.

cmd/tailscale/ui.go
```diff
@@ -839,12 +839,22 @@ func (ui *UI) layoutPeer(gtx layout.Context, sysIns system.Insets, p *UIPeer, us
...
+							}
 						}
-						addrs = append(addrs, addr.IP.String())
+					}
+					if ip4 != "" {
+						addrs = append(addrs, ip4)
+					} else {
+						addrs = append(addrs, ip6)
```
eliasnaur, line 857: If you only ever want to display one address, then I suggest avoiding the "addrs" slice altogether. However, it still seems useful to display one of each address type, if available. That is:

```
if ipv4 != "" {
    addrs = append(addrs, ipv4)
}
if ipv6 != "" {
    addrs = append(addrs, ipv6)
}
```

DentonGentry: I tried displaying both, if present, but I don't think it works very well:
- it puts a wall of numbers on the screen, but
- the IPv6 address cannot be copied (only the first address is) yet is too long to type it in by hand.

Changed to remove the addrs slice, and display ipv4 if present and ipv6 otherwise.

Screenshot_20210625-171205
Screenshot_20210625-171042

cmd/tailscale/ui.go
```diff
@@ -839,15 +838,25 @@ func (ui *UI) layoutPeer(gtx layout.Context, sysIns system.Insets, p *UIPeer, us
 					})
 				}),
 				layout.Rigid(func(gtx C) D {
-					var addrs []string
+					var addr string
```
bradfitz, line 841: I"d move this down closer to its use.

But simpler and fewer string allocations might be:

```go
    var bestIP netaddr.IP // IP to show; first IPv4, or first IPv6 if no IPv4
    for _, addr := range p.Peer.Addresses {
           if ip := addr.IP(); bestIP.IsZero() || bestIP.Is6() && ip.Is4() {
                 bestIP = ip
           }
     }
     l := material.Body2(ui.theme, bestIP.String())

Comment (DentonGentry) [reply cmd/tailscale/ui.go]: Done.

**Review by bradfitz — APPROVED** [review]

**Comment by bradfitz** [comment]
For commit message, perhaps:

```
ui: show IPv6 address if there's no IPv4 address
```

cmd/tailscale/ui.go
DentonGentry: I think it selects the first address, but regardless: if we ever do support multiple IPv4 or multiple IPv6 addresses on a Tailscale interface, we'll need to consider which to display. Right now the first one will generally be the best one to display.

PR #15, opened 2021-07-16
Update for 1.11 unstable release

DentonGentry:


android/build.gradle
```diff
@@ -25,8 +25,8 @@ android {
     defaultConfig {
         minSdkVersion 22
         targetSdkVersion 29
-        versionCode 48
-        versionName "v1.10.2-36fe8addc39-g74a18b3359b"
+        versionCode 49
```
eliasnaur, line 28: Do you intend to publish a 1.11 unstable version? If not, bumping the version code is not necessary.

eliasnaur (approved):

android/build.gradle
DentonGentry: I'm planning to use the Open testing track in the Play Store to deliver 1.11.x builds, like those done for other platforms in https://pkgs.tailscale.com/unstable/

We'd link to it on Twitter to find a few users willing to try early builds.

DentonGentry:
I'm going to merge this to be able to get a build into the Internal Testing track in the Play Store.

Brad if you want anything changed we can do that before promoting the build to Open Testing.

PR #16, opened 2021-07-26
Makefile: download Tailscale's Go toolchain

DentonGentry:
Tailscale maintains a patched Go toolchain, pulling in
fixes early. Download the toolchain and use it to build.

Fixes https://github.com/tailscale/tailscale/issues/2450
in a better way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,13 @@ TAILSCALE_COMMIT=$(shell echo $(TAILSCALE_VERSION) | cut -d - -f 2 | cut -d t -f
 VERSIONCODE=$(lastword $(shell grep versionCode android/build.gradle))
 VERSIONCODE_PLUSONE=$(shell expr $(VERSIONCODE) + 1)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -30,13 +33,24 @@ tag_release:
 	git commit -sm "android: bump version code" android/build.gradle
 	git tag -a "$(VERSION_LONG)"
 
+$(TOOLCHAINDIR):
```
bradfitz, line 36: does this handle resuming if the mkdir succeeds but the tar is interrupted?

bradfitz:
Adding @danderson as a reviewer because I suck at Make and shell.

Makefile
DentonGentry: Done.

It always spends the time to compute the sum of the toolchain even for make targets which don't use the toolchain, because I cannot get the in-target conditional to work. This makes tag_release take longer that it otherwise might, which we do once per release.

DentonGentry:
I need to submit this to build 1.12.1. Can be further updated with another PR.

PR #20, opened 2021-10-01
cmd/tailscale: report ChromeOS.

DentonGentry:
Fixes https://github.com/tailscale/tailscale/issues/2971


bradfitz (approved):
modelName is called twice but that's probably fine

PR #21, opened 2021-10-04
WIP: cmd/tailscale: implement getInterfaces.

DentonGentry:
This is intended to be the Android implementation of a net.Interfaces() API able to function in an SDK30 runtime.
Updates https://github.com/tailscale/tailscale/issues/2293

Passing primitive types across JNI is relatively straightforward, passing a single object of a complex class is annoying but still possible, but passing lists and other more complex data structures is way harder. As such, I added a Java routine to render the interface information to a string and pass that across JNI as a primitive type for Go code to parse.


cmd/tailscale/main.go
```diff
@@ -240,6 +242,7 @@ func (a *App) runBackend() error {
 		rcfg *router.Config
 		dcfg *dns.OSConfig
 	}
+	interfaces.RegisterInterfaceGetter(a.getInterfaces)
```
bradfitz, line 245: Can you try doing this line earlier in func main, before the goroutine to start `runBackend`?

I think it's fine here, but I can't convince myself there's not a race from something else.

DentonGentry:
I think this will work for 1.16.0, and lets us move to SDK 30.

PR #23, opened 2021-11-20
Android: retrieve current DNS servers.

DentonGentry:
Add getDnsConfigAsString() to retrieve the current DNS
configuration from the Android platform. This implements
several mechanisms to retrieve DNS information, suitable
for different Android versions:

Android 7 and later use ConnectivityManager getAllNetworks(),
then iterate over each network to retrieve DNS servers and
search domains using the LinkProperties.

Android 6 and earlier can only retrieve the currently active
interface using ConnectivityManager getActiveNetwork(), but have
two additional fallback options which leverage the system
properties available in older Android releases.


android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -372,4 +379,267 @@ String getInterfacesAsString() {
 
             return sb.toString();
         }
+
+	// getDnsConfigAsString returns the current DNS configuration as a multiline string:
+	// line[0] DNS server addresses separated by spaces
+	// line[1] search domains
```
bradfitz, line 385: also space-separated?

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -372,4 +379,267 @@ String getInterfacesAsString() {
 
             return sb.toString();
         }
+
+	// getDnsConfigAsString returns the current DNS configuration as a multiline string:
```
bradfitz, line 383: document what empty string means?

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -372,4 +379,267 @@ String getInterfacesAsString() {
...
+			// which will have 100.100.100.100 for its DNS server. We reject
+			// TYPE_VPN in getPreferabilityForNetwork, so it won't be returned.
+			Network active = cMgr.getActiveNetwork();
+			if (active == null) {
+				return "";
+			}
+			networks = new Network[]{active};
+		}
+
+		String[] dnsConfigs = new String[]{"", "", "", ""};
```
bradfitz, line 449: why 4?

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -372,4 +379,267 @@ String getInterfacesAsString() {
...
+				sb.append(ip.getHostAddress() + " ");
+			}
+
+			String d = linkProp.getDomains();
+			if (d != null) {
+				sb.append("\n");
+				sb.append(d);
+			}
+
+			dnsConfigs[idx] = sb.toString();
```
bradfitz, line 470: any check that idx < 4?

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -372,4 +379,267 @@ String getInterfacesAsString() {
...
+			return String.join(" ", servers);
+		} catch (Exception e) {
+			return "";
+		}
+	}
+
+
+	String intToInetString(int hostAddress) {
+		return String.format("%d.%d.%d.%d", (byte)(0xff & hostAddress),
+			(byte)(0xff & (hostAddress >> 8)), (byte)(0xff & (hostAddress >> 16)),
```
bradfitz, line 519: put the >> 16 expression on its own line (and the first one too without the shift) so they all line up

cmd/tailscale/backend.go
```diff
@@ -55,19 +58,28 @@ const (
 	loginMethodWeb    = "web"
 )
 
-var fallbackNameservers = []netaddr.IP{netaddr.IPv4(8, 8, 8, 8), netaddr.IPv4(8, 8, 4, 4)}
+// googleDnsServers are used on ChromeOS, where an empty VpnBuilder DNS setting results
+// in erasing the platform DNS servers. The developer docs say this is not supposed to happen,
+// but nonetheless it does.
+var googleDnsServers = []netaddr.IP{netaddr.IPv4(8, 8, 8, 8), netaddr.IPv4(8, 8, 4, 4),
+	netaddr.IPv6Raw([16]byte{0x20, 0x01, 0x48, 0x60, 0x48, 0x60, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x88, 0x88}),
```
bradfitz, line 65: I'd just use netaddr.MustParseIP for all 4 of these ... it's super fast, doesn't allocate, etc. And it's much more readable.

cmd/tailscale/backend.go
```diff
@@ -313,3 +327,92 @@ func (b *backend) SetupLogs(logDir string, logID logtail.PrivateID) {
 		log.Printf("SetupLogs: filch setup failed: %v", filchErr)
 	}
 }
+
+func (b *backend) logDnsConfigMechanisms() {
```
bradfitz, line 331: for Go, logDNSConfigMechanisms (initialisms are all one case, so no "Dns")

do whatever in Java.

bradfitz (commented):
Very superficial review.

Maybe you can walk me through this on Monday.

cmd/tailscale/backend.go
```diff
@@ -87,9 +99,11 @@ func newBackend(dataDir string, jvm *jni.JVM, store *stateStore, settings settin
 		logID.UnmarshalText([]byte(storedLogID))
 	}
 	b.SetupLogs(dataDir, logID)
+
 	cb := &router.CallbackRouter{
-		SetBoth:  b.setCfg,
-		SplitDNS: false, // TODO: https://github.com/tailscale/tailscale/issues/1695
+		SetBoth:         b.setCfg,
+		SplitDNS:        false,
+		GetBaseConfigFn: b.getDnsBaseConfig,
```
bradfitz, line 106: getDNSBaseConfig

cmd/tailscale/backend.go
```diff
@@ -313,3 +327,92 @@ func (b *backend) SetupLogs(logDir string, logID logtail.PrivateID) {
...
+	})
+	if err != nil {
+		log.Printf("getDnsConfigFromLinkProperties JNI: %v", err)
+	} else {
+		log.Printf("getDnsConfigFromLinkProperties: %s", baseConfig)
+	}
+
+	err = jni.Do(b.jvm, func(env *jni.Env) error {
+		cls := jni.GetObjectClass(env, b.appCtx)
+		m := jni.GetMethodID(env, cls, "getDnsServersFromSystemProperties", "()Ljava/lang/String;")
```
bradfitz, line 348: factor out these JNI line(s)?

DentonGentry: Moved the loop inside JNI.Do. It creates a temporary JNI environment, we can't hold Java objects across multiple calls to JNI.Do().

cmd/tailscale/backend.go
DentonGentry: Done

cmd/tailscale/backend.go
DentonGentry: Done.

cmd/tailscale/backend.go
DentonGentry: Done.

android/src/main/java/com/tailscale/ipn/App.java
DentonGentry: Done.

android/src/main/java/com/tailscale/ipn/App.java
DentonGentry: Added one.

android/src/main/java/com/tailscale/ipn/App.java
DentonGentry: Added comment.

android/src/main/java/com/tailscale/ipn/App.java
DentonGentry: Added comment.

android/src/main/java/com/tailscale/ipn/App.java
DentonGentry: Done

DentonGentry:
Still testing, it didn't work reliably when my phone would wake from sleep on an LTE network. The app was being notified of the link change before the LTE interface had been configured with DNS servers.

Switching callback to onLinkPropertiesChanged is working better.

DentonGentry:
As far as I can tell this is working reliably now. Switching to onLinkPropertiesChanged fixed the issue I had when roaming on LTE.

bradfitz (approved):

PR #27, opened 2021-12-20
Hacky implementation of a QR code for login on fire stick

bradfitz:
Hey @eliasnaur, when you have time, can you do something like this but better?

And only when the deviceModel is an Android TV?

For https://github.com/tailscale/tailscale/issues/1611


eliasnaur:
Superseded by #40.

PR #33, opened 2022-01-07
Add secret Run Exit Node option when "debug" is searched for.

bradfitz:
And bump tailscale dep, to bring in new ipn.Prefs API and wire up
ExitDNS for Android.

This change has no visible behavior change to anybody unless they
search for "debug" and then hit the "..." menu.

Updates tailscale/tailscale#1738

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -29,6 +29,7 @@ import (
 	"tailscale.com/types/logger"
 	"tailscale.com/util/dnsname"
 	"tailscale.com/wgengine"
+	"tailscale.com/wgengine/netstack"
 	"tailscale.com/wgengine/router"
 )
 
@@ -120,6 +121,9 @@ func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateSto
 		return nil, fmt.Errorf("runBackend: NewUserspaceEngine: %v", err)
 	}
 	b.logIDPublic = logID.Public().String()
+	if err := startNetstack(log.Printf, dialer, engine); err != nil {
+		return nil, fmt.Errorf("startNetstack: %w", err)
+	}
 	local, err := ipnlocal.NewLocalBackend(logf, b.logIDPublic, store, dialer, engine)
 	if err != nil {
 		engine.Close()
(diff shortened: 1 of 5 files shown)


cmd/tailscale/ui.go
```diff
@@ -260,7 +265,9 @@ func (ui *UI) layout(gtx layout.Context, sysIns system.Insets, state *clientStat
 
 	for _, e := range ui.search.Events() {
 		if _, ok := e.(widget.ChangeEvent); ok {
-			events = append(events, SearchEvent{Query: ui.search.Text()})
+			text := ui.search.Text()
+			ui.showDebugMenu = strings.EqualFold(text, "debug")
```
DentonGentry, line 269: I think this means that if the debug menu is enabled, searching for anything which isn't "debug" will disable the menu.

If that is the intent that is fine, this is clearly a transitory mechanism. Otherwise, it could be conditional to only set not clear ui.showDebugMenu.

DentonGentry (approved):
The code looks reasonable, though I'm not very enthused the timing of being so close to 1.20.

cmd/tailscale/backend.go
```diff
@@ -428,3 +432,17 @@ func (b *backend) getDNSBaseConfig() (dns.OSConfig, error) {
...
+	tunDev, magicConn, ok := e.(wgengine.InternalsGetter).GetInternals()
+	if !ok {
+		return fmt.Errorf("%T is not a wgengine.InternalsGetter", e)
+	}
+	ns, err := netstack.Create(logf, tunDev, e, magicConn, dialer)
+	if err != nil {
+		return fmt.Errorf("netstack.Create: %w", err)
+	}
+	ns.ProcessLocalIPs = false
+	ns.ProcessSubnets = true
```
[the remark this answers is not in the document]
bradfitz: Yeah, an exit node is just a subnet router.

cmd/tailscale/ui.go
bradfitz: oh, that's an interesting idea, to make it a one-time enable and not go away once you've entered the magic code.

bradfitz: Done.

DentonGentry (approved):

PR #38, opened 2022-02-16
android: make apk android tv compatible

tmm1:
These bits signal to Android TV OS that the apk is compatible with "leanback", so the apk can be more easily installed and launched on Android TV (and FireTV) devices.

This is also a pre-requisite to submitting the apk to the Play Store for Android TV.

See https://developer.android.com/training/tv/start/start#tv-activity

cc https://github.com/tailscale/tailscale/issues/1611

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -10,6 +10,10 @@
 	<!-- Disable input emulation on ChromeOS -->
 	<uses-feature android:name="android.hardware.type.pc" android:required="false"/>
 
+	<!-- Signal support for Android TV -->
+	<uses-feature android:name="android.software.leanback" android:required="false" />
+	<uses-feature android:name="android.hardware.touchscreen" android:required="false" />
+
 	<application android:label="Tailscale" android:icon="@mipmap/ic_launcher" android:roundIcon="@mipmap/ic_launcher_round"
 		android:name=".App" android:allowBackup="false">
 		<activity android:name="IPNActivity"
@@ -21,6 +25,7 @@
 			<intent-filter>
 				<action android:name="android.intent.action.MAIN" />
 				<category android:name="android.intent.category.LAUNCHER" />
+				<category android:name="android.intent.category.LEANBACK_LAUNCHER" />
 			</intent-filter>
 			<intent-filter>
 				<action android:name="android.service.quicksettings.action.QS_TILE_PREFERENCES" />


DentonGentry (changes requested):
I'm sorry, but there are too many things broken with the Android TV experience to actually list it in the Play Store yet:
- requires that a browser be present to handle login. We need to get a QR code up onto the login page so people can point a phone at the TV and use their phone browser to authenticate.
- GUI widgets don't handle focus properly when using a remote control.
- no way in the UI to easily select to be an exit node (TV devices which never leave the home are good choices for being an exit node). We took the first step toward this in the current release, we need to finish it.

You're welcome to build the app locally to load onto your TV device, but I'm not ready to put the support in the product listed in the Play Store until we're actually ready to support TV devices.

tmm1:
I'm not suggesting you submit to the Play Store right away. Clearly a bit of work is required before that step.

But I think this makes sense to merge, because the resulting apk will be easier to use on Android TV even when installed manually.

As explained on the link above:

>If you do not include the [CATEGORY_LEANBACK_LAUNCHER](https://developer.android.com/reference/android/content/Intent#CATEGORY_LEANBACK_LAUNCHER) intent filter in your app, it is not visible to users running Google Play on TV devices. Also, **if your app does not have this filter when you use developer tools to load it onto a TV device, the app does not appear in the TV user interface**.

bradfitz:
Which Android TV are you using?

This does seem mostly harmless at least meanwhile, before we fix all the stuff @DentonGentry mentioned before actually submitting to the Play Store.

But I'd like to see this running on a device first.

tmm1:
Different Android TV versions use different leanback launchers, which generally search for leanback compatible installed apks and show those on the home screen or list of apps.

If you're looking for an Android TV device with the Play Store, I would recommend one of:

- Walmart's onn streamer 4k
- TiVo Stream 4K
- Chromecast Google TV

You may also find https://vysor.io helpful, to interact with the devices without a HDMI display present.

Note that to actually get an APK onto the Play Store for Android TV, many more steps are required including a manual review process that is initiated via the Play console. You also need all sorts of art work and screenshots before the app can be approved. So the changes in this PR won't really affect anything on that side.

I tested this on my TiVo Stream 4K. When I install the apk off of the `main` branch, it doesn't appear anywhere on the home screen, under apps, or under Settings > Apps

When I install the apk off this branch, I can see the app listed in the TV interface (albeit with the wrong icon, since `android:banner` is still missing):

screenshot-1645125328141

tmm1:
On my TiVo Stream 4K, there is no browser installed. When I click "sign in with other" it says "no app that can do that".

However, in my own apps I use an embedded `android.webkit.WebView` which does work on this device. And it even supports basic dpad navigation and focus. That may be worth looking into.

tmm1:
FYI, I tried to `go get -u gioui.org` to include https://git.sr.ht/~eliasnaur/gio/commit/27762fdd465845931cd3bdec73ad1fdad1188e17, but it fails to compile:

```
# github.com/tailscale/tailscale-android/cmd/tailscale
cmd/tailscale/ui.go:523:8: undefined: pointer.Rect
cmd/tailscale/ui.go:791:15: not enough arguments in call to c.Layout
	have (layout.Context)
	want (layout.Context, layout.Widget)
cmd/tailscale/ui.go:1148:28: not enough arguments in call to material.Switch
	have (*material.Theme, *widget.Bool)
	want (*material.Theme, *widget.Bool, string)
cmd/tailscale/ui.go:1170:32: not enough arguments in call to material.IconButton
	have (*material.Theme, *widget.Clickable, *widget.Icon)
	want (*material.Theme, *widget.Clickable, *widget.Icon, string)
```

tmm1:
>albeit with the wrong icon, since `android:banner` is still missing

@bradfitz can you surface this to a designer?

The specs are here: https://developer.android.com/training/tv/start/start#banner

>The banner should be an xhdpi resource with a size of 320 x 180 px.

bradfitz:
@tmm1, I filed https://github.com/tailscale/tailscale/issues/4179 for that.

I'll merge this one.

PR #41, opened 2022-03-13
java: format strings containing integers in ROOT locale.

DentonGentry:
We use strings to pass structured data from the JVM to Go.
In a locale using Indian-Arabic numerals: ۰ ۱ ۲ ۳ ۴ ۵ ۶ ۷ ۸ ۹
the Java code will format decimal strings using Indian-Arabic
glyphs.

Go doesn't get a locale set automatically by the Android
runtime, so it always parses strings in a default en-US
`unable to parse "lo ١ ٦٥٥٣٦ true false true false false |": expected integer`

Make the Java code format using the ROOT locale. These strings
are purely internal to pass between the two runtimes, they are
not shown to the user.

Fixes https://github.com/tailscale/tailscale/issues/4156

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.java
+++ b/android/src/main/java/com/tailscale/ipn/App.java
@@ -54,6 +54,7 @@ import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.NotificationManagerCompat;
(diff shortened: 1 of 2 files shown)


DentonGentry:
I don't have a phone with a Saudi Arabian or other similar local available to try. This doesn't break the US locale, and should fix locales using Indian-Arabic numerals.

bradfitz:
Is this PR meant to be bumping versions of things and disabling vulkan? Seems unrelated

DentonGentry:
Created the branch from the wrong place.

bradfitz (approved):
I think you want Locale.ROOT, though, which is like locale "C" in C.

> The root locale is the locale whose language, country, and variant are empty ("") strings. This is regarded as the base locale of all locales, and is used as the language/country neutral locale for the locale sensitive operations.

PR #42, opened 2022-03-16
android: try to add a Android TV Leanback launcher icon

bradfitz:
Maybe it works on Android TV, but this doesn't work on a Fire Stick.

Updates tailscale/tailscale#4179


bradfitz:
(No clue whether this works)

DentonGentry (approved):

bradfitz:
@tmm1, can you try this out before I merge it? From reading Amazon docs, they're supposed to be the same in this regard, so if doesn't work on my Fire Stick, I imagine it also might not work on an Android TV device.

tmm1:
Diff LGTM

I can try it on device later. But you can verify in the emulator too.

It is expected that it won't work on FireOS. You have to upload the banner into the Amazon App Store separately and it's delivered during install, for some unknown reason.

bradfitz:
Ah, bizarre. Thanks for the info!

tmm1:
screenshot-1647406241814

looks good

rosszurowski (changes requested):
Change seems good, but let's use the icons I linked in https://github.com/tailscale/tailscale/issues/4179 before merging! Happy to push to the PR if it'd be easier for you.

bradfitz:
@rosszurowski, push away. I'm not sure what the rules are for which resolution PNG files go in which directories. I was guessing.

tmm1:
All you need is drawable-xhdpi/tv_banner.png using the smaller image

The bigger image is not in the apk but uploaded to the store separately

rosszurowski:
320x180 seems shockingly small, but [the docs agree](https://developer.android.com/training/tv/start/start#banner). I'll update my commit to only contain the 320w file in the xhdpi folder and we can upload the larger one to the app store.

rosszurowski (approved):
Ok, this should be good to go now!

bradfitz:
Thanks!

PR #45, opened 2022-04-11
Add ability to choose a custom coordination server

half-duplex:
This allows users to set a custom coordination/login/control server via a 3-dots menu on the login view.

This is sub-optimal in at least three ways, but I'm already out of my depth.
- The app restart is gross and shouldn't be necessary, but ipnlocal can't update ControlURL after Start() and I couldn't find another clean way to trigger ipnlocal.Start() again. I don't know golang, so don't dare go deeper.
- It should be clear on the login view (and maybe logged-in view) that a custom server is configured, maybe with an :x:
- It should be more obvious how to reset to default (erase the contents of the field and save)

<details><summary>Screenshots</summary>

Screenshot of the normal login page, with a small three-dot menu button at the top right
Screenshot of the menu open, with a header, "Advanced settings", and one item, "Use login server"
Screenshot of the login server change page, showing three elements: a textbox with the placeholder "https://controlplane.tailscale.com", a "Save and restart" button, and a "Cancel" button
</details>

Related:
https://twitter.com/dave_universetf/status/1415046381996167170
https://github.com/juanfont/headscale/issues/58
https://github.com/juanfont/headscale/issues/146

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -56,8 +56,9 @@ type settingsFunc func(*router.Config, *dns.OSConfig) error
 const defaultMTU = 1280 // minimalMTU from wgengine/userspace.go
 
 const (
-	logPrefKey         = "privatelogid"
-	loginMethodPrefKey = "loginmethod"
+	logPrefKey               = "privatelogid"
+	loginMethodPrefKey       = "loginmethod"
+	customLoginServerPrefKey = "customloginserver"
 )
 
 const (
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -180,6 +180,10 @@ type FileSendEvent struct {
 	Updates func(FileSendInfo)
 }
 
+type SetLoginServerEvent struct {
+	URL string
+}
+
 // UIEvent types.
 type (
 	ToggleEvent       struct{}
@@ -429,6 +433,12 @@ func (a *App) runBackend() error {
 					go b.backend.StartLoginInteractive()
 					signingIn = true
 				}
+			case SetLoginServerEvent:
+				state.Prefs.ControlURL = e.URL
+				b.backend.SetPrefs(state.Prefs)
+				// A hack to get around ipnlocal's inability to update
+				// ControlURL after Start()... Can we re-init instead?
+				os.Exit(0)
 			case LogoutEvent:
 				go b.backend.Logout()
 			case ConnectEvent:
@@ -1106,6 +1116,9 @@ func (a *App) processUIEvents(w *app.Window, events []UIEvent, act jni.Object, s
 			requestBackend(e)
 		case WebAuthEvent:
 			a.store.WriteString(loginMethodPrefKey, loginMethodWeb)
+			requestBackend(e)
+		case SetLoginServerEvent:
+			a.store.WriteString(customLoginServerPrefKey, e.URL)
 			requestBackend(e)
 		case LogoutEvent:
 			a.signOut()
--- a/cmd/tailscale/ui.go
+++ b/cmd/tailscale/ui.go
@@ -63,6 +63,11 @@ type UI struct {
 
 	signinType signinType
 
+	setLoginServer    bool
+	loginServer       widget.Editor
+	loginServerSave   widget.Clickable
+	loginServerCancel widget.Clickable
+
 	self  widget.Clickable
 	peers []widget.Clickable
 
@@ -93,12 +98,13 @@ type UI struct {
 		dismiss Dismiss
 		show    bool
 
-		copy   widget.Clickable
-		reauth widget.Clickable
-		bug    widget.Clickable
-		beExit widget.Clickable
-		exits  widget.Clickable
-		logout widget.Clickable
+		useLoginServer widget.Clickable
+		copy           widget.Clickable
+		reauth         widget.Clickable
+		bug            widget.Clickable
+		beExit         widget.Clickable
+		exits          widget.Clickable
+		logout         widget.Clickable
 	}
 
 	// The current pop-up message, if any
@@ -240,6 +246,7 @@ func newUI(store *stateStore) (*UI, error) {
 	ui.root.Axis = layout.Vertical
 	ui.intro.list.Axis = layout.Vertical
 	ui.search.SingleLine = true
+	ui.loginServer.SingleLine = true
 	ui.exitDialog.list.Axis = layout.Vertical
 	ui.shareDialog.list.Axis = layout.Vertical
 	return ui, nil
@@ -342,6 +349,21 @@ func (ui *UI) layout(gtx layout.Context, sysIns system.Insets, state *clientStat
 	if ui.webSignin.Clicked() {
 		ui.signinType = webSignin
 		events = append(events, WebAuthEvent{})
+	}
+
+	if ui.loginServerSave.Clicked() {
+		text := ui.loginServer.Text()
+		ui.showMessage(gtx, "Login server saved, relaunch the app")
+		events = append(events, SetLoginServerEvent{URL: text})
+	}
+	if ui.loginServerCancel.Clicked() {
+		ui.setLoginServer = false
+	}
+
+	if ui.menuClicked(&ui.menu.useLoginServer) {
+		ui.setLoginServer = true
+		savedLoginServer, _ := ui.store.ReadString(customLoginServerPrefKey, "")
+		ui.loginServer.SetText(savedLoginServer)
 	}
 
 	if ui.menuClicked(&ui.menu.copy) && localAddr != "" {
@@ -502,7 +524,7 @@ func (ui *UI) layout(gtx layout.Context, sysIns system.Insets, state *clientStat
 
 	// 3-dots menu.
 	if ui.menu.show {
-		ui.layoutMenu(gtx, sysIns, expiry, exitID != "" || len(state.backend.Exits) > 0)
+		ui.layoutMenu(gtx, sysIns, expiry, exitID != "" || len(state.backend.Exits) > 0, needsLogin)
 	}
 
 	if ui.qr.show {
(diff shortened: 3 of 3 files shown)


DentonGentry:
I expect to support setting the control server URL via MDM deployments, as a variable which can be set. I don't especially want to have it in the UI like this, most people using the app won't know what it is or why it would be so emphasized.

half-duplex:
I'd have stuck it in the debug stuff so it's not in the UI for people who aren't specifically looking for it, but the search bar to access debug isn't accessible until after logging in to the default server =\
Failing that, I figured having it behind the menu button under an "Advanced settings" header was sufficiently obscure.

bradfitz:
There's already precedent for a hidden debug menu:

https://github.com/tailscale/tailscale/issues/1738#issuecomment-1007633354

You type "debug" into the search box and then debug options show up in the menu.

I'd be fine with that for Android. For iOS we should only do the MDM thing.

half-duplex:
That's the debug stuff I meant - Enabling the debug menu options requires having access to the search bar, which is only shown post-login. I like how you got that in without needing any added UI elements :smile_cat:
Creating an account on tailscale.com and signing in to that before being able to switch to a custom server would _work_, but it would be a pretty annoying UX, especially for corporate users (unless I'm unaware of android MDM features for that sort of thing?)

DentonGentry:
This is how one would set up for an MDM on Android:
https://developer.android.com/work/managed-configurations.html#define-configuration
https://developers.google.com/android/work/play/emm-api/managed-configurations
https://developers.google.com/zero-touch/guides/customer/how-it-works

Once done, users who don't have a full MPM can use an app to toggle settings like TestDPC: https://play.google.com/store/apps/details?id=com.afwsamples.testdpc&hl=en_US&gl=US

0x1a8510f2:
>Once done, users who don't have a full MPM can use an app to toggle settings like TestDPC

FWIW I (as another soon-to-be user of headscale) would strongly prefer not to have to rely on another app to do this. Obviously I'll take what I can get, but access to a developer menu via the login screen where the option is available would be much better IMO, even on top of MDM.

For one, this allows users of low-end devices with minimal storage and no MDM to still self-host. Also, it means that more debug/developer options can easily be included in the future and also that this feature can be documented if required without relying on or pointing to a 3rd party.

VelocityDesign:
Is there any progress on this PR, or are people just brainstorming for how to implement this feature?

DentonGentry:
An approach which uses the Android MDM APIs to set properties would be reasonable, as there are ways to set properties which don't involve installing a full MDM.

fernandoenzo:
> FWIW I (as another soon-to-be user of headscale) would strongly prefer not to have to rely on another app to do this. Obviously I'll take what I can get, but access to a developer menu via the login screen where the option is available would be much better IMO, even on top of MDM.
> 
> For one, this allows users of low-end devices with minimal storage and no MDM to still self-host. Also, it means that more debug/developer options can easily be included in the future and also that this feature can be documented if required without relying on or pointing to a 3rd party.

I totally agree with you. I think this pull request should be accepted as is. This would be a very nice feature identical to what other apps like [Element](https://github.com/vector-im/element-android) (Matrix client) already implement, letting the default server as `matrix.org` but allowing the user to easily change it for another custom one.

An MDM approach is the complete opposite of ease of use. No one would really understand or agree with the need to install an app like TestDPC just to configure a single URL for another.

bradfitz:
> I think this pull request should be accepted as is. 

We _do_ want Android users to be able to configure their control plane server URL.

We do _not_ want to add gunk to menus that most users won't use and will be confused by. (Keep in mind that us nerds in these comments on this PR and related issues are not Tailscale's typical users. Many Tailscale users aren't even tech people; they just need to use it for work because their company told them that's what their corporate VPN is.)

So we need to do something slightly hidden to make this option not appear right away. In the past, we hid an option unless you typed "debug" into the search box. We could also wire up the Tailscale CLI to the `adb shell` or the Android GUI. Or we could require that the phone's in developer mode. Or some combination.

But what we won't accept as is is adding nerdy options to menus. So let's find something more hidden.

half-duplex:
> I think this pull request should be accepted as is.

It definitely needs some tweaks from someone who knows what they're doing to be in a state I'd consider good even for powerusers, but thanks :heart:

> We need to do something slightly hidden to make this option not appear right away.

The app still has tools that show up if you search "debug", but that search box isn't available until well after this setting would need to be changed.

I'm fine with using `adb shell` to configure it, or the menu only showing if developer options or ADB are enabled. MDM-only would still be better than the current situation.

If you want to make it __really__ hidden with __zero__ changes to this app:
- Build a trivial app that does nothing but set the required settings
- Give it the same package name, and v0.0.1 or so
- Sign it with the same key
Users wanting a custom server would sideload that, set the setting, then update it to the normal version using the play store.

fernandoenzo:
> Many Tailscale users aren't even tech people; they just need to use it for work because their company told them that's what their corporate VPN is.)
Indeed, I have no doubt that most users will be from companies that simply tell them to use this program. But what if those companies are using their own `headscale` server? So we are in the same situation, or worse, as we would require non-expert users to access the `adb shell`, configure via MDM, or look for a hidden debug menu.

I still think that the best option, for everyone, without a doubt, is the one proposed in this pull request.

bradfitz:
> But, what if those companies are using their own headscale server?

That's why we really want MDM support. Then companies can push out provisioning profiles or whatnot.

But generally, most Headscale users are nerdy folk (like me) doing their own self-hosting. Hiding this a bit for nerds for personal use seems fine.

> I still think that the best option, for everyone, without a doubt, is the one proposed in this pull request.

Yes, you've made your position clear. I hope I've made ours clear.

We're not going to accept this being super prominent in the UI where it'll confuse non-tech users.

bradfitz:
> We're not going to accept this being super prominent in the UI

That said, I haven't patched this in and built it. My Android dev environment always rots (and/or I forget where things are at) whenever I go to hack on it more.

@half-duplex, have screenshots?

half-duplex:
Yep, or you can play with the APK in [my fork's tag](https://github.com/half-duplex/tailscale-android/releases/tag/v1.23.152ish).

<details><summary>Screenshots</summary>

Screenshot of the normal login page, with a small three-dot menu button at the top right
Screenshot of the menu open, with a header, "Advanced settings", and one item, "Use login server"
Screenshot of the login server change page, showing three elements: a textbox with the placeholder "https://controlplane.tailscale.com", a "Save and restart" button, and a "Cancel" button
</details>

0x1a8510f2:
I personally like the "if developer options are enabled" if there is an API for that, or the "v0.0.1 config app" idea.

My usecase involves letting non-tech people connect to my VPN (obviously without setting up MDM on their phones) so I would prefer that the feature is easy enough to access that I can give them instructions to do so, without relying on ADB or a third party app, even if it's not necessarily visible right away or as prominent as other options.

While I don't think @half-duplex 's solution would be confusing necessarily (I also somewhat hope you think so too after seeing the screenshots :P), I can see your point of view and, ultimately, we can't demand anything given that everyone requesting this feature is explicitly NOT paying tailscale :P.

That said, I ultimately hope for some sort of compromise like "tap out 'debug' in Morse code on the screen to open the debug menu" or use the konami code instead or "click an empty area 8 times in a row and then hold it", all of which should be sufficiently hidden.

nnsee:
It's unfortunate that the maintainers of Tailscale are holding this position regarding having a relatively hidden option in the UI.

> We do not want to add gunk to menus that most users won't use and will be confused by.

This "gunk" is hidden behind very deliberate on-screen selections. It's hard to imagine someone accidentally entering that screen and being unable to back out of it. But if that's _why_ this functionality isn't being added, then it's pretty easy to add a scary red warning text that instructs the user to press "Back" if they don't know what this menu is for. Regardless, I think for most people, insinuating that they're too technologically illiterate or "confused" to back out of a menu is somewhat insulting.

Meanwhile, this over-simplification ends up alienating the users who find Tailscale to be an excellent piece of software but want to, for one reason or another, self-host the control plane. If I want my partner or mother to access an internal file server, I'm not going to be installing MDM software on their devices. I'm not going to be enabling developer mode, or dropping into an adb shell either. It should be possible to specify the URL without relatively dirty hacks. I'm not sure why Tailscale insists that the _end users_ of a self-hosted control plane must be _as_ technologically literate as the one who hosts the control plane. I can tell my mother to press the three dots, select the "advanced option", and enter an URL over the phone. I cannot do the same with any of the other suggested options.

I'm not demanding that Tailscale merge this PR in particular, but rather seriously reconsider their stance on having an option similar to this in the pre-auth UI. I'm not entitled, just very disappointed.

half-duplex:
That solution would be great, or android-style build number rapid tap or long-press.

As I've said, I don't really speak golang, so anything I try would probably need nontrivial edits. Any changes you want to make to this PR are entirely welcome, or closing it if you'd prefer to start from scratch. It's here in equal parts for "in case you're too busy and this is good enough" and to progress the discussion. Thanks!

bradfitz:
Hey @eliasnaur, do you have time to help us finish this up?

Another thing I noticed: when the menu is open, the dotdotdot menu appears but doesn't do anything. It shouldn't be present on this screen:

Screen Shot 2022-08-01 at 9 10 48 AM

bradfitz:
@half-duplex, I just sent https://github.com/tailscale/tailscale-android/pull/55 with this commit plus a bit of tweaking in a second commit.

bradfitz:
Merged in https://github.com/tailscale/tailscale-android/pull/55

bradfitz:
Quoting @DentonGentry from #55:

---

This functionality is available in the 1.29.72 in the Open Testing track in the Play Store.
- Join via web: https://play.google.com/apps/testing/com.tailscale.ipn
- Join via Play Store app: https://play.google.com/store/apps/details?id=com.tailscale.ipn

It will be present in the next major release 1.30.0.

PR #48, opened 2022-04-21
cmd/tailscale: fix netstack init, call SetLocalBackend

bradfitz:
The netstack code on Android was never told about the LocalBackend,
so the peerapi interception wasn't working.

Fixes tailscale/tailscale#4449
Fixes tailscale/tailscale#4293

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -121,16 +121,27 @@ func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateSto
 		return nil, fmt.Errorf("runBackend: NewUserspaceEngine: %v", err)
 	}
 	b.logIDPublic = logID.Public().String()
-	if err := startNetstack(log.Printf, dialer, engine); err != nil {
-		return nil, fmt.Errorf("startNetstack: %w", err)
+	tunDev, magicConn, ok := engine.(wgengine.InternalsGetter).GetInternals()
+	if !ok {
+		return nil, fmt.Errorf("%T is not a wgengine.InternalsGetter", engine)
 	}
-	local, err := ipnlocal.NewLocalBackend(logf, b.logIDPublic, store, dialer, engine, 0)
+	ns, err := netstack.Create(logf, tunDev, engine, magicConn, dialer)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/tailscale/backend.go
```diff
@@ -121,16 +121,27 @@ func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateSto
...
 	}
-	local, err := ipnlocal.NewLocalBackend(logf, b.logIDPublic, store, dialer, engine, 0)
+	ns.ProcessLocalIPs = false
+	ns.ProcessSubnets = true
+	lb, err := ipnlocal.NewLocalBackend(logf, b.logIDPublic, store, dialer, engine, 0)
 	if err != nil {
 		engine.Close()
 		return nil, fmt.Errorf("runBackend: NewLocalBackend: %v", err)
 	}
+	ns.SetLocalBackend(lb)
```
bradfitz, line 139: this is the one-line fix.

the rest is just for ordering: SetLocalBackend has to be called before ns.Start, so the func didn't work.

cmd/tailscale/backend.go
```diff
@@ -121,16 +121,27 @@ func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateSto
...
+	tunDev, magicConn, ok := engine.(wgengine.InternalsGetter).GetInternals()
+	if !ok {
+		return nil, fmt.Errorf("%T is not a wgengine.InternalsGetter", engine)
+	}
+	ns, err := netstack.Create(logf, tunDev, engine, magicConn, dialer)
+	if err != nil {
+		return nil, fmt.Errorf("netstack.Create: %w", err)
 	}
-	local, err := ipnlocal.NewLocalBackend(logf, b.logIDPublic, store, dialer, engine, 0)
+	ns.ProcessLocalIPs = false
```
maisem, line 132: While you're here, can you add comments on why we enable subnets but not local IPs?

maisem (approved):

cmd/tailscale/backend.go
bradfitz: done

PR #51, opened 2022-07-24
DnsConfig: don't use signed bytes when printing.

DentonGentry:
Two commits:

--------

intToInetString(0x0101a8c0) returns "-64.-88.1.1" because Java
integers are always signed. There is not a %u format specifier.

Though the quads of an IP address literally are bytes, they can
be left as an int to pass to String.format. This allows room for
sign bits, so intToInetString(0x0101a8c0) returns "192.168.1.1"

--------

adds JUnit dependencies and basic gradle support to run unit tests,
and a test for DnsConfig.intToInetString().

--------

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/DnsConfig.java
+++ b/android/src/main/java/com/tailscale/ipn/DnsConfig.java
@@ -229,7 +229,7 @@ public class DnsConfig {
 	}
 
 
-	String intToInetString(int hostAddress) {
+	public String intToInetString(int hostAddress) {
 		return String.format(java.util.Locale.ROOT, "%d.%d.%d.%d",
 			(0xff & hostAddress),
 			(0xff & (hostAddress >> 8)),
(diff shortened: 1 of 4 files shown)


maisem (approved):

android/src/main/java/com/tailscale/ipn/DnsConfig.java
```diff
@@ -231,10 +231,10 @@ String getDnsServersFromSystemProperties() {
...
 		return String.format(java.util.Locale.ROOT, "%d.%d.%d.%d",
-			(byte)(0xff & hostAddress),
-			(byte)(0xff & (hostAddress >> 8)),
-			(byte)(0xff & (hostAddress >> 16)),
-			(byte)(0xff & (hostAddress >> 24)));
+			(0xff & hostAddress),
+			(0xff & (hostAddress >> 8)),
+			(0xff & (hostAddress >> 16)),
+			(0xff & (hostAddress >> 24)));
 	}
```
bradfitz, line 238: Perfect little thing to add a test for if we want to get fancy and start learning how to do Java testing 😉

DentonGentry: Added a test.

bradfitz (approved):
Nice!

PR #52, opened 2022-07-26
go.mod: update from OSS.

DentonGentry:
Includes netaddr -> netip changes.

main.go getInterfaces still uses netaddr as it needs to return a net.IPNet (we'd need to add a new implementation in https://github.com/tailscale/tailscale/blob/7c671b0220a5a42303dd68c8b63a5baedcce885e/net/interfaces/interfaces.go#L110 to handle a different type).

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/netip"
 	"path/filepath"
 	"reflect"
 	"strings"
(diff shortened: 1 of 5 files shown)


DentonGentry:
To be honest, right now I don't understand how cmd/tailscale/main.go:getInterfaces manages to work at all. It uses `netaddr.ParseIPPrefix` and appends an `net.IPNet()` to what is supposed to be `[]net.Addr` yet somehow it works.

bradfitz:
I can help with/review this when back. Net.Addr is an interface btw.

DentonGentry:
Ah, I see now. https://github.com/tailscale/tailscale/blob/7c671b0220a5a42303dd68c8b63a5baedcce885e/net/interfaces/interfaces.go#L110 supports net.IPNet now, we'd need to implement different handling to return something else.

sailorfrag (approved):

PR #54, opened 2022-08-01
fix "invalid IP" regression from netaddr to netip migration

bradfitz:


DentonGentry (approved):

bradfitz:
Caveat is I didn't even compile this.

bradfitz:
`make tailscale-fdroid.apk` built.

DentonGentry:
I built a tailscale-debug.apk and tried it, and now see IP addresses instead of "invalid IP"
I'll build an unstable release.

PR #55, opened 2022-08-01
Bradfitz/custom server

bradfitz:
Two commits. First is from @half-duplex that was sent in https://github.com/tailscale/tailscale-android/pull/45; the second hides it a bit behind a few taps.

https://user-images.githubusercontent.com/2621/182201806-d7e96844-d935-4574-90a3-50fe2a47822b.mov

### DIFF
--- a/cmd/tailscale/ui.go
+++ b/cmd/tailscale/ui.go
@@ -62,6 +62,11 @@ type UI struct {
 
 	signinType signinType
 
+	setLoginServer    bool
+	loginServer       widget.Editor
+	loginServerSave   widget.Clickable
+	loginServerCancel widget.Clickable
+
 	self  widget.Clickable
 	peers []widget.Clickable
 
@@ -91,12 +96,13 @@ type UI struct {
 		dismiss Dismiss
 		show    bool
 
-		copy   widget.Clickable
-		reauth widget.Clickable
-		bug    widget.Clickable
-		beExit widget.Clickable
-		exits  widget.Clickable
-		logout widget.Clickable
+		useLoginServer widget.Clickable
+		copy           widget.Clickable
+		reauth         widget.Clickable
+		bug            widget.Clickable
+		beExit         widget.Clickable
+		exits          widget.Clickable
+		logout         widget.Clickable
 	}
 
 	// The current pop-up message, if any
@@ -238,6 +244,7 @@ func newUI(store *stateStore) (*UI, error) {
 	ui.root.Axis = layout.Vertical
 	ui.intro.list.Axis = layout.Vertical
 	ui.search.SingleLine = true
+	ui.loginServer.SingleLine = true
 	ui.exitDialog.list.Axis = layout.Vertical
 	ui.shareDialog.list.Axis = layout.Vertical
 	return ui, nil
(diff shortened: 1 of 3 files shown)


half-duplex:
Thanks!
The pitfalls from my attempt in #45 still apply though, if you want a UX better than "debugging hack":
- The app restart is gross and shouldn't be necessary, but ipnlocal can't update ControlURL after Start() and I couldn't find another clean way to trigger ipnlocal.Start() again. I don't know golang, so don't dare go deeper.
- It should be clear on the login view (and maybe logged-in view) that a custom server is configured, maybe with an x to remove
- It should be more obvious in the "custom server" view how to reset to default (erase the contents of the field and save)

fernandoenzo:
This looks really good! Hope to see it soon in F-Droid!

cmd/tailscale/ui.go
```diff
@@ -271,6 +288,36 @@ func (ui *UI) activeDialog() *bool {
...
+		if now.Sub(hi.at) < recent {
+			filt = append(filt, hi)
+		}
+	}
+	ui.menu.showHistory = filt
+	ui.menu.showHistory = append(ui.menu.showHistory, showChange{
+		at:    now,
+		shown: v,
+	})
+	if len(ui.menu.showHistory) >= 6 {
```
DentonGentry, line 315: Ideally we'd clear showHistory somewhere, so it doesn't accumulate forever.

DentonGentry (approved):

cmd/tailscale/ui.go
bradfitz: It's cleaned up on line 310 above. We only ever even allocate a slice backing array large enough to store how many times you can click in 5 seconds.

bradfitz: Clarified this a bit and removed the debug print on line 318

DentonGentry:
This functionality is available in the 1.29.72 in the Open Testing track in the Play Store.
- Join via web: https://play.google.com/apps/testing/com.tailscale.ipn
- Join via Play Store app: https://play.google.com/store/apps/details?id=com.tailscale.ipn

It will be present in the next major release 1.30.0.

nnsee:
Kudos! Thank you for listening to user feedback and reconsidering your stance, and settling on a very reasonable compromise.

madjam002:
It would be amazing if we could see something like this dev menu to set the login server on iOS too!

Modificator:
On version 1.35.0, UI working well, but can not apply custom server url 😅

megamorf:
Just fyi, the headscale support has been broken since 1.34.0 - see https://github.com/tailscale/tailscale/issues/6671

tonygauderman:
I don't see a method to make this work from the Android TV platform.  Once you select the 3 ellipses, I do not see a way to deselect them.  The back arrow, which works in other apps to do the same thing, returns you to the Android TV home screen.

frustrate3200:
@tonygauderman Hi, Have you found a solution for android tv yet? Thank you!

basnappl:
@tonygauderman & @frustrate3200, I suggest trying to use this tool with your Android TV: https://github.com/Genymobile/scrcpy

This might allow you to tap the ellipsis as on a touch-enable android device.

godsgood33:
Running 1.48.2 on my Pixel and it appears to still be broke.  Changed the server, but not allowing me to use the preauthkey.  Anybody found a fix?

PR #57, opened 2022-08-24
UI: remove "No internet connection"

DentonGentry:
This has never worked well. I think we'd need to periodically re-check
link state, and possibly hook up a generate_204 function, to make this
really work.

Fixes https://github.com/tailscale/tailscale/issues/5422


sailorfrag (approved):

colemickens:
Thank you so much @DentonGentry !

PR #64, opened 2022-09-09
android: add about dialog with license link

willnorris:
Add "about" menu item in both the logged out and logged in menus.  About dialog shows Tailscale version and link to URL with open source licenses.

The version entry in the about dialog does duplicate the version at the top of the logged-out menu.  Maybe that's okay?  Maybe we remove it from the menu?  I seem to recall it was really only there so that the menu wasn't completely empty if you hadn't yet activated the debug menu?  We could also remove it from the About dialog; I just added it because it seemed to make sense.

Updates tailscale/corp#5780

Screen Shot 2022-09-09 at 4 38 16 PM

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -193,6 +193,7 @@ type (
 	WebAuthEvent      struct{}
 	GoogleAuthEvent   struct{}
 	LogoutEvent       struct{}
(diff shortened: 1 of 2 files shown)


DentonGentry (approved):
I think this is fine as is.

PR #65, opened 2022-09-11
cmd/tailscale: show an error when we hit the multiple users Android bug

andrew-d:
Updates tailscale/tailscale#2180

Signed-off-by: Andrew Dunham <andrew@du.nham.ca>

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -81,6 +81,11 @@ var googleDNSServers = []netip.Addr{
 // VPN status was revoked.
 var errVPNNotPrepared = errors.New("VPN service not prepared or was revoked")
 
+// errMultipleUsers is used when we get a "INTERACT_ACROSS_USERS" error, which
+// happens due to a bug in Android. See:
+//    https://github.com/tailscale/tailscale/issues/2180
+var errMultipleUsers = errors.New("VPN cannot be created on this device due to an Android bug with multiple users")
+
 func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateStore,
 	settings settingsFunc) (*backend, error) {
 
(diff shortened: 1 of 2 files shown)


andrew-d:
Note: I don't have an Android device to test on, so I would appreciate someone else helping with that 😃

cmd/tailscale/backend.go
```diff
@@ -81,6 +81,11 @@ var googleDNSServers = []netip.Addr{
 // VPN status was revoked.
 var errVPNNotPrepared = errors.New("VPN service not prepared or was revoked")
 
+// errMultipleUsers is used when we get a "INTERACT_ACROSS_USERS" error, which
+// happens due to a bug in Android. See:
+//    https://github.com/tailscale/tailscale/issues/2180
+var errMultipleUsers = errors.New("VPN cannot be created on this device due to an Android bug with multiple users")
```
bradfitz, line 87: Include some short link users can go to for more info?

Like `tailscale.com/s/multi-user-bug` ?

That can just be a www redirects.txt entry to some open or closed bug in oss.

bradfitz: Whoops, I meant to put this comment on the const multipleUsersText below, not this error.

bradfitz (approved):
Code seems fine, but haven't run it.

Does it compile at least? :)

cmd/tailscale/backend.go
andrew-d: Done in https://github.com/tailscale/tailscale-www/pull/1610 - good idea

andrew-d:
> Code seems fine, but haven't run it.
> 
> Does it compile at least? :)

Yep, compiles when I run `make dockershell` and `make tailscale-debug.apk` inside the container.

PR #71, opened 2022-12-12
cmd/tailscale: implement new batch APIs

raggi:


DentonGentry (approved):
I patched this PR locally, built tailscale-debug.apk, and verified that it is able to reach the tailnet. As expected, the speedtest results are unchanged as Android blocks the relevant kernel APIs.

PR #73, opened 2022-12-27
cmd/tailscale: update to latest gioui.org.

DentonGentry:
There were several API changes since we last updated. We followed examples from the upstream Gio repository:

1. Replace unit.Px with unit.Dp. Followed example in: https://github.com/gioui/gio/commit/3d374913420204ff764171e21a25cb910fccccb1

2. op.Offset now takes int coordinates instead of float. Followed example in: https://github.com/gioui/gio/commit/a63e0cb44a020c2a5c21a524765e98e547226c8a

3. clip.RRect now takes int coordinates instead of float. Followed example in: https://github.com/gioui/gio/commit/48a8540a68c778ca74d792aa8c76193c721199e3

4. Replace system.CommandEvent with key.Event. Followed example in: https://github.com/gioui/gio/commit/6c76fa6dec3990f542c2d6309a1a5619db6427bc

Updates https://github.com/tailscale/tailscale/issues/6824


cmd/tailscale/ui.go
```diff
@@ -1277,14 +1278,21 @@ func (ui *UI) showMessage(gtx layout.Context, msg string) {
 	op.InvalidateOp{}.Add(gtx.Ops)
 }
 
+func MaxDp(p1, p2 unit.Dp) unit.Dp {
```
bradfitz, line 1281: Just max and use generics? ;)

(I don't actually care)

bradfitz (approved):

cmd/tailscale/ui.go
DentonGentry: So far as I can see Gio has moved to integer coordinates everywhere and no longer uses float in its APIs. It just uses float internally now. I don't expect to need a Max for anything other than Dp (Display Pixels), so I'll plan to submit this as it is after a bit more testing.

PR #75, opened 2023-01-02
ui: Fix background color of exit node status.

DentonGentry:
Prior to https://github.com/tailscale/tailscale-android/pull/73/, the exit node status pane was set to a background color using:
    paint.Fill(gtx.Ops, bg)

paint.Fill() is documented to fill the entire clipped area. It assumes that one has already applied a clip area... but no clip area had been set in this code path.

As far as I can tell, that this worked prior to pull #73 was a bug, something had a side-effect of setting a clipping rectangle.

We updated to the head of the Gio repo, which apparently fixed that bug. After pull #73, the paint.Fill() painted the entire app window.

Fix this using a stacked layout. A color filled widget is the lower layer, and will expand to the size of the widget sitting atop of it.

Fixes https://github.com/tailscale/tailscale/issues/6873

### DIFF
--- a/cmd/tailscale/ui.go
+++ b/cmd/tailscale/ui.go
@@ -668,35 +668,48 @@ func (ui *UI) layoutExitStatus(gtx layout.Context, state *BackendState) layout.D
 		text = "Using exit node"
 		bg = rgb(0x338b51)
 	}
-	paint.Fill(gtx.Ops, bg)
 	return material.Clickable(gtx, &ui.openExitDialog, func(gtx C) D {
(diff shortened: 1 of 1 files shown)


DentonGentry:
GitHub isn't doing a great job of rendering the diffs.

What used to be the sole `layout.Inset(...).Layout(...)` is now called within a `layout.Stacked()`

What used to be the `paint.Fill(...)` is now another widget within the `layout.Stacked()`

bradfitz (approved):

bradfitz:
GitHub has a view optional to ignore whitespace.

PR #77, opened 2023-01-29
Dockerfile: go 1.19.5

GiddyGoatGaming:
Updates go from 1.17.5 to 1.19.5 in the dockerfile


DentonGentry (approved):
The comment above this line is true: this Go download is barely used, we use https://github.com/tailscale/go to build the actual client.

GiddyGoatGaming:
Thank you❤️

PR #80, opened 2023-02-12
IPNService: add Chromecast to the apps allowed to bypass the VPN.

DentonGentry:
Needed for LAN discovery of Chromecast devices.
Fixes https://github.com/tailscale/tailscale/issues/3636


bradfitz:
But then you couldn't cast via Plex over Tailscale presumably? I guess it's broken either way?

sanjay900:
This PR just straight up fixes my ability to cast things, before i was not able to cast anything at all.
Previously, i was able to see my local chromecast devices but attempting to cast would just be met with it getting stuck trying to connect, but with this PR, I am able to cast while connected to tailscale.

bradfitz (approved):

PR #84, opened 2023-03-09
IPNReceiver: Add intents to connect and disconnect VPN

brettjenkins:
Added a new class IPNReceiver to listen to intents silently and connect and disconnect the VPN

Updates: https://github.com/tailscale/tailscale/issues/3547

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/IPNReceiver.java
+++ b/android/src/main/java/com/tailscale/ipn/IPNReceiver.java
@@ -8,9 +8,9 @@ public class IPNReceiver extends BroadcastReceiver {
 
     @Override
     public void onReceive(Context context, Intent intent) {
-        if (intent.getAction() == "com.tailscale.ipn.CONNECT_VPN"){ //  android.intent.action.CONNECT_VPN
+        if (intent.getAction() == "com.tailscale.ipn.CONNECT_VPN") {
             connect();
-        }else if(intent.getAction() == "com.tailscale.ipn.DISCONNECT_VPN"){
+        } else if (intent.getAction() == "com.tailscale.ipn.DISCONNECT_VPN") {
             disconnect();
         }
     }
--- a/android/src/main/java/com/tailscale/ipn/IPNReceiver.java
+++ b/android/src/main/java/com/tailscale/ipn/IPNReceiver.java
@@ -1,0 +1,20 @@
+package com.tailscale.ipn;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+
+public class IPNReceiver extends BroadcastReceiver {
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tailscale/ipn/IPNReceiver.java
```diff
@@ -0,0 +1,20 @@
...
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+
+public class IPNReceiver extends BroadcastReceiver {
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        if (intent.getAction() == "com.tailscale.ipn.CONNECT_VPN"){ //  android.intent.action.CONNECT_VPN
```
bradfitz, line 11: formatting: space before final { (and on line 13, which needs a spaces around `else if` too)

bradfitz (commented):
Thanks!

MohitDeshwal:
@brettjenkins Thanks a lot.

android/src/main/java/com/tailscale/ipn/IPNReceiver.java
brettjenkins: No worries, made those changes

Aaroneisele55:
When will this be available to users?

bradfitz:
@Aaroneisele55, first it needs to be approved & merged. Then then it'll be available in the following builds, first with unstable builds that happen regularly, and then the next stable (1.EVEN-NUMBER.0) release, but not the patch releases (1.x.1, 1.x.2, etc).

You're welcome to build & run this on your own anytime, though.

Aaroneisele55:
> @Aaroneisele55, first it needs to be approved & merged. Then then it'll be available in the following builds, first with unstable builds that happen regularly, and then the next stable (1.EVEN-NUMBER.0) release, but not the patch releases (1.x.1, 1.x.2, etc).
> 
> You're welcome to build & run this on your own anytime, though.
> 
@bradfitz,
Thanks for the information,
I just wanted to know when I could expect that build to be released and **where** I could then get it.

Thanks
Aaron

bradfitz:
Looks good to me but I'll leave this to @DentonGentry to do the final review/testing/merge-with-squash.

DentonGentry (approved):
Built and tested that the app's main functionality is not impacted by the change.

josepira58:
I can't get send intent to work, I'm testing from Tasker and from Macrodroid. From tasker nothing happens, macrodroid returns a permissions error. How did you try? greetings
I'm testing with -
Action: com.tailscale.ipn.CONNECT_VPN
package: com.tailscale.ipn
class: com.tailscale.ipn.IPNService
target: service

peiyuanix:
For anyone wondering how to use these intents.

With adb, we can connect/disconnect tailscale with `am broadcast`

```
# connect
adb shell am broadcast -n com.tailscale.ipn/.IPNReceiver -a com.tailscale.ipn.CONNECT_VPN

# disconnect
adb shell am broadcast -n com.tailscale.ipn/.IPNReceiver -a com.tailscale.ipn.DISCONNECT_VPN
```

PR #86, opened 2023-03-26
Add DNS/routing prefs like on desktop

Empty2k12:
Fixes https://github.com/tailscale/tailscale/issues/2155

This is my first time working with Go on Android, so bear with me. 

Review is much appreciated.

Screenshot: 
Screenshot_20230326-235754

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -188,18 +188,18 @@ type SetLoginServerEvent struct {
 
 // UIEvent types.
 type (
-	ToggleEvent       struct{}
-	ReauthEvent       struct{}
-	BugEvent          struct{}
-	WebAuthEvent      struct{}
-	GoogleAuthEvent   struct{}
-	LogoutEvent       struct{}
-	OSSLicensesEvent  struct{}
-	BeExitNodeEvent   bool
-	ExitAllowLANEvent bool
+	ToggleEvent                    struct{}
+	ReauthEvent                    struct{}
+	BugEvent                       struct{}
+	WebAuthEvent                   struct{}
+	GoogleAuthEvent                struct{}
+	LogoutEvent                    struct{}
+	OSSLicensesEvent               struct{}
+	BeExitNodeEvent                bool
+	ExitAllowLANEvent              bool
 	AllowIncomingTransactionsEvent bool
-	UseTailscaleDNSEvent bool
-	UseTailscaleSubnetsEvent bool
+	UseTailscaleDNSEvent           bool
+	UseTailscaleSubnetsEvent       bool
 )
 
 // serverOAuthID is the OAuth ID of the tailscale-android server, used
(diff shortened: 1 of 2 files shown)


Aaroneisele55 (approved):

cmd/tailscale/main.go
```diff
@@ -434,6 +437,15 @@ func (a *App) runBackend() error {
 			case ExitAllowLANEvent:
 				state.Prefs.ExitNodeAllowLANAccess = bool(e)
 				go b.backend.SetPrefs(state.Prefs)
+			case UseTailscaleDNSEvent:
+				state.Prefs.CorpDNS = bool(e)
+				go b.backend.SetPrefs(state.Prefs)
+			case UseTailscaleSubnetsEvent:
+				state.Prefs.NoSNAT = !bool(e)
```
DentonGentry, line 444: This should be state.Prefs.RouteAll

NoSNAT is the preference for `tailscale up --snat-subnet-routes` on Linux.

cmd/tailscale/ui.go
```diff
@@ -1080,6 +1114,70 @@ func (ui *UI) layoutExitNodeDialog(gtx layout.Context, sysIns system.Insets, exi
...
+	}
+	d.dismiss.Add(gtx, argb(0x66000000))
+	layout.Inset{
+		Top:    sysIns.Top + unit.Dp(16),
+		Right:  sysIns.Right + unit.Dp(16),
+		Bottom: sysIns.Bottom + unit.Dp(16),
+		Left:   sysIns.Left + unit.Dp(16),
+	}.Layout(gtx, func(gtx C) D {
+		return layout.Center.Layout(gtx, func(gtx C) D {
+			gtx.Constraints.Min.X = gtx.Dp(unit.Dp(250))
```
DentonGentry, line 1134: Could this be made wide enough that "Allow Incoming Connections" doesn't wrap?

Screenshot_20230402-163534

cmd/tailscale/main.go
Empty2k12: Thanks for the input. I will adapt this ASAP!

cmd/tailscale/ui.go
Empty2k12: I am looking into it.

Empty2k12: I have made the box bigger. Sadly, I cannot test in your case, since on my Pixel 3XL it looked fine even before the fix. The 250dp measurement was used because it is used in multiple places, in the app. In the future, we should probably make the box a fraction of the device width, rather than hardcoding a value.

DentonGentry: The text still wraps on my Pixel-3a-XL, even with 300 pixel width, but i think we'll merge it as is and try to work with it in future commits.

DentonGentry (approved):

DentonGentry:
This came up in platform testing for the 1.40 release, that the preferences in the UI are always unchecked even when the actual preference is enabled by default. Users found this confusing.

We will likely revert this PR for 1.40, and try again for 1.42.

Blightwidow:
> We will likely revert this PR for 1.40, and try again for 1.42.

Any update on this as we are on 1.42 ?

PR #89, opened 2023-04-06
Add split tunneling

Myned:
Based on https://github.com/tailscale/tailscale-android/pull/56
Closes https://github.com/tailscale/tailscale/issues/6912

Huge thanks to @vquicksilver for the initial implementation.

This PR adds split tunneling to the Android application. An entry is added to the dropdown menu that displays a list of apps that can be checked/unchecked to allow/disallow that app to connect through the tunnel. All apps are by default allowed, except for the previously hardcoded apps such as Android Auto. Re-enabling Tailscale is required to apply the changes.

#### Major differences from https://github.com/tailscale/tailscale-android/pull/56
- Load app list on menu press instead of application start
- App list fits more items on-screen
- Logging outside of [main.go](https://github.com/Myned/tailscale-android/blob/add-split-tunneling/cmd/tailscale/main.go) removed
(other logging exists there, unsure of the extent that this feature should log)
- More apps are available to toggle (less filtering)

#### Known issues
- App names without whitespace do not wrap and are truncated (bug in Gio UI?)
- Duplicate app names are possible, but icons should help differentiate
- Icon generation causes stutter in app list when scrolling, mitigated by low icon resolution
- ~~As [noted](https://github.com/tailscale/tailscale-android/pull/56#discussion_r960149361), possible license concerns with certain snippets of code in [AppsConfig.java](https://github.com/Myned/tailscale-android/blob/add-split-tunneling/android/src/main/java/com/tailscale/ipn/AppsConfig.java)~~

#### Next steps
A [form](https://play.google.com/console/developers/app/app-content/permission-declarations), subject to approval, is required for submission to the Play Store, as [commented](https://github.com/tailscale/tailscale-android/pull/56#discussion_r1137963216).

Screenshot_1680779006


android/src/main/java/com/tailscale/ipn/AppsConfig.java
```diff
@@ -0,0 +1,246 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
```
DentonGentry, line 1: This file is the primary issue in the PR. It appeared to have been substantially copied from OpenVPN, and carries a GPL license.

There are only so many ways to use the Android APIs in this area and any code to use those APIs will have some resemblances. It needs to be developed afresh from the APIs, and not be GPL code.

vquicksilver: Sorry for the very late reply @Myned , I am not using tailscale anymore so my motivation for finishing this pull request was gone and on top of that all my Github emails/notifications were in the Spam folder. 

As far as I remember I wrote that class from scratch (minus the method I was pointing in my comment), so perhaps that's the only method that needs to be rewritten but probably @DentonGentry can give you more details since he did the initial review. Remember to also filter the tailscale app from the list of apps as I was pointing in my comment, I don't remember if I did that in the end or not.

Good luck!.

DentonGentry: Yes, `getInstalledApps()` appears to be the thing to focus on.

Myned:
Should mention that the [fix to loading apps](https://github.com/tailscale/tailscale-android/pull/89/commits/e815baf0a92373ddf51c468274ec5c9f818b763b) only applies to the UI, not the Android system query, since the query is currently required to build the VPN service on startup with split tunneling enabled.

cdmoss:
This is still a highly needed feature, really hoping more effort is put in. I would but it's beyond my expertise :P

589290:
"Allowed Apps" seems confusing to me... should it not instead be "Apps Allowed to Bypass VPN"? From there instead of starting with a list of all apps selected (ON) you would start with an empty list of all apps (OFF) and then proceed to select the individual apps that you want to grant the ability to bypass?

Myned:
Agreed there. I attempted to keep the original intentions of the PR author as allowed apps is how it's structured in the code, but it should be simple enough to reverse the list selection in the UI to make it more like how split tunneling is presented in other apps.

vquicksilver:
If you mark the option to route all the traffic through the VPN in the android settings and you toggle off an app in the list then that app is effectively forbidden to connect anywhere. This was my original use case (block proprietary apps that I cannot uninstall) and hence I called it that way.

If you don't mark that option in the android settings then it is "split tunneling" as you guys call it (but you also tell which apps are allowed to connect to the VPN and which ones are not allowed), so it is also an "allow list".



On September 4, 2023 2:23:21 PM UTC, Myned ***@***.***> wrote:
>Agreed there. I attempted to keep the original intentions of the PR author as allowed apps is how it's structured in the code, but it should be simple enough to reverse the list selection in the UI to make it more like how split tunneling is presented in other apps.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/tailscale/tailscale-android/pull/89#issuecomment-1705356571
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Myned:
Gotcha. As the change would only be reversing how the allowed list is displayed, it should keep that functionality, but also be reversed in the terminology (checking the box would mean it is disallowed to connect).

vquicksilver:
I don't think that is exactly correct... but I might have forgotten the exact details.

The correct name perhaps would be something like "routed apps" that would suit both use cases.

The setting that controls which use case (Routing everything vs allowing apps to connect without the VPN) is outside of tailscale, in the android settings, so if you change it from "allowed apps" to suit only your use case it would also be confusing.

If the setting is on ALL the traffic is routed through the VPN and tailscale (Android's ebpf to be more precise) acts as an outbound firewall, and app which checkbox is off cannot connect in this mode.

If the setting is off apps are allowed to both use or don't use the VPN (If the checkbox in tailscale's list is toggled on) and the other apps can use the regular connection (but in this case we have an outbound firewall to the VPN and they are not allowed iirc).

Hopefully that is clearer.


On September 4, 2023 3:52:08 PM UTC, Myned ***@***.***> wrote:
>Gotcha. As the change would only be reversing how the allowed list is displayed, it should keep that functionality, but also be reversed in the terminology (checking the box would mean it is disallowed to connect).
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/tailscale/tailscale-android/pull/89#issuecomment-1705475624
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Myned:
There's also the issue of the `Block connections without VPN` setting on Android blocking all apps from the network unless an exit node is connected, regardless of whether it is bypassing the tunnel, since Tailscale doesn't route everything by default. I don't know enough about Android networking yet to say whether this is fixable, but that setting seems to be out of the scope of this PR in general.

While I don't have a problem with keeping the functionality as a firewall, it is more of a side effect specifically for Android and not the main purpose of Tailscale which is as a mesh VPN, so the terminology should reflect that and align with other VPN apps, in my opinion. There are more apps that have "split tunneling" than "allowed/routed apps".

I was able to reverse the app list while keeping the same behavior in regards to the firewall, with the below mockup:

<details>
<summary>Before</summary>

Screenshot_1693855294

</details>

<details>
<summary>After</summary>

Screenshot_1693852711

</details>

expoli:
How is the development progress of this feature? I feel that this feature is very necessary. Why has there been no corresponding progress for such a long time? Have there been any issues encountered?

Myned:
At the time of the last commit, this PR was ready for review to the extent of the original implementation that was previously reviewed. I've been waiting for confirmation of whether there was anything else that needed to be modified before updating to the current master (the "Allowed Apps" vs "Bypass VPN" semantics was just a QoL change).

If using an older version is acceptable as a stopgap, building from my fork should still work fine, though I haven't tested it in a while.

jiz4oh:
Hi guys, I'm personally dying for this feature. Could it please get added asap?

agottardo:
https://github.com/tailscale/tailscale-android/pull/435 will finally add this functionality, targeting the current app as it was entirely redesigned since this PR was opened.

We hope to land it in the next app update.

Myned:
Thank you greatly for the rewrite, it was definitely a requirement after the UI changes. I'll close this PR as it is no longer applicable to the current version.

PR #87, opened 2023-04-02
Add intents

brettjenkins:
~~This is still a WIP, still testing, but created this draft so I can start to get the PR ready~~

Ready for review now.

PR #95, opened 2023-05-01
Prevent connecting loops when using connect intent more than once

brettjenkins:
Added a check to check if the VPN is connected before triggering the connect behaviour.

Was introduced in https://github.com/tailscale/tailscale-android/pull/87

Fixes: https://github.com/tailscale/tailscale/issues/8013

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.java
+++ b/android/src/main/java/com/tailscale/ipn/App.java
@@ -106,9 +106,6 @@ public class App extends Application {
 				NetworkInfo active = cMgr.getActiveNetworkInfo();
 				// https://developer.android.com/training/monitoring-device-state/connectivity-status-type
 				boolean isConnected = active != null && active.isConnectedOrConnecting();
-				if (isConnected) {
-					((App)getApplicationContext()).autoConnect = false;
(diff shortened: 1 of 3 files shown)


CrazyWolf13:
> Doing some more testing before I convert this into a full PR.

Huge thank you for creating this fix! 🔥❤️

anoninthekop:
Hi, when will this fix be available on the playstore. Now whenever I start tailscale with tasker I have to manually force close the app to disconnect.

plia7:
> Ready for review, @DentonGentry mind checking this over? Thanks :)

Hello, was this PR approved and pulled into the main release branch? It's been 2 months now, and we're eager for this fix.

Thank you.

plia7:
> Hi, when will this fix be available on the playstore. Now whenever I start tailscale with tasker I have to manually force close the app to disconnect.

@anoninthekop I got it working without the fix, all you need to do is to send the disconnect intent twice, just give like a 5 seconds wait after each send intent... works like a charm.

DentonGentry:
Tested locally.

DentonGentry (approved):

PR #96, opened 2023-05-03
Exit node in notification

RoboMagus:
This adds the currently used (if any) exit node to the persistent Tailscale notification.

I was going to open a new FR for this, but found that the request has been made before tailscale/tailscale#4642 (over a year ago) and (imo incorrectly) closed as a duplicate of tailscale/tailscale#4383. 

Below is a screen snippet of what this PR adds. Note that the exit node that I'm using here is actually called `tailscale` and may have been a bit poorly chosen for demonstration purposes.

Screenshot_20230503_cropped

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -473,6 +473,9 @@ func (a *App) runBackend() error {
 				go b.backend.SetPrefs(state.Prefs)
 				state.updateExitNodes()
 				a.notify(state)
+				if service != 0 {
+					a.updateNotification(service, state.State, state.ExitStatus, state.Exit)
+				}
 			}
 		case s := <-onConnect:
 			jni.Do(a.jvm, func(env *jni.Env) error {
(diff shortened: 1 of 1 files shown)


RoboMagus:
The notification is currently **not** updated when the user changes the use of an exit node. 
I'm currently looking into updating the notification when this happens too, but I'm open to input as to how this should be done propperly.

DentonGentry:
One thing: we do look for Signed-off-by lines in commit messages, the person is signing off that the code is suitable for submission using the open source license in use — which is the BSD 3-clause license for this codebase.

You could add a signed-off-by to the last commit in the sequence via:
1. `git commit --amend --signoff`
1. `git push -f`

I'll make sure to retain the Signed-off-by line when we squash the commits into one.

RoboMagus:
Thanks for the feedback @DentonGentry.
Latest commit should be propperly signed off as per your instructions.

PR #97, opened 2023-05-04
cmd/tailscale: use Google as DNS of last resort

bradfitz:
Sometimes we try a dozen different ways to read the phone's DNS
settings and it still comes back empty. In that case, if we're already
on a Google-ified Android phone, just use Google's Public DNS servers
as the ultimate fallback, as we already do on ChromeOS to work around
ChromeOS Android VpnBuilder bugs.

Updates tailscale/tailscale#8006 etc etc

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -426,7 +426,19 @@ func (b *backend) getPlatformDNSConfig() string {
 	return baseConfig
 }
 
-func (b *backend) getDNSBaseConfig() (dns.OSConfig, error) {
+func (b *backend) getDNSBaseConfig() (ret dns.OSConfig, _ error) {
+	defer func() {
+		// If we couldn't find any base nameservers, ultimately fall back to
+		// Google's. Normally Tailscale doesn't ever pick a default nameserver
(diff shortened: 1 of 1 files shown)


bradfitz:
Untested.

cmd/tailscale/backend.go
```diff
@@ -426,7 +426,21 @@ func (b *backend) getPlatformDNSConfig() string {
...
+		// If we couldn't find any base nameservers, ultimately fall back to
+		// Google's. Normally Tailscale doesn't ever pick a default nameserver
+		// for users but in this case Android's APIs for reading the underlying
+		// DNS config are lacking, and almost all Android phones use Google
+		// services anyway, so it's a reasonable default: it's an ecosystem the
+		// user has selected by having an Android device.
+		if len(ret.Nameservers) == 0 {
+			log.Printf("getDNSBaseConfig: none found; falling back to Google public DNS")
+			ret.Nameservers = append(ret.Nameservers,
+				netip.MustParseAddr("8.8.8.8"),
```
danderson, line 440: And 8.8.4.4 and the v6 equivalent?

danderson (approved):

PR #110, opened 2023-06-21
Dockerfile: update build tools in dockerfile

willnorris:
Use a slightly more current version of Java (jdk-11), as well as more recent versions of Android command line tools, build tools, and Go.

This was the dockerfile Charlotte used for the 1.44.0 release today, so I'm pretty confident it works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 # This is a Dockerfile for creating a build environment for
 # tailscale-android.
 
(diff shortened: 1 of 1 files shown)


DentonGentry:
Previous releases were also built using Java 11, definitely fine to update the Dockerfile.

```
$ javac -version
javac 11.0.19
```

DentonGentry (approved):

PR #111, opened 2023-06-28
cmd/tailscale: if first notification, set hostname even if prefs is nil

kari-ts:
Currently, if the notification (ipn.Notify) has a nil Prefs, the hostname defaults to localhost

Fixes #7875

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -351,13 +351,14 @@ func (a *App) runBackend() error {
 		case n := <-notifications:
 			exitWasOnline := state.ExitStatus == ExitOnline
 			if p := n.Prefs; p != nil && n.Prefs.Valid() {
-				first := state.Prefs == nil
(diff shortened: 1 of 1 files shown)


cmd/tailscale/main.go
```diff
@@ -351,13 +351,14 @@ func (a *App) runBackend() error {
...
 				state.updateExitNodes()
-				if first {
-					state.Prefs.Hostname = a.hostname()
-					go b.backend.SetPrefs(state.Prefs)
-				}
+				a.setPrefs(state.Prefs)
+			}
+			first := state.Prefs == nil
+			if first {
+				state.Prefs = &ipn.Prefs{Hostname: a.hostname()}
```
DentonGentry, line 360: Oh, interesting. So it ends up calling the Go `os.Hostname()` later.

DentonGentry (approved):

PR #112, opened 2023-07-11
android: update gradle version

kari-ts:
-Update Gradle version
-Replace jcenter() with mavenCentral() because of jcenter deprecation -Use testImplementation instead of the obsolete testCompile

Fixes #12997

Tested with docker


raggi (approved):
Works for me in a main shell too. Thanks!

PR #114, opened 2023-07-21
Makefile,Dockerfile,README.md: improve build dependency setup and documentation

raggi:
- Teach Makefile to install an Android SDK and the components we need
- Make Dockerfile delegate to Makefile for Android SDK setup
- Detect Android Studio and other known SDK paths, and use them if found.
- Update documentation to more consistently point to Android Studio & make.
- Add a task that checks for the SDK components and produces a useful error.
- Build an APK by default.
- Allow TOOLCHAINDIR to be passed in, and strip the first go/ component
  so that it is the path a user would expect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,15 +21,55 @@ TAILSCALE_COMMIT=$(shell echo $(TAILSCALE_VERSION) | cut -d - -f 2 | cut -d t -f
 # Extract the version code from build.gradle.
 VERSIONCODE=$(lastword $(shell grep versionCode android/build.gradle))
 VERSIONCODE_PLUSONE=$(shell expr $(VERSIONCODE) + 1)
+ifeq ($(shell uname),Linux)
+	ANDROID_TOOLS_URL="https://dl.google.com/android/repository/commandlinetools-linux-9477386_latest.zip"
+	ANDROID_TOOLS_SUM="bd1aa17c7ef10066949c88dc6c9c8d536be27f992a1f3b5a584f9bd2ba5646a0  commandlinetools-linux-9477386_latest.zip"
+else
+	ANDROID_TOOLS_URL="https://dl.google.com/android/repository/commandlinetools-mac-9477386_latest.zip"
+	ANDROID_TOOLS_SUM="2072ffce4f54cdc0e6d2074d2f381e7e579b7d63e915c220b96a7db95b2900ee  commandlinetools-mac-9477386_latest.zip"
+endif
+ANDROID_SDK_PACKAGES='platforms;android-31' 'extras;android;m2repository' 'ndk;23.1.7779620' 'platform-tools' 'build-tools;33.0.2'
 
-TOOLCHAINREV=$(shell go run tailscale.com/cmd/printdep --go)
-TOOLCHAINDIR=${HOME}/.cache/tailscale-android-go-$(TOOLCHAINREV)
-TOOLCHAINSUM=$(shell $(TOOLCHAINDIR)/go/bin/go version >/dev/null && echo "okay" || echo "bad")
-TOOLCHAINWANT=okay
-export PATH := $(TOOLCHAINDIR)/go/bin:$(PATH)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -21,15 +22,37 @@ TAILSCALE_COMMIT=$(shell echo $(TAILSCALE_VERSION) | cut -d - -f 2 | cut -d t -f
...
+export ANDROID_SDK_ROOT=$(PWD)/android-sdk
+endif
+export ANDROID_HOME ?= $(ANDROID_SDK_ROOT)
+
+# Set JAVA_HOME to the Android Studio bundled JDK.
+# TODO(raggi): what is the equivalent location on Linux?
+export JAVA_HOME ?= $(shell find /Applications/Android\ Studio.app/Contents/jre/Contents/Home -maxdepth 1 -type d 2>/dev/null | head -n 1)
 
-TOOLCHAINREV=$(shell go run tailscale.com/cmd/printdep --go)
+TOOLCHAINREV=$(shell go run tailscale.com/cmd/printdep@$(TAILSCALE_MOD_VERSION) --go)
```
DentonGentry, line 47: Changes here might break the F-Droid builder. We'll have to go find it, but I recall they patch the Makefile to remove the Tailscale toolchain.

I imagine we'd proceed to do this, but if we can find where it is we might file an issue explaining what they'll run into.

Makefile
```diff
@@ -41,41 +64,72 @@ bumposs: toolchain
 	GOPROXY=direct go get tailscale.com@main
 	go mod tidy -compat=1.20
 
-toolchain:
-ifneq ($(TOOLCHAINWANT),$(TOOLCHAINSUM))
-	@echo want: $(TOOLCHAINWANT)
-	@echo got: $(TOOLCHAINSUM)
+$(TOOLCHAINDIR)/go/bin/go:
```
DentonGentry, line 67: If I understand this correctly, removing the ifneq TOOLCHAINWANT,TOOLCHAINSUM means it will remove ${HOME}/.cache/tailscale-android-go-* and re-download it on every build?

One problem we've had is that GitHub expires the build artifacts more rapidly than we update the Android toolchain checksum. I'm concerned that if we expect to re-download each time, it will only work for a couple weeks and then Android builds begin failing.

Makefile
raggi: Is this comment referring to a generality of changing lines that will cause a patch hunk to fail to apply, or something specific about the change?

Makefile
raggi: It won't download it each time - currently our android toolchain builds contain version information in the path, i.e. `~/.cache/tailscale-android-go-40dc4d834a5fde9872bcf470be50069f56c3e3b3/` so this will only fetch the go toolchain if that path does not exist.

If we need to override versions in the future we can produce a stamp file that is a prerequisite and have the conditional update the stamp, which would retrigger the action.

raggi: I think we need to address the issue you describe here with our expiring toolchains separately. I'll make a bug for that.

raggi: tailscale/corp#13478

Makefile
DentonGentry: I believe this PR will break the FDroid build https://gitlab.com/fdroid/fdroiddata/-/commit/e0b078c0db72c959a73caedbb14f6ad54f1a783a

However I don't see a way to avoid breaking the FDroid build. A sed done on our Makefile will break when we change the Makefile.

Therefore I think we proceed, and we help get FDroid updated once submitted.

raggi: Yeah, agreed. I think perhaps f-droid might want a clean(er) way to override the toolchain behaviors in the makefile.

DentonGentry (approved):

Makefile
raggi: It was possible to clean this up some more relatively easily, as the old want/check dance was not in use in the new arrangement. Now the user can simply pass a custom TOOLCHAINDIR, which must contain a `bin/go` or we'll attempt to download our preferred version.

PR #115, opened 2023-07-25
cmd/tailscale/main.go: use NewPrefs(), not an empty struct.

DentonGentry:
NewPrefs() enables accepting DNS and routes, among other things. Fixes https://github.com/tailscale/support/issues/12


bradfitz (commented):
Which empty struct?

DentonGentry:
Updated commit message to state it more clearly.

agottardo (approved):

PR #126, opened 2023-08-11
cmd/tailscale: rebind magicsock.Conn onConnect

catzkorn:
We have been getting into routing loops due to the timing of when we bind sockets on starting the tailscale app. At this point, we do not have access to `VpnService.protect()` and are unable to protect the magicsock sockets, which causes a routing loop issue until we forcibly rebind about 10 minutes into the service being started.

This change rebinds when we detect that the protect function is no longer nil, and restores connectivity.

Updates tailscale/corp#13814

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -498,6 +498,17 @@ func (a *App) runBackend() error {
 						return nil // even on error. see big TODO above.
 					})
 				})
+				log.Printf("onConnect: rebind required")
+				// TODO(catzkorn): When we start the android application
+				// we bind sockets before we have access to the VpnService.protect()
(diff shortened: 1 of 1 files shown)


cmd/tailscale/main.go
```diff
@@ -498,6 +498,18 @@ func (a *App) runBackend() error {
 						return nil // even on error. see big TODO above.
 					})
 				})
+				if !hadPrevProtect {
```
raggi, line 501: We should always DebugRebind, the condition here is slightly fiddly and the "had" does not provide us direct insight into / relationship with when the backend last bound its sockets.

catzkorn: Agreed and updated

raggi (approved):

PR #130, opened 2023-09-06
go.mod: bump oss

maisem:


raggi (approved):

android/build.gradle
```diff
@@ -26,8 +26,8 @@ android {
 	defaultConfig {
 		minSdkVersion 22
 		targetSdkVersion 33
-		versionCode 175
-		versionName "1.47.108-td5ac18d2c-g4b67f47e88f"
+		versionCode 176
+		versionName "1.49.211-tc2241248c-gc73f8533f0a"
```
maisem, line 30: I think i screwed this up, going to revert this and do the bump in a separate PR

cmd/tailscale/main.go
```diff
@@ -454,7 +454,7 @@ func (a *App) runBackend() error {
 					}
 				}()
 			case LogoutEvent:
-				go b.backend.Logout()
+				go b.backend.Logout(context.Background())
```
bradfitz, line 457: don't want to bound this at all?

Maybe the LocalBackend.Logout itself should set some upper bound (like N minutes) to avoid leaks in crazy situations where either the EditPrefs or ControlClient.Logout blocks forever for some reason?

bradfitz (approved):

cmd/tailscale/main.go
maisem: bounded it to 5 minutes here

bradfitz (approved):

PR #132, opened 2023-09-12
cmd/tailscale: avoid creating multiple netmon instances

raggi:
This is a quick fix, we should come back and try to reorganize this later.

Updates tailscale/tailscale#9374

### DIFF
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -46,6 +46,7 @@ type backend struct {
 	settings   settingsFunc
 	lastCfg    *router.Config
 	lastDNSCfg *dns.OSConfig
+	netMon     *netmon.Monitor
 
 	logIDPublic string
 	logger      *logtail.Logger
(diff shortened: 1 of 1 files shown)


cmd/tailscale/backend.go
```diff
@@ -136,6 +137,9 @@ func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateSto
 		DNS:          cb,
 		Dialer:       dialer,
 		SetSubsystem: sys.Set,
+		// TODO(raggi): see note in SetupLogs, this is curently awkward. Ensure
+		// that SetupLogs is called first.
+		NetMon: b.netMon,
```
bradfitz, line 142: how about panicking just before this if b.netMon is nil?

cmd/tailscale/backend.go
```diff
@@ -333,6 +337,9 @@ func (b *backend) SetupLogs(logDir string, logID logid.PrivateID) {
 	if err != nil {
 		log.Printf("netmon.New: %w", err)
 	}
+	// TODO(raggi): there's an awkward circular dependency here, between logf,
+	// netMon and the log transport. This should all get decoupled.
```
bradfitz, line 341: I don't follow but if this comment helps you remember, that's fine.

bradfitz (approved):

bradfitz (approved):
Merge away.

PR #136, opened 2023-09-26
Build fixes

dblohm7:
These changes pick up the latest OSS and fixes the build.

I've split it into two commits so that I can `cherry-pick` the breakage fix onto 1.50.0.


DentonGentry (approved):
It is fine to do this, however we should not proceed to build 1.50.0 yet (https://github.com/tailscale/corp/issues/14710#issuecomment-1728565874). We need another week to fix things for Android.

PR #140, opened 2023-10-25
cmd/tailscale: add tun Wrapper and call Start()

DentonGentry:
Fixes https://github.com/tailscale/corp/issues/15388


awly (commented):
Looks reasonable, but I have very little context on this machinery :|

DentonGentry:
Verified that a tailscale-debug.apk built with this PR is able to start and run.

bradfitz (approved):

PR #147, opened 2024-01-12
use network callback to update DNS config when network changes

kari-ts:
-Use requestNetwork, which gets the best network matching the passed in network request, to listen for changes to network and cache DNS config 
-Call netmon.InjectEvent on network change to indicate a change 
Follow-up will fix issue in netmon where IsMajorChangeFrom doesn't identify major changes when a network is added


android/src/main/java/com/tailscale/ipn/DnsConfig.java
```diff
@@ -35,31 +38,13 @@
...
-// - If Wi-Fi is active we use that.
-// - If LTE is active we use that.
-// - We never use a VPN's DNS servers. That VPN is likely us. Even if not us, Android
-// only allows one VPN at a time so a different VPN's DNS servers won't be available
-// once Tailscale comes up.
-//
-// getAllNetworks() is used as the sole mechanism for retrieving the DNS config with
-// Android 7 and later.
+// ## getDnsConfigs
+// Android provides a requestNetwork interface in the ConnectivityManager. This returns the best
```
danderson, line 42: According to android docs, requestNetwork was added in API level 21, which is Android 5. So, I think we can actually replace all of this, for all versions, with the new code!

android/src/main/java/com/tailscale/ipn/DnsConfig.java
```diff
@@ -120,80 +107,47 @@ String getDnsConfigAsString() {
...
-			}
-
-			LinkProperties linkProp = cMgr.getLinkProperties(network);
-			NetworkCapabilities nc = cMgr.getNetworkCapabilities(network);
-			List<InetAddress> dnsList = linkProp.getDnsServers();
-			StringBuilder sb = new StringBuilder("");
-			for (InetAddress ip : dnsList) {
-				sb.append(ip.getHostAddress() + " ");
-			}
+	public String getDnsConfigs(){
```
danderson, line 110: I haven't done java in for ever, but I think you could make the locking more idiomatic by using `synchronized` ? Instead of the try/finally blocks and the explicit lock you should be able to write:

```
public String getDnsConfigs() {
  synchronized(this) {
    return this.dnsConfigs;
  }
}
```

And same for the place that updates this.dnsConfigs

cmd/tailscale/backend.go
```diff
@@ -438,6 +438,7 @@ func (b *backend) getPlatformDNSConfig() string {
 		log.Printf("getPlatformDNSConfig JNI: %v", err)
 		return ""
 	}
+	log.Printf("Testing 10107: %s", baseConfig)
```
danderson, line 441: Remove leftover debug logging?

android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -87,40 +87,28 @@ public class App extends Application {
...
-
-			@Override
-			public void onLinkPropertiesChanged(Network network, LinkProperties linkProperties) {
-				super.onLinkPropertiesChanged(network, linkProperties);
-				this.reportConnectivityChange();
-			}
-		});
+	private void setAndRegisterNetworkCallbacks() {
+		dnsCallback = dns.getDnsConfigCallback(connectivityManager);
+		connectivityManager.requestNetwork(dns.getDNSConfigNetworkRequest(), dnsCallback);
```
danderson, line 111: I'm wondering, now that the new callback has replaced the old one: should it live here on the App object and just push an update into the DnsConfig object to capture new DNS settings? It's a bit odd to have a callback on DnsConfig actually be responsible for all network change notifications, not just DNS.

So, like, have an UpdateDNSFromNetwork(Network net) method on DnsConfig, move the networkrequest callback back to App, and have the callback in App call UpdateDNSFromNetwork as needed, before calling the native function?

cmd/tailscale/callbacks.go
```diff
@@ -42,6 +39,9 @@ var (
 
 	// onWriteStorageGranted is notified when we are granted WRITE_STORAGE_PERMISSION.
 	onWriteStorageGranted = make(chan struct{}, 1)
+
+	// onDNSConfigChanged is notified when the network changes and the DNS config needs to be updated.
+	onDNSConfigChanged = make(chan struct{}, 1)
```
danderson, line 44: The new callback no longer plumbs through the bool for online/offline notifications. Do we still need that, or am I missing something?

danderson (commented):
Looks good in general, a couple of suggestions for code organization and stuff.

danderson:
Oh, and you need to make the commit say `Updates tailscale/tailscale#10107` for the issue linking to work, since it's in another repository

cmd/tailscale/callbacks.go
kari-ts: I actually don't think we ever needed that? We never checked what the bool was, we just did stuff when there was an update.

android/src/main/java/com/tailscale/ipn/DnsConfig.java
kari-ts: Yay!

android/src/main/java/com/tailscale/ipn/DnsConfig.java
kari-ts: Done.

android/src/main/java/com/tailscale/ipn/App.java
kari-ts: Done.

cmd/tailscale/backend.go
kari-ts: Done.

android/src/main/java/com/tailscale/ipn/DnsConfig.java
```diff
@@ -102,260 +42,22 @@ public DnsConfig(Context ctx) {
...
-		if (android.os.Build.VERSION.SDK_INT >= 23) {
-			// If ConnectivityManager.getAllNetworks() works, it is the
-			// authoritative mechanism and we rely on it. The other methods
-			// which follow involve more compromises.
-			return "";
-		}
-
-		s = getDnsServersFromSystemProperties();
+		String s = getDnsConfigs();
 		if (!s.trim().isEmpty()) {
```
danderson, line 46: unnecessary? We're now returning empty string regardless

danderson (approved):
Looks right to me! One small change suggestion but no need for another round of review.

android/src/main/java/com/tailscale/ipn/DnsConfig.java
kari-ts: We return the string if it's nonempty

danderson: Sure, but if the string is empty, we return `""` below. So, you could skip the conditional and always return `getDnsConfigs().trim()` for the same effect

PR #150, opened 2024-01-18
cmd/tailscale: remove obsolete DNS config logging

danderson:
Causes a JNI crash because 1b42117791266251b41959fc18e6c6215a85f739 removed those methods when fixing our connectivity monitoring.


kari-ts (approved):
Thank you!

PR #151, opened 2024-01-18
Record DNS search domains as well as nameservers.

danderson:
We accidentally removed this in the big connectivity monitor fix.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.java
+++ b/android/src/main/java/com/tailscale/ipn/App.java
@@ -115,6 +115,12 @@ public class App extends Application {
 				for (InetAddress ip : dnsList) {
 					sb.append(ip.getHostAddress()).append(" ");
 				}
+				String searchDomains = linkProperties.getDomains();
+				if (searchDomains != null) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/App.java
```diff
@@ -115,6 +115,13 @@ public void onAvailable(Network network){
 				for (InetAddress ip : dnsList) {
 					sb.append(ip.getHostAddress()).append(" ");
 				}
+				String searchDomains = linkProperties.getDomains();
+				if (searchDomains != null) {
+					sb.append("\n");
+					sb.append(searchDomains);
+				}
+				android.util.Log.d("X", sb.toString());
```
kari-ts, line 123: Do you want to keep this?

kari-ts (approved):

android/src/main/java/com/tailscale/ipn/App.java
danderson: Oops, no I don't!

PR #152, opened 2024-01-19
android: bump version code

kari-ts:


android/build.gradle
```diff
@@ -26,8 +26,8 @@ android {
 	defaultConfig {
 		minSdkVersion 22
 		targetSdkVersion 33
-		versionCode 194
-		versionName "1.57.72-tca48db0d6-gdcca09fe7f8"
+		versionCode 195
+		versionName "-g47a7d5b9ad1"
```
DentonGentry, line 30: The versionname doesn't look correct, I'd expect `1.59-something-g47a7d5b9ad1`

DentonGentry (changes requested):

android/build.gradle
kari-ts: Thanks for flagging. I pulled from HEAD and reran.

android/build.gradle
```diff
@@ -26,8 +26,8 @@ android {
 	defaultConfig {
 		minSdkVersion 22
 		targetSdkVersion 33
-		versionCode 194
-		versionName "1.57.72-tca48db0d6-gdcca09fe7f8"
+		versionCode 195
+		versionName "1.57.100-t8250582fe-g4799cc1ef"
```
DentonGentry, line 30: Because OSS still has VERSION.txt of 1.57, this is correct.

If at all possible, please get through the step in the client release process to update VERSION.txt in OSS to 1.59 before the end of the day today. Right now commits going in to OSS and corp are being misattributed as being part of 1.58, because they are still in the 1.57.x dev cycle.

DentonGentry (approved):

PR #154, opened 2024-01-24
android: bump version code and update OSS

kari-ts:


android/build.gradle
```diff
@@ -26,8 +26,8 @@ android {
 	defaultConfig {
 		minSdkVersion 22
 		targetSdkVersion 33
-		versionCode 195
-		versionName "1.57.100-t8250582fe-g4799cc1ef"
+		versionCode 196
+		versionName "1.58.2-tb0e1bbb62-gcd1f0649958"
```
DentonGentry, line 30: Because this is a merge from the release branch, it is proposing to update the version to 1.58.2. `main` cannot build 1.58.

I would suggest abandoning this PR and instead:
- manually increment versionCode to 196, to avoid conflicting with the release branch
- use `make bumposs` to import from `HEAD` of OSS.

PR #155, opened 2024-01-31
cmd/tailscale/main: use localapi for generating bug report

kari-ts:
-Create a custom http.ResponseWriter whose responses can be read and in the future, sent to Java/Kotlin via JNI
-Make http request and call localapi directly

The purpose of this change is to test out swapping an existing feature's implementation for the new pattern of using localapi. I added a global channel for the bug report in order to contain the scope of this change to just this. 

Updates #10992

### DIFF
--- a/cmd/localapiclient/localapiclient.go
+++ b/cmd/localapiclient/localapiclient.go
@@ -1,0 +1,102 @@
+package localapiclient
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"sync"
+
+	"tailscale.com/ipn/localapi"
+)
+
+// Response represents the result of processing an http.Request.
+type Response struct {
+	headers              http.Header
+	status               int
+	bodyWriter           net.Conn
+	bodyReader           net.Conn
+	startWritingBody     chan interface{}
+	startWritingBodyOnce sync.Once
+}
+
+func (r *Response) Header() http.Header {
+	return r.headers
+}
+
+// Write writes the data to the response body and will send the data to Java.
+func (r *Response) Write(data []byte) (int, error) {
+	r.Flush()
+	if r.status == 0 {
+		r.WriteHeader(http.StatusOK)
+	}
+	return r.bodyWriter.Write(data)
+}
+
+func (r *Response) WriteHeader(statusCode int) {
+	r.status = statusCode
+}
+
+func (r *Response) Body() net.Conn {
+	return r.bodyReader
+}
+
+func (r *Response) StatusCode() int {
+	return r.status
+}
+
+func (r *Response) Flush() {
+	r.startWritingBodyOnce.Do(func() {
+		close(r.startWritingBody)
+	})
+}
+
+type LocalAPIClient struct {
+	h *localapi.Handler
+}
+
+func New(h *localapi.Handler) *LocalAPIClient {
+	return &LocalAPIClient{h: h}
+}
+
+// Call calls the given endpoint on the local API using the given HTTP method
+// optionally sending the given body. It returns a Response representing the
+// result of the call and an error if the call could not be completed or the
+// local API returned a status code in the 400 series or greater.
+// Note - Response includes a response body available from the Body method, it
+// is the caller's responsibility to close this.
+func (cl *LocalAPIClient) Call(ctx context.Context, method, endpoint string, body io.Reader) (*Response, error) {
+	req, err := http.NewRequestWithContext(ctx, method, "/localapi/v0/"+endpoint, body)
+	if err != nil {
+		return nil, fmt.Errorf("error creating new request for %s: %w", endpoint, err)
+	}
+	deadline, _ := ctx.Deadline()
+	pipeReader, pipeWriter := net.Pipe()
+	pipeReader.SetDeadline(deadline)
+	pipeWriter.SetDeadline(deadline)
+
+	resp := &Response{
+		headers:          http.Header{},
+		status:           http.StatusOK,
+		bodyReader:       pipeReader,
+		bodyWriter:       pipeWriter,
+		startWritingBody: make(chan interface{}),
+	}
+
+	go func() {
+		cl.h.ServeHTTP(resp, req)
+		resp.Flush()
+		pipeWriter.Close()
+	}()
+
+	select {
+	case <-resp.startWritingBody:
+		if resp.StatusCode() >= 400 {
+			return resp, fmt.Errorf("request failed with status code %d", resp.StatusCode())
+		}
+		return resp, nil
+	case <-ctx.Done():
+		return nil, fmt.Errorf("timeout for %s", endpoint)
+	}
+}
--- a/cmd/tailscale/backend.go
+++ b/cmd/tailscale/backend.go
@@ -48,7 +48,7 @@ type backend struct {
 	lastDNSCfg *dns.OSConfig
 	netMon     *netmon.Monitor
 
-	logIDPublic string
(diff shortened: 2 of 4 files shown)


cmd/tailscale/backend.go
```diff
@@ -150,7 +150,7 @@ func newBackend(dataDir string, jvm *jni.JVM, appCtx jni.Object, store *stateSto
 		return nil, fmt.Errorf("runBackend: NewUserspaceEngine: %v", err)
 	}
 	sys.Set(engine)
-	b.logIDPublic = logID.Public().String()
```
oxtoacart, line 153: If you want to avoid dealing with casing and such, you could convert logIDPublic to an `atomic.Pointer` like [this](https://go.dev/play/p/E5cMWQal4wh).

cmd/tailscale/main.go
```diff
@@ -425,6 +473,31 @@ func (a *App) runBackend() error {
...
+				req, err := http.NewRequest("POST", "/localapi/v0/bugreport", nil)
+				if err != nil {
+					log.Printf("error creating new request for bug report: %v", err)
+					return err
+				}
+				w := newLocalApiResponseWriter()
+				h.ServeHTTP(w, req)
+				if w.StatusCode() > 300 {
+					log.Printf("bug report bad http status: %v", w.StatusCode())
+					a.bugReport <- fallbackLog
```
oxtoacart, line 489: I see that we read from a.bugReport in response to a UI `BugEvent`. Is it ever possible that we'll submit multiple bugs to the `a.bugReport` channel without getting corresponding UI `BugEvent`s to read them? If so, this could end up blocking because the channel is full. Should we have a timeout or something here and drop the log if we can't write to the channel?

oxtoacart (commented):
Generally LGTM, just a concern about blocking on sending to that new channel, and a small potential improvement on `logIDPublic`.

cmd/tailscale/main.go
```diff
@@ -425,6 +473,31 @@ func (a *App) runBackend() error {
 			}
 		case e := <-backendEvents:
 			switch e := e.(type) {
+			case BugEvent:
+				backendLogID, _ := a.logIDPublicAtomic.Load().(string)
```
barnstar, line 477: The API logic should probably all be in it's own self-contained module separate from the backend event handler.  This main.go is already ummm... large.

Now's probably the time to start lumping all of that together.  Don't know exactly how that looks.  Probably in interface and an implementation that defines a method for each localAPI thing you invoke, takes a channel as an argment and parses and writes the response to it directly.

something like

```
...

result := make( chan string) 
err := LocalAPI.GetBugReportID(result); err != nil {
   return err
}

reportId := <-result
a.bugReport <- reportId

...

type LocalAPI interface {
    func (a *localAPI) GetBugReportID(chan result)
}

func (a *LocalAPI) GetBugReportID(chan result) {
     // Do non-web-things things
     result <- reportId
}


```

Maybe even pass the result channel itself around, and/or return it from the API... 

I barely have my head above water when it comes to golang concurrency (let along go in general) so take this all with a grain of salt (except the "put this in it's own module" part).

cmd/tailscale/main.go
```diff
@@ -65,6 +67,47 @@ type App struct {
...
+}
+
+// LocalApiResponseWriter is our custom implementation of http.ResponseWriter
+type LocalApiResponseWriter struct {
+	headers http.Header
+	body    bytes.Buffer
+	status  int
+}
+
+// NewLocalApiResponseWriter creates and initializes a new LocalApiResponseWriter
```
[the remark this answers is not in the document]
kari-ts: Agreed.

cmd/tailscale/main.go
kari-ts: Created a new package for localapi logic, wdyt?

cmd/tailscale/main.go
kari-ts: I don't think that is possible - we only get a backend BugEvent when there is a UI BugEvent, and when there's a UI BugEvent we always clear the channel before sending the backend BugEvent

kari-ts (commented):
Please disregard middle commit.

cmd/tailscale/main.go
```diff
@@ -280,9 +286,14 @@ func (a *App) runBackend() error {
 	if err != nil {
 		return err
 	}
-	a.logIDPublicAtomic.Store(b.logIDPublic)
+	a.logIDPublicAtomic = b.logIDPublic
```
oxtoacart, line 289: If you assign to this, it's no longer functioning as an atomic. If assignment to this needs to be atomic, you should use `a.logIDPublicAtomic.Store()`.

cmd/tailscale/backend.go
```diff
@@ -48,7 +49,7 @@ type backend struct {
 	lastDNSCfg *dns.OSConfig
 	netMon     *netmon.Monitor
 
-	logIDPublic string
+	logIDPublic atomic.Pointer[logid.PublicID]
```
oxtoacart, line 52: It looks like we only assign to this once on startup, so this doesn't need to be atomic.

cmd/tailscale/backend.go
oxtoacart: My bad, I meant convert just the logIDPublicAtomic on *App from atomoic.Value to atomic.Pointer.

kari-ts: Done.

cmd/tailscale/main.go
kari-ts: Done.

cmd/tailscale/backend.go
kari-ts: Done.

oxtoacart:
@kari-ts I find it easiest to work on the code directly, so I'm submitting a batch of suggestions in the form of a [DRAFT PR](https://github.com/tailscale/tailscale-android/pull/160).

barnstar (approved):

cmd/localapiclient/localapiclient.go
```diff
@@ -0,0 +1,105 @@
...
+		status:           http.StatusOK,
+		bodyReader:       pipeReader,
+		bodyWriter:       pipeWriter,
+		startWritingBody: make(chan interface{}),
+	}
+
+	go func() {
+		cl.h.ServeHTTP(resp, req)
+		resp.Flush()
+		pipeWriter.Close()
```
oxtoacart, line 93: Good catch.

oxtoacart, line 93: Good catch.

cmd/localapiclient/localapiclient.go
```diff
@@ -0,0 +1,105 @@
...
+	return &LocalAPIClient{h: h}
+}
+
+// Call calls the given endpoint on the local API using the given HTTP method
+// optionally sending the given body. It returns a Response representing the
+// result of the call and an error if the call could not be completed or the
+// local API returned a status code in the 400 series or greater.
+// Note - Response includes a response body available from the Body method, it
+// is the caller's responsibility to close this.
+func (cl *LocalAPIClient) Call(ctx context.Context, deadline time.Time, method, endpoint string, body io.Reader) (*Response, error) {
```
oxtoacart, line 70: The idiomatic way to handle this is to have the caller set a deadline on the context instead of passing it into here. If you need to get the context's deadline, you can do this

```
deadline, hasDeadline := ctx.Deadline()
```

oxtoacart (approved):
LGTM functionally, just one stylistic comment about not passing in the deadline to `Call()`.

PR #157, opened 2024-02-06
cmd/localapiclient: use localapi to login, and add tests

kari-ts:


localapiclient/localapiclient.go
```diff
@@ -0,0 +1,81 @@
+package localapiclient
```
barnstar, line 1: It looks like you've defined LocalAPIResponseWriter twice.  Is that necessary?  (seems obtuse, but maybe go requires it)  Should this just be "type LocalApi interface" and it's implementation?

kari-ts: Oops. This was leftover from a previous attempt. I'm bad at git.

cmd/localapiclient/localapiclient_test.go
```diff
@@ -0,0 +1,73 @@
...
+type TimeoutHandler struct{}
+
+var successfulReport = "successful report!"
+
+func (b *TimeoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	time.Sleep(6 * time.Second)
+	w.Write([]byte(successfulReport))
+}
+
+func TestBugTimeout(t *testing.T) {
```
oxtoacart, line 40: Is "Bug" left over from when the `localapiclient` package still had bug reporting functionality in it?

cmd/localapiclient/localapiclient_test.go
```diff
@@ -0,0 +1,73 @@
...
+	_, err := client.Call(ctx, "POST", "test", nil)
+
+	if err.Error() != "request failed with status code 400" {
+		t.Error("Expected bad status error, but got", err)
+	}
+}
+
+type TimeoutHandler struct{}
+
+var successfulReport = "successful report!"
```
oxtoacart, line 33: Is "report" left over from when the `localapiclient` package still had bug reporting functionality in it?

cmd/tailscale/main.go
```diff
@@ -591,6 +591,18 @@ func (a *App) getBugReportID(ctx context.Context, bugReportChan chan<- string, f
...
 
+func (a *App) login(ctx context.Context, errHandler func()) {
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(2*time.Second))
+	defer cancel()
+	r, err := a.localAPIClient.Call(ctx, "POST", "login-interactive", nil)
+	defer r.Body().Close()
+
+	if err != nil {
+		log.Printf("login: %s", err)
+		errHandler()
```
oxtoacart, line 602: Instead of having to pass in an error handler here (which is a bit unidiomatic in Go), you could just store the backend on the app itself and then call `a.b.StartLoginInteractive()` here.

oxtoacart (commented):
Just a couple of questions and 1 suggestion.

cmd/localapiclient/localapiclient_test.go
kari-ts: Renaming.

cmd/localapiclient/localapiclient_test.go
kari-ts: Yup, thanks for pointing this out!

cmd/tailscale/main.go
kari-ts: Sounds good

oxtoacart (approved):
LGTM

PR #158, opened 2024-02-08
cmd/tailscale: do not show location based exit nodes in main view

catzkorn:
This change stops us from clogging up the main UI view with location based exit nodes, which can be in their hundreds. They will still appear in the exit node UI.
Screenshot from 2024-02-07 16-34-42
Screenshot from 2024-02-07 16-34-34

Updates tailscale/tailscale#9438

### DIFF
--- a/cmd/tailscale/ui.go
+++ b/cmd/tailscale/ui.go
@@ -560,6 +560,12 @@ func (ui *UI) layout(gtx layout.Context, sysIns system.Insets, state *clientStat
 							ui.showCopied(gtx, a)
 						}
 					}
+
+					if p.Peer.Hostinfo.Location() != nil && p.Peer.IsWireGuardOnly {
+						// If the peer has location information set and is a wireguard
+						// only peer then it should not be displayed in the main list.
+						return D{}
+					}
 					return ui.layoutPeer(gtx, sysIns, p, userID, clk)
 				}
 			}


cmd/tailscale/ui.go
```diff
@@ -560,6 +560,12 @@ func (ui *UI) layout(gtx layout.Context, sysIns system.Insets, state *clientStat
 							ui.showCopied(gtx, a)
 						}
 					}
+
+					if p.Peer.Hostinfo.Location() != nil {
```
kari-ts, line 564: Would we ever set location for some reason other than it being a location based exit node?

kari-ts (commented):
Just curious about the guarantees of the location field, but otherwise LGTM

cmd/tailscale/ui.go
catzkorn: Not currently, no - it is only implemented for one specific type of node currently.

kari-ts: We should either guarantee that it will never be used for anything else, or add a comment to change the filtering in the clients if we ever do

catzkorn: Added the isWireguardOnly check, that should give us some room to breathe

kari-ts (approved):
Thanks!

PR #159, opened 2024-02-08
cmd/tailscale: improve exit node menu for location based exit nodes

catzkorn:
This change provides minor improvements to the exit node menu when there are location based exit nodes present. It will ensure that non location based exit nodes are displayed at the top of the list, followed by a the best node for a country/city combination, and followed by all location based exit nodes.

Updates tailscale/tailscale#9421

Screenshot from 2024-02-08 10-27-50
Screenshot from 2024-02-08 10-30-25

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"cmp"
 	"context"
 	"crypto/rand"
 	"crypto/sha1"
@@ -31,6 +32,7 @@ import (
 	"gioui.org/io/system"
 	"gioui.org/layout"
 	"gioui.org/op"
+	"golang.org/x/exp/maps"
 	"inet.af/netaddr"
 
 	"github.com/tailscale/tailscale-android/jni"
@@ -135,9 +137,11 @@ const (
 )
 
 type Peer struct {
-	Label  string
-	Online bool
-	ID     tailcfg.StableNodeID
+	Label             string
+	Online            bool
+	ID                tailcfg.StableNodeID
+	Location          *tailcfg.Location
+	PreferredExitNode bool
 }
 
 type BackendState struct {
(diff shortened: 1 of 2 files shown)


cmd/tailscale/main.go
```diff
@@ -675,10 +679,19 @@ func (s *BackendState) updateExitNodes() {
 		myExit := p.StableID() == exitID
 		hasMyExit = hasMyExit || myExit
 		exit := Peer{
-			Label:  p.DisplayName(true),
-			Online: canRoute,
-			ID:     p.StableID(),
+			Label:    p.DisplayName(true),
+			Online:   canRoute,
+			ID:       p.StableID(),
+			Location: p.Hostinfo().Location(),
```
kari-ts, line 685: Right now we call updateExitNodes under three circumstances: 
-the prefs change
-the netmap changes
-the user wants to use an exit node
Is there any case where the location for an exit node might change, but none of the three cases above is triggered? Or is it ok for the location to be stale because we never really use it anyway except for filtering/sorting exit nodes for display?

cmd/tailscale/main.go
```diff
@@ -689,9 +702,80 @@ func (s *BackendState) updateExitNodes() {
 			s.Exits = append(s.Exits, exit)
 		}
 	}
+
```
kari-ts, line 705: Should the sorting happen here, or in ui.go? (I'm not opposed to the logic being here)

catzkorn: The current code sorts here, and changing it into ui.go would cause issues with rendering (ask me how I know 😆)

cmd/tailscale/main.go
catzkorn: Nope - The location is sent from control, so the netmap will have to be resent in that case.

kari-ts (approved):
Thanks for the improvement!

PR #160, opened 2024-02-08
Code review suggestions

oxtoacart:
This is not intended to be merged (unless you want to), it's just an easier way for me to provide suggestions.


cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
 )
 
-// LocalAPIResponseWriter substitutes for http.ResponseWriter in order to write byte streams directly
-// to a receiver function in the application.
-type LocalApiResponseWriter struct {
-	headers http.Header
-	body    bytes.Buffer
-	status  int
+// Response represents the result of processing an http.Request.
+type Response struct {
```
oxtoacart, line 14: I think the consumer of the `localapiclient` package doesn't necessarily care that this is a fake ResponseWriter, they just want to able to use it as a `Response` that gives them the info they need. Thus the suggested name and comment change.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
+	headers    http.Header
+	status     int
+	bodyWriter io.WriteCloser
+	bodyReader io.ReadCloser
 }
 
-func newLocalApiResponseWriter() *LocalApiResponseWriter {
-	return &LocalApiResponseWriter{headers: http.Header{}, status: http.StatusOK}
```
oxtoacart, line 23: Since this is only used once, we could inline it at the point where we use it.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
 }
 
-func (w *LocalApiResponseWriter) StatusCode() int {
-	return w.status
+func (r *Response) StatusCode() int {
+	return r.status
 }
 
-type LocalApiClient struct {
+type LocalAPIClient struct {
```
oxtoacart, line 46: Go convention is to capitalize all of the acronym, e.g. `Api` -> `API`.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
 }
 
-var ErrBadHttpStatus = errors.New("bad http status for localapi response")
+// Call calls the given endpoint on the local API using the given HTTP method
+// optionally sending the given body. It returns a Response representing the
+// result of the call and an error if the call could not be completed or the
+// local API returned a status code in the 400 series or greater.
+// Note - Response includes a response body available from the Body method, it
+// is the caller's responsibility to close this.
+func (cl *LocalAPIClient) Call(ctx context.Context, method, endpoint string, body io.Reader) (*Response, error) {
```
oxtoacart, line 60: If we make this an method on `LocalAPIClient` which already has a `localapi.Handler`, we don't need to pass it in to `Call`.

oxtoacart: Also, at some point we'll want to send request bodies, so I suggest adding the parameter `body`.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
 
-func (w *LocalApiResponseWriter) WriteHeader(statusCode int) {
-	w.status = statusCode
+func (r *Response) WriteHeader(statusCode int) {
+	r.status = statusCode
 }
 
-func (w *LocalApiResponseWriter) Body() []byte {
-	return w.body.Bytes()
+func (r *Response) Body() io.ReadCloser {
```
oxtoacart, line 38: Note that this changes the return type from `[]byte` to `io.ReadCloser`. This is less convenient for the caller because it requires them to read from the reader, and it requires them to close the reader when finished, but it has the advantage of not having to read the whole response body into memory if we don't want to or have to.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
 
-var ErrBadHttpStatus = errors.New("bad http status for localapi response")
+// Call calls the given endpoint on the local API using the given HTTP method
+// optionally sending the given body. It returns a Response representing the
+// result of the call and an error if the call could not be completed or the
+// local API returned a status code in the 400 series or greater.
+// Note - Response includes a response body available from the Body method, it
+// is the caller's responsibility to close this.
+func (cl *LocalAPIClient) Call(ctx context.Context, method, endpoint string, body io.Reader) (*Response, error) {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
oxtoacart, line 61: I really like that you included a timeout. A more idiomatic and less verbose way to do this is with contexts.

Note also that passing a context to `Call` is idiomatic and useful if we have some larger event that should cancel processing. In practice, we're passing the appCtx, which means that when that context is cancelled, any in-flight requests will be cancelled too, which is convenient.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
-			log.Printf("%s bad http status: %v", endpoint, w.StatusCode())
-			responseError = ErrBadHttpStatus
-		}
-		done <- w
-	}()
+	req, err := http.NewRequestWithContext(ctx, method, "/localapi/v0/"+endpoint, body)
+	if err != nil {
+		return nil, fmt.Errorf("error creating new request for %s: %w", endpoint, err)
+	}
+	pipeReader, pipeWriter := io.Pipe()
```
oxtoacart, line 68: `io.Pipe` allows us to present the http API with an `io.Writer` to which to write the response body and present our client with an `io.Reader` from which to read the response body.

cmd/localapiclient/localapiclient.go
```diff
@@ -1,87 +1,82 @@
...
-	case <-time.After(2 * time.Second):
-		return nil, fmt.Errorf("request to %s timed out", endpoint)
+	resp := &Response{
+		headers:    http.Header{},
+		status:     http.StatusOK,
+		bodyReader: pipeReader,
+		bodyWriter: pipeWriter,
+	}
+	cl.h.ServeHTTP(resp, req)
+	if resp.StatusCode() >= 400 {
```
oxtoacart, line 78: I'm not sure whether 300 or 400 is better here(?) The 300 series is just redirects, whereas 400 and above are true errors. Does localapi even use redirects? If so, should we follow them here?

cmd/tailscale/main.go
```diff
@@ -58,6 +58,8 @@ type App struct {
 	store             *stateStore
 	logIDPublicAtomic atomic.Pointer[logid.PublicID]
 
+	localAPIClient *localapiclient.LocalAPIClient
```
oxtoacart, line 61: We can keep one instance of localAPIClient on the app for use in methods like getBugReportID().

cmd/localapiclient/localapiclient.go
oxtoacart: Note also that io.ReadCloser won't be able to be bound to Swift, so to maintain similar streaming semantics we'll probably need to get creative with our own interface using callbacks or something like that.

kari-ts: This is definitely an improvement, thanks!

cmd/localapiclient/localapiclient.go
kari-ts: Thanks for the explanation!

cmd/localapiclient/localapiclient.go
kari-ts: It looks like the timeout now only applies to making the request - is that right? It'd be nice to have timeouts for the rest of the work done by this function too.

cmd/localapiclient/localapiclient.go
kari-ts: localapi currently isn't using redirects. 400 SGTM, I am flagging this in #apple

cmd/localapiclient/localapiclient.go
oxtoacart: Great question! Because we use the function [http.NewRequestWithContext](https://pkg.go.dev/net/http#NewRequestWithContext), the context governs the entire lifetime of the request. As noted in the doc:

> For an outgoing client request, the context controls the entire lifetime of a request and its response: obtaining a connection, sending the request, and reading the response headers and body.

cmd/tailscale/main.go
```diff
@@ -563,13 +566,22 @@ func (a *App) runBackend() error {
 	}
 }
 
-func getBugReportID(h *localapi.Handler, bugReportChan chan<- string, fallbackLog string) {
-	w, err := localapiclient.CallLocalApi(h, "POST", "bugreport")
-	if w == nil || err != nil {
+func (a *App) getBugReportID(bugReportChan chan<- string, fallbackLog string) {
+	r, err := a.localAPIClient.Call(a.appCtx, "POST", "bugreport", nil)
```
kari-ts, line 570: a.appCtx is an Android Context, which is mostly for accessing application environment/system resources. I'm going to create a new golang context here instead, is that ok?

cmd/localapiclient/localapiclient.go
kari-ts: Since localapi has its own implementation of ServeHTTP, though, can we expect that this is still true? Iiuc, the handlers in localapi should use r.ctxt?

oxtoacart: Oh yikes, good point 🤦   So you will need something like your OG implementation of `Call()`, maybe just with the added feature of respecting the deadline on a passed-in `ctx` and returning a body reader that also respects the context?

cmd/tailscale/main.go
oxtoacart: Sorry, I misunderstood what that was. Is there some kind of global context that we cancel when the app shuts down? That's really what I was looking for here.

PR #162, opened 2024-02-09
bump version and oss, and update build.yml to set up go version

kari-ts:


catzkorn (approved):
bumping for yesterday unstable release + bumping again to get the correct build version.

PR #164, opened 2024-02-14
cmd/tailscale/main: use localapi for logging out

kari-ts:


oxtoacart:
@kari-ts Does this need to rebased to the latest localclient stuff?

cmd/tailscale/main.go
```diff
@@ -602,6 +598,20 @@ func (a *App) login(ctx context.Context) {
...
 
+func (a *App) logout(ctx context.Context) error {
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(5*time.Minute))
+	defer cancel()
+	r, err := a.localAPIClient.Call(ctx, "POST", "logout", nil)
+	defer r.Body().Close()
+
+	if err != nil {
+		log.Printf("logout: %s", err)
+		a.backend.Logout(ctx)
```
oxtoacart, line 609: If localAPIClient.Call() timed out, this will immediately fail with a timeout as well. Maybe create a new context with its own context?

cmd/tailscale/main.go
```diff
@@ -602,6 +598,20 @@ func (a *App) login(ctx context.Context) {
 	}
 }
 
+func (a *App) logout(ctx context.Context) error {
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(5*time.Minute))
```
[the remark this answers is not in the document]
oxtoacart: nit: context.WithTimeout() seems pithier here

```
context.WithTimeout(ctx, 5*time.Minute)
```

cmd/tailscale/main.go
kari-ts: Sounds good. Using the 5 minute timeout that was originally used for the logout timeout here, and using a shorter timeout for the localapi call.

oxtoacart (approved):

PR #166, opened 2024-02-15
cmd/tailscale/main: use custom server if exists

kari-ts:
If a custom login server has been saved, use that on login

Fixes tailscale/tailscale#17470


bradfitz:
> `Fixes tailscale/tailscale#17470`

That issue doesn't exist.

cmd/tailscale/main.go
```diff
@@ -1293,6 +1293,11 @@ func (a *App) processUIEvents(w *app.Window, events []UIEvent, act jni.Object, s
 		case ExitAllowLANEvent:
 			requestBackend(e)
 		case WebAuthEvent:
+			if srv, _ := a.store.ReadString(customLoginServerPrefKey, ""); srv != state.backend.Prefs.ControlURL {
+				requestBackend(SetLoginServerEvent{URL: srv})
+				// wait a moment for the backend to restart
+				<-time.After(200 * time.Millisecond)
```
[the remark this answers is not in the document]
kari-ts: thanks for the suggestion!

cmd/tailscale/main.go
```diff
@@ -197,12 +197,15 @@ type SetLoginServerEvent struct {
 	URL string
 }
 
+type WebAuthAndMaybeSetServerEvent struct {
```
bradfitz, line 200: This name invites enough questions to deserve a comment :)

bradfitz (commented):
Commit message + PR reference to the issue is still the wrong repo.

cmd/tailscale/main.go
```diff
@@ -197,12 +197,16 @@ type SetLoginServerEvent struct {
 	URL string
 }
 
+// Web login event. If there is a URL, set the login server first.
```
bradfitz, line 200: per https://go.dev/wiki/CodeReviewComments#comment-sentences this should be a complete sentence

And you never say what the URL is a URL of. If there's a URL of what, set the login server first?

```go
// WebAuthAndMaybeSetServerEvent is a web login event, after the
// user [does _________].
type WebAuthAndMaybeSetServerEvent struct {
    // URL optionally specifies a URL of _______.
    // If set, ___________.
    URL string
}
```

cmd/tailscale/main.go
```diff
@@ -1292,7 +1310,10 @@ func (a *App) processUIEvents(w *app.Window, events []UIEvent, act jni.Object, s
 			requestBackend(e)
 		case ExitAllowLANEvent:
 			requestBackend(e)
-		case WebAuthEvent:
+		case WebAuthAndMaybeSetServerEvent:
+			if srv, _ := a.store.ReadString(customLoginServerPrefKey, ""); srv != state.backend.Prefs.ControlURL {
+				e.URL = srv
```
bradfitz, line 1315: I see. It's a control plane URL but only set if it differs from what we had before.

cmd/tailscale/main.go
```diff
@@ -458,8 +462,22 @@ func (a *App) runBackend(ctx context.Context) error {
 			case ExitAllowLANEvent:
 				state.Prefs.ExitNodeAllowLANAccess = bool(e)
 				go b.backend.SetPrefs(state.Prefs)
-			case WebAuthEvent:
-				if !signingIn {
+			case WebAuthAndMaybeSetServerEvent:
+				if e.URL != "" && !signingIn {
+					state.Prefs.ControlURL = e.URL
+					b.backend.SetPrefs(state.Prefs)
```
bradfitz, line 468: This is disk I/O on the UI thread, no? Note line 460 does this in a goroutine.

Should this move to line 474?

But you'll want to make a copy of state.Prefs first to close over in that goroutine, otherwise it'd be a data race reading state.Prefs from another goroutine.  (on line 460 it's not a problem because the immediate "go F(arg)" copies arg to the arguments of F before the goroutine starts)

PR #168, opened 2024-02-16
cmd/tailscale, tailscale/ipn: fix alway-on VPN

kari-ts:
-If a ConnectEvent is received before the first notification, (as happens when a connection is attempted due to Always-on after device reboot) create state.Prefs.
-Create an intent to start the VPN worker in the case of an always-on intent received on device reboot
-Rename onConnect channel to onVPNRequested, since this isn't doing the actual connecting

Fixes https://github.com/tailscale/tailscale/issues/2481

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.java
+++ b/android/src/main/java/com/tailscale/ipn/App.java
@@ -135,13 +135,13 @@ public class App extends Application {
 
 	public void startVPN() {
 		Intent intent = new Intent(this, IPNService.class);
-		intent.setAction(IPNService.ACTION_CONNECT);
+		intent.setAction(IPNService.ACTION_REQUEST_VPN);
 		startService(intent);
 	}
 
 	public void stopVPN() {
 		Intent intent = new Intent(this, IPNService.class);
-		intent.setAction(IPNService.ACTION_DISCONNECT);
+		intent.setAction(IPNService.ACTION_STOP_VPN);
 		startService(intent);
 	}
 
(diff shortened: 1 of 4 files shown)


aalokkamble:
It is not working despite keeping following enabled in my phone. 

'Always-on VPN
Stay connected to VPN at all times' 


Also it should remember setting such as 'exit node' .

mrrfv:
Works on my Google Pixel 7a running the latest, stock Android 14 and F-Droid build. A manually disconnected VPN (as expected) doesn't get immediately auto-reconnected by the system, but an automatic connection does happen a few seconds after a reboot _if Tailscale was connected prior to it_. Thank you.

Bug reports, sent in the hope that they might speed things up, even if everything seems to be working as intended:

Tailscale disabled, then phone rebooted - doesn't automatically reconnect:
`BUG-fafba6038bcf26cc0f678b7b0d7366d3b5a180c7d04335acac40f3335bb2960b-20240304153841Z-ac0122f283ce7182`
Tailscale enabled, then phone rebooted - automatically reconnects:
`BUG-fafba6038bcf26cc0f678b7b0d7366d3b5a180c7d04335acac40f3335bb2960b-20240304154106Z-5f6cedd2a4ac306a`

barnstar (approved):

mattalxndr:
Tailscale is not starting automatically for me, even though I have exactly the same env as @mrrfv.


Google Pixel 7a 
Android 14 (build AP1A.240305.019.A1)
Tailscale 1.60.1-t2caffeeb4-g6f7ae79a31c (FDroid)

mattalxndr:
Clicking the quick tile does nothing. I have to long press it and to open tailscale, then press the on switch manually

bradfitz:
@mattalxndr, the rewrite of the Tailscale Android app has landed between this bug being closed and your comment. We won't be debugging the old version of the app anymore. See https://www.reddit.com/r/Tailscale/comments/1c2jdhp/android_users_can_you_keep_a_secret/

PR #171, opened 2024-02-26
cmd/tailscale: pass nil TailFSForLocal to netstack.Create

oxtoacart:
In the latest tailscale.com, netstack.Create has a new parameter that must be supplied.

Updates #cleanup


kari-ts:
Thank you!

PR #172, opened 2024-02-27
cmd/tailscale/main: use localapi for edit prefs

kari-ts:
-Use PrefsView instead of Prefs
-Replace SetPrefs calls with localapi using the prefs endpoint 
-Follow-up will be to hide option to use an exit node if the node is being run as an exit node and to hide the option to run as an exit node if an exit node is being used 

Updates tailscale/tailscale#10992

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -5,11 +5,13 @@
 package main
 
 import (
+	"bytes"
 	"cmp"
 	"context"
 	"crypto/rand"
 	"crypto/sha1"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -66,15 +68,17 @@ type App struct {
 	// netStates receives the most recent network state.
 	netStates chan BackendState
 	// prefs receives new preferences from the backend.
-	prefs chan *ipn.Prefs
+	prefs chan ipn.PrefsView
 	// browseURLs receives URLs when the backend wants to browse.
 	browseURLs chan string
 	// targetsLoaded receives lists of file targets.
 	targetsLoaded chan FileTargets
 	// invalidates receives whenever the window should be refreshed.
 	invalidates chan struct{}
-	// bugReport receives the bug report from the backend's localapi call
+	// bugReport receives the bug report from the backend's localapi call.
 	bugReport chan string
+	// editFns receives instructions for which preferences to edit and enforces the ordering of these edits.
+	editFns chan func(currPrefs ipn.PrefsView, updatedPrefs *ipn.MaskedPrefs)
 }
 
 var (
@@ -153,7 +157,7 @@ type Peer struct {
 }
 
 type BackendState struct {
-	Prefs        *ipn.Prefs
+	Prefs        ipn.PrefsView
 	State        ipn.State
 	NetworkMap   *netmap.NetworkMap
 	LostInternet bool
@@ -227,10 +231,11 @@ func main() {
 		appCtx:        jni.Object(app.AppContext()),
 		netStates:     make(chan BackendState, 1),
 		browseURLs:    make(chan string, 1),
-		prefs:         make(chan *ipn.Prefs, 1),
+		prefs:         make(chan ipn.PrefsView, 1),
 		targetsLoaded: make(chan FileTargets, 1),
 		invalidates:   make(chan struct{}, 1),
 		bugReport:     make(chan string, 1),
+		editFns:       make(chan func(currPrefs ipn.PrefsView, updatedPrefs *ipn.MaskedPrefs), 1),
 	}
 	err := jni.Do(a.jvm, func(env *jni.Env) error {
 		loader := jni.ClassLoaderFor(env, a.appCtx)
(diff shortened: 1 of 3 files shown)


cmd/tailscale/main.go
```diff
@@ -57,6 +60,8 @@ type App struct {
 	// appCtx is a global reference to the com.tailscale.ipn.App instance.
 	appCtx jni.Object
 
+	mu sync.Mutex
```
bradfitz, line 63: whenever you add a mutex, you need to document what it guards.

cmd/tailscale/main.go
```diff
@@ -153,7 +158,7 @@ type Peer struct {
 }
 
 type BackendState struct {
-	Prefs        *ipn.Prefs
+	Prefs        *ipn.PrefsView
```
bradfitz, line 161: not a pointer. a View is just a pointer-sized struct wrapper around a pointer so it's used as a value type.

cmd/tailscale/main.go
```diff
@@ -978,8 +1103,8 @@ func (a *App) notify(state BackendState) {
 	}
 }
 
-func (a *App) setPrefs(prefs *ipn.Prefs) {
-	wantRunning := jni.Bool(prefs.WantRunning)
+func (a *App) setPrefs(prefs *ipn.PrefsView) {
```
bradfitz, line 1106: no pointer

cmd/tailscale/main.go
```diff
@@ -227,7 +232,7 @@ func main() {
 		appCtx:        jni.Object(app.AppContext()),
 		netStates:     make(chan BackendState, 1),
 		browseURLs:    make(chan string, 1),
-		prefs:         make(chan *ipn.Prefs, 1),
+		prefs:         make(chan *ipn.PrefsView, 1),
```
bradfitz, line 235: here too.

cmd/tailscale/main.go
```diff
@@ -451,26 +473,71 @@ func (a *App) runBackend(ctx context.Context) error {
...
+						updatedPrefs.AdvertiseRoutes = routes.AsSlice()
+						updatedPrefs.SetAdvertiseExitNode(bool(e))
+						updatedPrefs.AdvertiseRoutesSet = true
+					})
+					prefsView := prefs.View()
+					state.Prefs = &prefsView
+					a.mu.Unlock()
+					if prefs.AdvertisesExitNode() != bool(e) {
+						log.Printf("localapi failed to edit prefs for BeExitNodeEvent")
+					}
```
bradfitz, line 493: if a function returns (T, error), conventionally (unless documented to the contrary) T is not valid for inspection when err is non-nil. So this could crash.

cmd/tailscale/main.go
```diff
@@ -370,16 +375,19 @@ func (a *App) runBackend(ctx context.Context) error {
 		case n := <-notifications:
 			exitWasOnline := state.ExitStatus == ExitOnline
 			if p := n.Prefs; p != nil && n.Prefs.Valid() {
-				state.Prefs = p.AsStruct()
+				// make a copy of prefs and use the readonly view
+				state.Prefs = p.AsStruct().View()
```
oxtoacart, line 379: Why do we need to copy the PrefsView if we're just getting another read-only view to them anyway?

cmd/tailscale/main.go
```diff
@@ -370,16 +375,19 @@ func (a *App) runBackend(ctx context.Context) error {
...
 				state.updateExitNodes()
-				a.setPrefs(state.Prefs)
+				a.prefs <- state.Prefs
 			}
-			first := state.Prefs == nil
+			first := !state.Prefs.Valid()
```
oxtoacart, line 383: Can the first notification ever contain a valid prefs?

cmd/tailscale/main.go
```diff
@@ -370,16 +375,19 @@ func (a *App) runBackend(ctx context.Context) error {
...
-				state.Prefs = ipn.NewPrefs()
-				state.Prefs.Hostname = a.hostname()
-				go b.backend.SetPrefs(state.Prefs)
-				a.setPrefs(state.Prefs)
+				newPrefs := ipn.NewPrefs()
+				newPrefs.Hostname = a.hostname()
+				go b.backend.SetPrefs(newPrefs)
+				state.Prefs = newPrefs.View()
+
+				go a.editPrefs(ctx, state.Prefs)
```
oxtoacart, line 390: Instead of starting this here lazily, it might be safer to just start it before we enter the for loop on line 359. Otherwise, if we get more than two notifications with a valid PrefsView, we'll end up deadlocking because we push them to a.prefs but it can only buffer one.

If you do that, then I think(?) you should be able to make a.prefs an unbuffered channel.

cmd/tailscale/main.go
```diff
@@ -570,6 +585,46 @@ func (a *App) runBackend(ctx context.Context) error {
...
+			if err != nil {
+				log.Printf("error reading localapi response %v", err)
+			}
+			var newPrefs ipn.PrefsView
+			err = json.Unmarshal(data, &newPrefs)
+			if err != nil {
+				log.Printf("error unmarshaling localapi response %v", err)
+			}
+			prefs = newPrefs
+		default:
```
oxtoacart, line 623: Having this `default` here is going to cause your for loop to spin in a tight loop.

I think what you want is just `for fn := range a.editFns {}`.

See https://go.dev/play/p/z5HWzCU93_e

cmd/tailscale/main.go
```diff
@@ -704,8 +759,8 @@ func (a *App) isChromeOS() bool {
 func (s *BackendState) updateExitNodes() {
 	s.ExitStatus = ExitNone
 	var exitID tailcfg.StableNodeID
```
oxtoacart, line 761: This looks like it's read/written from different goroutines, so probably better to make it an `atomic.Pointer` or guard it with a mutex.

PR #173, opened 2024-02-27
android: add kotlin dependencies and localAPIClient hooks

barnstar:
updates tailscale/tailscale#10992

Adds all of the kotlin build dependencies and a rudimentary implementation of a localAPIClient in the front end, wired up via JNI.

Partial implementation of the required localAPI data classes and the required dependencies for JSON serialization.


android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIRequest.kt
```diff
@@ -0,0 +1,34 @@
...
+
+class LocalAPIRequest(
+    val path: String,
+    val method: String,
+    val handler: RawAPIResponse,
+) {
+    companion object {
+        var cookieCounter: Int = 0
+
+        fun getCookie(): String {
```
kari-ts, line 21: Is there potential for data racing here if we have multiple threads creating LocalAPIRequest concurrently?

android/src/main/java/com/tailscale/ipn/ui/model/Ipn.kt
```diff
@@ -0,0 +1,106 @@
...
+// license that can be found in the LICENSE file.
+
+package com.tailscale.ipn.ui.model
+
+import kotlinx.serialization.*
+
+class ipn {
+
+    // Represents the overall state of the Tailscale engine.
+    enum class State(val value: Int) {
```
kari-ts, line 12: I know on Windows there's this extra state on top of this that expresses "UI state," and includes a "needsAdminSigning" state for tailnet lock. I'm curious how this is done on other platforms

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIRequest.kt
barnstar: There is!  Fixed.

android/src/main/java/com/tailscale/ipn/ui/model/Ipn.kt
barnstar: That's handled differently on iThings.  There's a separate set of localAPI calls for it.   

I'm going to add one API at a time along with the models it needs.  I suspect we'll have to revisit those as we start to actually use them to drive UI.  I'm going to use the iOS version as the basis here, since it's going to functionally be the closest in terms of language, idioms, and general functionality. I want to keep the commits a reasonable size though.

kari-ts (approved):

PR #176, opened 2024-02-28
cmd/tailscale/ui: only show exit node options if relevant

kari-ts:
-If exit node is being run, don't show an option to use an exit node
-If exit node is being used, don't show an option to run/stop running as an exit node

### DIFF
--- a/cmd/tailscale/main.go
+++ b/cmd/tailscale/main.go
@@ -5,11 +5,13 @@
 package main
 
 import (
+	"bytes"
 	"cmp"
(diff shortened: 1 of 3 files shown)


oxtoacart:
Sorry for the glacially slow response on reviewing.

In testing, I noticed that the first time I choose an exit node, nothing happens. I have to choose it a 2nd time, then Tailscale starts using it.

I don't know if that's related to these changes or not.

Other than the above, this change appears to be functioning as intended.

PR #179, opened 2024-03-04
Jonathan/notifier

barnstar:
android: add notifier support a data model and compose dependencies and some basic UI

fixes ENG-2084
fixes ENG-2086

Adds support for the ipnBusWatcher directly via a JNI API rather than HTTP via LocalAPIClient

Adds a rudimentary controller class and a model from which we can construct ViewModels

Added some basic Compose UI for the main screen and some placeholders.

Cleans up some of the JNI bindings

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,6 +1,7 @@
 
 buildscript {
 	ext.kotlin_version = "1.9.22"
+	ext.kotlin_compose_version = "1.5.10"
 
 	repositories {
 		google()
@@ -38,6 +39,12 @@ android {
 		sourceCompatibility JavaVersion.VERSION_17
 		targetCompatibility JavaVersion.VERSION_17
 	}
+	buildFeatures {
+		compose true
+	}
+	composeOptions {
+        kotlinCompilerExtensionVersion = "$kotlin_compose_version"
+    }
 	flavorDimensions "version"
 	productFlavors {
 		fdroid {
@@ -52,15 +59,33 @@ android {
 }
 
 dependencies {
+	// Android dependencies.
 	implementation "androidx.core:core:1.9.0"
     implementation 'androidx.core:core-ktx:1.9.0'
 	implementation "androidx.browser:browser:1.5.0"
 	implementation "androidx.security:security-crypto:1.1.0-alpha06"
 	implementation "androidx.work:work-runtime:2.8.1"
+
+	// Kotlin dependencies.
 	implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3"
 	implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.0"
+	implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.8.0"
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version" 
+
+	// Compose dependencies.
+	def composeBom = platform('androidx.compose:compose-bom:2023.06.01')
+    implementation composeBom
+    androidTestImplementation composeBom
+    implementation 'androidx.compose.material3:material3:1.0.0'
+	implementation "androidx.compose.ui:ui:1.4.3"
+	implementation "androidx.compose.ui:ui-tooling:1.4.3"
+	implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.3.1'
+	implementation 'androidx.activity:activity-compose:1.7.2'
+
+	// Tailscale dependencies.
 	implementation ':ipn@aar'
+
+	// Tests
 	testImplementation "junit:junit:4.12"
 
 	// Non-free dependencies.
--- a/android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
@@ -6,18 +6,33 @@ package com.tailscale.ipn.ui.localapi
 
 import android.util.Log
 import com.tailscale.ipn.ui.model.*
-import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 
-// A response from the echo endpoint.
 typealias StatusResponseHandler = (Result<IpnState.Status>) -> Unit
-
 typealias BugReportIdHandler = (Result<BugReportID>) -> Unit
-
 typealias PrefsHandler = (Result<Ipn.Prefs>) -> Unit
 
 class LocalApiClient {
     constructor() {
         Log.d("LocalApiClient", "LocalApiClient created")
+    }
+
+    companion object {
+        private val _isReady = MutableStateFlow(false)
+        val isReady: StateFlow<Boolean> = _isReady
+
+        // Called by the backend when the localAPI is ready to accept requests.
+        @JvmStatic
+        fun onReady() {
+            _isReady.value = true
+            Log.d("LocalApiClient", "LocalApiClient is ready")
+        }
     }
 
     // Perform a request to the local API in the go backend.  This is
@@ -33,26 +48,31 @@ class LocalApiClient {
     //         the corresponding request.  Cookies must be unique for each request.
     external fun doRequest(request: String, method: String, body: String, cookie: String)
 
+    protected val scope = CoroutineScope(Dispatchers.IO + Job())
+
     fun <T> executeRequest(request: LocalAPIRequest<T>) {
-        Log.d("LocalApiClient", "Executing request:${request.method}:${request.path}")
-        addRequest(request)
-        // The jni handler will treat the empty string in the body as null.
-        val body = request.body ?: ""
-        doRequest(request.path, request.method, body, request.cookie)
+        scope.launch {
+            isReady.first { it == true }
+            Log.d("LocalApiClient", "Executing request:${request.method}:${request.path}")
+            addRequest(request)
+            // The jni handler will treat the empty string in the body as null.
+            val body = request.body ?: ""
+            doRequest(request.path, request.method, body, request.cookie)
+        }
     }
 
     // This is called from the JNI layer to publish localAPIResponses.  This should execute on the
     // same thread that called doRequest.
     fun onResponse(response: String, cookie: String) {
         val request = requests[cookie]
-        if (request != null) {
+
+        request?.let {
             Log.d("LocalApiClient", "Reponse for request:${request.path} cookie:${request.cookie}")
-            // The response handler will invoked internally by the request parser 
+            // The response handler will invoked internally by the request parser
             request.parser(response)
             removeRequest(cookie)
-        } else {
-            Log.e("LocalApiClient", "Received response for unknown request: ${cookie}")
         }
+                ?: { Log.e("LocalApiClient", "Received response for unknown request: ${cookie}") }
     }
 
     // Tracks in-flight requests and their callback handlers by cookie. This should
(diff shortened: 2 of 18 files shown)


android/build.gradle
```diff
@@ -52,15 +59,33 @@ android {
...
+
+	// Kotlin dependencies.
 	implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3"
 	implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.0"
+	implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.8.0"
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version" 
+
+	// Compose dependencies.
+	def composeBom = platform('androidx.compose:compose-bom:2023.06.01')
+    implementation composeBom
```
dblohm7, line 77: indentation on this line and the next two below it

android/build.gradle
```diff
@@ -38,6 +39,12 @@ android {
 		sourceCompatibility JavaVersion.VERSION_17
 		targetCompatibility JavaVersion.VERSION_17
 	}
+	buildFeatures {
+		compose true
+	}
+	composeOptions {
+        kotlinCompilerExtensionVersion = "$kotlin_compose_version"
```
dblohm7, line 46: Indentation on this line and the one below it. Tabs vs spaces or something?

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -6,20 +6,35 @@ package com.tailscale.ipn.ui.localapi
...
 typealias PrefsHandler = (Result<Ipn.Prefs>) -> Unit
 
 class LocalApiClient {
     constructor() {
         Log.d("LocalApiClient", "LocalApiClient created")
     }
 
+    companion object {
+        private val _isReady = MutableStateFlow(false)
+        val isReady: StateFlow<Boolean> = _isReady
```
dblohm7, line 28: I think you can make this a bit more concise if you write it as:
`val isReady: _isReady.asStateFlow()`

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -33,26 +48,31 @@ class LocalApiClient {
     //         the corresponding request.  Cookies must be unique for each request.
     external fun doRequest(request: String, method: String, body: String, cookie: String)
 
+    protected val scope = CoroutineScope(Dispatchers.IO + Job())
```
dblohm7, line 51: Recognizing that there has been, is, and will be a lot of flux over the next little while, is there a way to plumb [lifecycle-aware coroutine scope](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope) down into this client?

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -33,26 +48,31 @@ class LocalApiClient {
...
+        request?.let {
             Log.d("LocalApiClient", "Reponse for request:${request.path} cookie:${request.cookie}")
-            // The response handler will invoked internally by the request parser 
+            // The response handler will invoked internally by the request parser
             request.parser(response)
             removeRequest(cookie)
-        } else {
-            Log.e("LocalApiClient", "Received response for unknown request: ${cookie}")
         }
+                ?: { Log.e("LocalApiClient", "Received response for unknown request: ${cookie}") }
```
dblohm7, line 75: indentation?

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -0,0 +1,154 @@
...
+//
+// The primary entry point here is watchIPNBus which will start a watcher on the IPN bus
+// and return you the session Id.  When you are done with your watcher, you must call
+// unwatchIPNBus with the sessionId.
+class Notifier {
+    constructor() {
+        Log.d("Notifier", "Notifier created")
+    }
+
+    protected val scope = CoroutineScope(Dispatchers.IO + Job())
```
dblohm7, line 41: Same re: scope

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -0,0 +1,154 @@
...
+        private fun generateSessionId(): String {
+            synchronized(sessionIdLock) {
+                sessionId += 1
+                return sessionId.toString()
+            }
+        }
+    }
+
+    // Starts an IPN Bus watcher.  **This is blocking** and will not return until
+    // the watcher is stopped and must be excuted in a suitable coroutine scope such
```
dblohm7, line 77: s/excuted/executed

android/src/main/java/com/tailscale/ipn/ui/service/IpnModel.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.first
+
+class IpnModel {
+    protected val scope = CoroutineScope(Dispatchers.Default + Job())
```
dblohm7, line 20: Scope

dblohm7 (changes requested):
Mostly some indentation and commenting nits which I don't need to see again.

I'm more interested in your thoughts on coroutine scope; please see relevant comment below.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -6,20 +6,35 @@ package com.tailscale.ipn.ui.localapi
...
 typealias BugReportIdHandler = (Result<BugReportID>) -> Unit
-
 typealias PrefsHandler = (Result<Ipn.Prefs>) -> Unit
 
 class LocalApiClient {
     constructor() {
         Log.d("LocalApiClient", "LocalApiClient created")
     }
 
+    companion object {
```
kari-ts, line 26: Why make this a companion vs making LocalAPIClient a singleton?

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -6,20 +6,35 @@ package com.tailscale.ipn.ui.localapi
...
+        private val _isReady = MutableStateFlow(false)
```
kari-ts, line 27: This is only going to get set once right? 
Maybe use something like a CompletableDeferred instead?

android/src/main/java/com/tailscale/ipn/ui/model/Ipn.kt
```diff
@@ -16,20 +16,14 @@ class Ipn {
...
-        enum class NotifyWatchOpt(val value: Int) {
-                engineUpdates(0),
-                initialState(1 shl 1),
-                prefs(1 shl 2),
-                netmap(1 shl 3),
-                noPrivateKeys(1 shl 4),
-                initialTailFSShares(1 shl 5)
+                companion object {
+                        fun fromInt(value: Int): State? {
+                                return State.values().first { s -> s.value == value }
```
kari-ts, line 23: State.values().firstOrNull { it.value == value } instead? otherwise it'll throw an exception if no match is found

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -0,0 +1,154 @@
...
+
+
+class Watcher(
+    val sessionId: String,
+    val mask: Int,
+    val callback: NotifierCallback
+)
+
+// Notifier is a wrapper around the IPN Bus notifier.  It provides a way to watch the
+// for changes in various parts of the Tailscale engine.  You will typically only use
```
kari-ts, line 29: is there ever a case where you might have more than one Notifier?

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -0,0 +1,154 @@
...
+// a single Notifier per instance of your application which lasts for the lifetime of
+// the process.
+//
+// The primary entry point here is watchIPNBus which will start a watcher on the IPN bus
+// and return you the session Id.  When you are done with your watcher, you must call
+// unwatchIPNBus with the sessionId.
+class Notifier {
```
kari-ts, line 36: make this a singleton?

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
kari-ts: +1 to lifecycle aware scopes!

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
barnstar: Singletons make testing (when we get there) hard.   They way I'm going to bolt this together is to have a manager object that owns a single instance of both the Notifier and the LocalAPIClient and the DataModel.   That manager will be a singleton.  This allows you to inject concrete instances of the various things later.

I may even opt to do away with that, and have each viewModel just have it's own notifier and LocalAPIClient.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
barnstar: I've added a comment to remind me to do this.  A little experimentation with the activity scope yielded nothing but problems but I'm probably misunderstanding how coroutine scopes and contexts work.  I have to RTFM.  Technically, we're a single activity app, so this isn't critical - at least until we're aren't a single activity app...

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
barnstar: Yup.  Different notifiers can give you different streams...  It's probably not the case that we will have more than one, but I wanted to write it so we could.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
barnstar: It is a defacto singleton because it's a property on a singleton... But for testing, I want to be able to inject these.

android/build.gradle
barnstar: Android studio to the rescue. "Reformat Code"... VSCode FTL.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
barnstar: FYI - I've removed all of the singletons in a future patch.  Actions are injected upstream.  Much cleaner.  There's still a single IPNManager owned by the MainActivity which owns a single Notifier and LocalAPIClient.  None of those references are passed around though.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -0,0 +1,155 @@
...
+// and return you the session Id.  When you are done with your watcher, you must call
+// unwatchIPNBus with the sessionId.
+class Notifier() {
+
+    private val scope = CoroutineScope(Dispatchers.IO + Job())
+
+    // NotifyWatchOpt is a bitmask of options supplied to the notifier to specify which
+    // what we want to see on the Noitfy bus
+    enum class NotifyWatchOpt(val value: Int) {
+        engineUpdates(1 shl 0),
```
barnstar, line 43: These are not correct, but I've fixed them upstream.

oxtoacart (approved):

oxtoacart:
@dblohm7 

> Mostly some indentation and commenting nits which I don't need to see again.
> 
> I'm more interested in your thoughts on coroutine scope; please see relevant comment below.

If it's okay with you, I'd like to address this in a subsequent PR.

dblohm7 (approved):
OK'd by @oxtoacart, so wfm

PR #180, opened 2024-03-05
android: restructure app

kari-ts:
Move existing Java code out of main/ into a new directory app/ and create newapp/ for new Android code. This will allow us to work on the new app without polluting the existing app, and more easily remove old code in the future


barnstar (changes requested):
This needs a rebase, it looks like it pulls in all of the changes from my initial branch which is now merged to main.  Let me know if you need any help with the Git stuff.

barnstar (approved):

barnstar (commented):
I suggest we don't mix and match here.   Let's cut and run.

Instead of a newapp folder, let's duplicate the entire android folder structure.  There's almost nothing in there we need to keep (or at least not in java).   So instead of /android/src/newapp,  let's do /androidv2/src/.  That way we don't have to worry about mixing with the old stuff and breaking it or gradle changes that might prevent us from shipping updates to v1.   The go stuff can remain common.  Keep all of the existing Java stuff for now - but because we have a copy of it, we can mangle it and replace it with Kotlin at will.   It is useful for now - because it allows the app to run with a simple replacement of the IPNActivity.

This will also make developing in Android Studio that much cleaner.  We will need to use the Android Studio environment to do any serious Kotlin work.

PR #181, opened 2024-03-06
Makefile, build.gradle: build newapp

kari-ts:
-Add tasks to build and compile native library for relevant ABIs for newapp
-Add newapp task to Makefile


Makefile
```diff
@@ -154,6 +154,10 @@ tailscale-fdroid.apk: $(AAR)
 	(cd android && ./gradlew test assembleFdroidDebug)
 	mv android/build/outputs/apk/fdroid/debug/android-fdroid-debug.apk $@
 
+tailscale-new.apk: 
+	(cd android && ./gradlew assembleNewAppDebug)
+	mv ./android/build/outputs/apk/newapp/debug/android-newapp-debug.apk $@
```
raggi, line 159: recommend using `install -C` here (and above) rather than `mv`, as that way you'll get near 0 cost incremental rebuilds if nothing actually changed

android/build.gradle
```diff
@@ -67,4 +104,37 @@ dependencies {
 	playImplementation 'com.google.android.gms:play-services-auth:20.7.0'
 }
 
+def ndkPath = project.hasProperty('ndkPath') ? project.property('ndkPath') : System.getenv('ANDROID_SDK_ROOT')
```
raggi, line 107: I put up a more portable variant of this in https://github.com/tailscale/tailscale-android/tree/raggi/buildport

raggi, line 107: the `:ipn@aar` dependency above needs to be removed, it's being included in all variants

PR #185, opened 2024-03-11
android/ui: add AboutView

agottardo:
Adds an AboutView with app icon and same content as the iOS version.


android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
```diff
@@ -0,0 +1,115 @@
...
+                    color = MaterialTheme.colorScheme.secondary
+                )
+            }
+            Column(
+                verticalArrangement = Arrangement.spacedBy(
+                    space = 4.dp, alignment = Alignment.CenterVertically
+                ), horizontalAlignment = Alignment.CenterHorizontally
+            ) {
+                OpenURLButton(
+                    stringResource(R.string.acknowledgements),"https://tailscale.com/licenses/apple"
```
raggi, line 80: /android

agottardo: LOL

android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
```diff
@@ -0,0 +1,115 @@
...
+                )
+                OpenURLButton(
+                    stringResource(R.string.privacy_policy), "https://tailscale.com/privacy-policy/"
```
raggi, line 83: Sometimes I put these kinds of urls into resources too, so if we later internationalize they can point at language specific versions too (once available), not terribly important though

raggi (approved):

PR #186, opened 2024-03-12
android: more code review suggestions

oxtoacart:
Suggestions for PR #179


Makefile
```diff
@@ -129,14 +130,18 @@ androidpath:
...
+	@mkdir -p android_legacy/libs && \
 	go run gioui.org/cmd/gogio \
 		-ldflags "-X tailscale.com/version.longStamp=$(VERSIONNAME) -X tailscale.com/version.shortStamp=$(VERSIONNAME_SHORT) -X tailscale.com/version.gitCommitStamp=$(TAILSCALE_COMMIT) -X tailscale.com/version.extraGitCommitStamp=$(OUR_VERSION)" \
 		-buildmode archive -target android -appid $(APPID) -tags novulkan,tailscale_go -o $@ github.com/tailscale/tailscale-android/cmd/tailscale
 
+$(AAR_NEXTGEN): $(AAR)
+	@mkdir -p android/libs && \
+	cp $(AAR) $(AAR_NEXTGEN)
+
+lib: $(AAR_NEXTGEN)
```
oxtoacart, line 143: This phony target makes it easy to build the aar for the new app.

android/src/main/java/com/tailscale/ipn/ui/service/IpnManager.kt
```diff
@@ -4,27 +4,26 @@
...
 class IpnManager {
-    var notifier = Notifier()
-    var apiClient = LocalApiClient()
-    val model: IpnModel
-    
-    constructor() {
-        model = IpnModel(notifier, apiClient)   
-    }
+    private var notifier = Notifier()
+    private var scope = CoroutineScope(SupervisorJob() + Dispatchers.Main)
```
oxtoacart, line 16: This doesn't quite get us to LifecycleScope, but I think it's an improvement to just have a single scope that we pass in to the other types that need it.

I'm not actually convinced about the current structure of using a singleton IpnManager and IpnModel, but I'm not ready to make any concrete suggestions on that.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -46,73 +46,57 @@ class LocalApiClient {
...
-    private var requestLock = Any()
-
-    fun addRequest(request: LocalAPIRequest<*>) {
-        synchronized(requestLock) { requests[request.cookie] = request }
-    }
-
-    fun removeRequest(cookie: String) {
-        synchronized(requestLock) { requests.remove(cookie) }
-    }
+    private var requests = ConcurrentHashMap<String, LocalAPIRequest<*>>()
```
oxtoacart, line 73: Using a ConcurrentHashMap eliminates the need to lock.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
```diff
@@ -46,73 +46,57 @@ class LocalApiClient {
     // body: The body of the request.
     // cookie: A unique identifier for this request.  This is used map responses to
     //         the corresponding request.  Cookies must be unique for each request.
-    external fun doRequest(request: String, method: String, body: String, cookie: String)
+    private external fun doRequest(request: String, method: String, body: ByteArray?, cookie: String)
```
oxtoacart, line 49: Using a ByteArray for request and response bodies should be more efficient than using strings. Eventually, I might even want to change the API to use streams, if I can figure out how to do that across the JNI boundary, but not worth chewing off right now.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIRequest.kt
```diff
@@ -41,59 +47,41 @@ private object Endpoint {
...
-        val method: String,
-        val body: String? = null,
-        val parser: (String) -> Unit,
+    path: String,
+    val method: String,
+    val body: ByteArray? = null,
+    val parser: (ByteArray) -> Unit,
 ) {
     val path = "/localapi/v0/$path"
+    val cookie = UUID.randomUUID().toString()
```
oxtoacart, line 65: Using a UUID eliminates the need for locking. In the past, I've run into weird issues combining Kotlin coroutines and Java locks, so whenever we can avoid the `synchronized` keyword, it's probably better to do so.

android/src/main/java/com/tailscale/ipn/ui/localapi/LocalAPIClient.kt
oxtoacart: Also, some things are not well suited to strings, for example Taildrop file PUTs.

PR #189, opened 2024-03-12
ui: add view to debug MDM settings

agottardo:
Adds a view to see the currently set MDM settings, we're going to need this to debug actual MDM integrations more effectively.


barnstar (approved):

agottardo:
Closing because Jonathan already pulled the commit into his branch

PR #193, opened 2024-03-12
.github/workflows,Makefile: add a build check CI for the new app

raggi:
Updates tailscale/tailscale#10992


Makefile
```diff
@@ -163,10 +163,13 @@ tailscale-new-fdroid.apk: $(AAR)
 	(cd android && ./gradlew test assembleFdroidDebug)
```
kari-ts, line 163: looks like the line above should be AAR_NEXTGEN too

PR #196, opened 2024-03-13
android: add settings screen and login flow

barnstar:
updates tailscale/corp#18202
updates ENG-2854

Adds a basic settings screen.  This isn't correctly localized, but that's on the way.

Adds the required hooks to edit prefs via localAPI.

Adds basic but incomplete login/logout flow.

Fixes the sorting of nodes on the main screen and fixes the proper display of your current node details.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/util/TimeUtil.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/util/TimeUtil.kt
@@ -3,9 +3,43 @@
 
 package com.tailscale.ipn.ui.util
 
+import java.time.Instant
+import java.time.format.DateTimeFormatter
+import java.util.Date
+
+
 class TimeUtil {
     fun keyExpiryFromGoTime(goTime: String?): String {
-        // (jonathan) TODO: Turn these time strings into 'in 4 months', 'in 2 days', 'in 1 year', etc
-        return goTime ?: "Never"
+
+        val time = goTime ?: return ""
+        val expTime = epochMillisFromGoTime(time)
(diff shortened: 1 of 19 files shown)


android/src/main/java/com/tailscale/ipn/ui/util/TimeUtil.kt
```diff
@@ -3,9 +3,43 @@
...
+        val now = Instant.now().toEpochMilli()
+
+        val diff = (expTime - now) / 1000
+
+        if(diff < 0){
+            return "expired"
+        }
+
+        return when (diff) {
+            in 0..60 -> "under a minute"
```
agottardo, line 25: use string resources?

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -173,7 +173,7 @@ fun ConnectView(user: IpnLocal.LoginProfile?, connectAction: () -> Unit, loginAc
             horizontalAlignment = Alignment.CenterHorizontally
     ) {
         Text(text = "Not Connected", style = MaterialTheme.typography.titleMedium)
```
agottardo, line 175: string resource

agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/util/TimeUtil.kt
barnstar: Yup... I have an open localization task...

PR #201, opened 2024-03-13
ui: add ManagedByView, hide MDMSettingsView on non-debug builds

agottardo:
Updates tailscale/corp#18202

- Adds the "Managed by OrganizationName" view we currently offer on iOS.
- Hides the MDM settings debug pane on non-debug builds.
- Refactored SettingsViewModel to take an `IpnManager` instead of an `IpnModel` (@barnstar, let me know whether this makes sense given your future plans)

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/ManagedByView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/ManagedByView.kt
@@ -1,0 +1,49 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package com.tailscale.ipn.ui.view
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.safeContentPadding
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/ManagedByView.kt
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2024 Tailscale Inc & AUTHORS All rights reserved.
```
barnstar, line 1: Wrong header... James update it.  You can set a default in Android Studio for new files.

android/src/main/java/com/tailscale/ipn/ui/viewModel/SettingsViewModel.kt
```diff
@@ -61,21 +98,52 @@ class SettingsViewModel(val model: IpnModel, val ipnActions: IpnActions, val nav
...
+            ),
+            mdmSettings.get(StringSetting.ManagedByOrganizationName)?.let {
+                Setting(
+                        ComposableStringFormatter(R.string.managed_by_orgName, it),
+                        SettingType.NAV,
+                        onClick = { navigation.onNavigateToManagedBy() },
+                        enabled = MutableStateFlow(true)
+                )
+            },
+            if (BuildConfig.DEBUG) {
```
raggi, line 128: Makes me think we should add a release builder to CI ASAP too

raggi (approved):

PR #203, opened 2024-03-14
ui: assorted UI tweaks, improve disconnected view

agottardo:
Updates tailscale/corp#18202

- Begin using the brand theme colors for both light and dark mode
- Adds a symbol to match the disconnected iOS view
- Adds a PrimaryActionButton for blue colored buttons
- Adds a static TailscaleLogoView, progress indicator mode to follow


android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -167,26 +174,79 @@ fun StartingView() {
...
+            Spacer(modifier = Modifier.size(1.dp))
+            PrimaryActionButton(onClick = loginAction) {
+                Text(text = stringResource(id = R.string.log_in))
+            }
         }
     }
 }
 
+@Composable
+fun PrimaryActionButton(
```
agottardo, line 234: - [ ] Move to separate file

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -167,26 +174,79 @@ fun StartingView() {
...
+    onClick: () -> Unit,
+    content: @Composable RowScope.() -> Unit
+) {
+    Button(
+        onClick = onClick,
+        colors = ButtonColors(
+            containerColor = ts_color_light_blue500,
```
agottardo, line 241: - [ ] Should use darker shade in dark mode

barnstar (approved):
Wondering if we should include a complete tailscale material theme here, but this will do for now.

PR #204, opened 2024-03-14
android/ui: fix time formatting strings and main view states

barnstar:
Updates tailscale/corp#18202

Localizations and some logical cleanup for formatting "in x time_unit" strings for the node details.

Removes the "starting" screens and shows the peer details, search and exit node picker even when the VPN isn't running since we can do this on Android.

Moves the ComposableStringFormatter to a central location since it'll be used in a number of places.

Fixes showing your current node as connected (when you're connected) and shows the magicDNS domain names in the peer details.

Signed-off-by: Jonathan Nobels <jonathan@tailscale.com>

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
@@ -45,6 +45,7 @@ import androidx.compose.ui.unit.dp
 import com.tailscale.ipn.R
 import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.model.IpnLocal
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
agottardo, line 1: See the Slack thread about this. Design might want to still hide outdated information.

agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
barnstar: Yup, killed it for now.

PR #205, opened 2024-03-14
android: optimize peer search

barnstar:
Updates tailscale/corp#18202

Switch to LazyColumn so we're not redrawing the entire list.

Modify the search logic so we're searching progressively and doing all of the sorting and categorization up front on netmap changes.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
@@ -15,9 +15,8 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.rememberScrollState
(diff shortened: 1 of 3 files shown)


agottardo (approved):

raggi (approved):

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -218,47 +217,52 @@ fun PeerList(searchTerm: StateFlow<String>, peers: StateFlow<List<PeerSet>>, onN
...
-                                    } else {
-                                        Color.Gray
+                    item {
+                        ListItem(
+                                modifier = Modifier.clickable {
+                                    onNavigateToPeerDetails(peer)
+                                },
+                                headlineContent = {
+                                    Row(verticalAlignment = Alignment.CenterVertically) {
+                                        // By definition, SelfPeer is online since we will not show the peer list unless you're connected.
```
raggi, line 241: Technically you can be offline, but you won't know it by netmap - if you can't reach control, but you had previously managed to get a netmap and are locally "up"

PR #207, opened 2024-03-14
android/ui: use lazy columns and optimize the peer search logic

barnstar:
Updates tailscale/corp#18202

Swapped to lazy columns and updated the peer search logic to eliminate all of the categorization and sorting during the filter pass.  Nice and fast now.


barnstar:
Most of this was already reviewed, this is just some cruft that got lost in the rebase shuffle.

PR #210, opened 2024-03-15
build.gradle, Makefile: make build.gradle more portable

kari-ts:


pkg/tailscale/log.go
```diff
@@ -0,0 +1,91 @@
...
+	"bufio"
+	"log"
+	"os"
+	"path/filepath"
+	"runtime"
+	"syscall"
+	"unsafe"
+)
+
+// 1024 is the truncation limit from android/log.h, plus a \n.
```
raggi, line 24: the header says implementation defined, "e.g. 1023", though I believe this is closer to 4k (and has been for a very long time), but leaving space for some kernel and userspace header bits, around 4000 bytes in almost all implementations, subtracting whatever tag is set.

pkg/tailscale/log.go
```diff
@@ -0,0 +1,91 @@
...
+
+var logTag = C.CString(ID)
+
+func init() {
+	// Android's logcat already includes timestamps.
+	log.SetFlags(log.Flags() &^ log.LstdFlags)
+	log.SetOutput(new(androidLogWriter))
+
+	// Redirect stdout and stderr to the Android logger.
+	logFd(os.Stdout.Fd())
```
raggi, line 37: move these two up above the log package setup, so that if we crash we maximize the chance of logging about it.

pkg/tailscale/log.go
```diff
@@ -0,0 +1,91 @@
...
+			if err != nil {
+				break
+			}
+			copy(buf, line)
+			buf[len(line)] = 0
+			C.__android_log_write(C.ANDROID_LOG_INFO, logTag, cbuf)
+		}
+		// The garbage collector doesn't know that w's fd was dup'ed.
+		// Avoid finalizing w, and thereby avoid its finalizer closing its fd.
+		runtime.KeepAlive(w)
```
raggi, line 89: This keeps w alive until this goroutine exits, but it will be closed afterward.

This could later result in a deadlock in the application if for example we're crashing and something is writing to stderr on the way down - if we fill the pipe to the point of blocking, we'll be done, and the GC may not run soon enough to actually close the file. We shoudl replace this with a deferred close of both r and w inside the goroutine, and the defer will have the side effect of keeping them alive.

Go mobile takes a different approach to bind these as globals, but still could suffer from the described lock up: https://cs.opensource.google/go/x/mobile/+/master:internal/mobileinit/mobileinit_android.go?q=__android_log_write

raggi (commented):
I left a good amount of commentary that is on stuff I know that's been brought over from the existing app - no need to do it all.

Let's definitely get the license markers covered at minimum, and anything else that is relatively straightforward. We can cover other details in follow-ups if that makes sense.

It's great to see this working!

raggi (commented):
I left a good amount of commentary that is on stuff I know that's been brought over from the existing app - no need to do it all.

Let's definitely get the license markers covered at minimum, and anything else that is relatively straightforward. We can cover other details in follow-ups if that makes sense.

It's great to see this working!

pkg/tailscale/net.go
```diff
@@ -0,0 +1,346 @@
...
+//	https://github.com/tailscale/tailscale/issues/2180
+var errMultipleUsers = errors.New("VPN cannot be created on this device due to an Android bug with multiple users")
+
+// Report interfaces in the device in net.Interface format.
+func (a *App) getInterfaces() ([]interfaces.Interface, error) {
+	var ifaceString string
+	err := jnipkg.Do(a.jvm, func(env *jnipkg.Env) error {
+		cls := jnipkg.GetObjectClass(env, a.appCtx)
+		m := jnipkg.GetMethodID(env, cls, "getInterfacesAsString", "()Ljava/lang/String;")
+		n, err := jnipkg.CallObjectMethod(env, a.appCtx, m)
```
raggi, line 44: this string needs to be released once we're done copying it

pkg/tailscale/callbacks.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"unsafe"
+
+	jnipkg "github.com/tailscale/tailscale-android/pkg/jni"
+)
+
+// #include <jni.h>
+import "C"
+
+var (
```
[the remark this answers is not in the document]
kari-ts: sounds good. I think a lot of these will just go away too when we switch to having kotlin call localapiclient to editprefs instead of having Android use a chan to send an event

pkg/tailscale/log.go
```diff
@@ -0,0 +1,91 @@
...
+	logFd(os.Stdout.Fd())
+	logFd(os.Stderr.Fd())
+}
+
+type androidLogWriter struct {
+	// buf has room for the maximum log line, plus a terminating '\0'.
+	buf [logLineLimit + 1]byte
+}
+
+func (w *androidLogWriter) Write(data []byte) (int, error) {
```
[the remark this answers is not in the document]
kari-ts: I took a stab at this in https://github.com/tailscale/tailscale-android/pull/206, let me know if it looks ok

pkg/tailscale/net.go
kari-ts: I didn't change this here because we're getting rid of a lot of this in our PR to use gomobile

PR #212, opened 2024-03-17
android: simplify local API client

oxtoacart:
Updates tailscale/corp#18202


android/src/main/java/com/tailscale/ipn/ui/localapi/Result.kt
```diff
@@ -1,32 +0,0 @@
-// Copyright (c) Tailscale Inc & AUTHORS
-// SPDX-License-Identifier: BSD-3-Clause
-
-package com.tailscale.ipn.ui.localapi
-
-// Go-like result type with an optional value and an optional Error
-// This guarantees that only one of the two is non-null
-class Result<T> {
```
oxtoacart, line 8: Kotlin already has a Result type, I've switched to using that.

android/src/main/java/com/tailscale/ipn/ui/localapi/Client.kt
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    // This is called from the JNI layer to publish responses.
+    @OptIn(ExperimentalSerializationApi::class)
+    @Suppress("unused", "UNCHECKED_CAST")
+    fun onResponse(respData: ByteArray) {
+        Log.d(TAG, "Response for request: $fullPath")
+
+        val response: Result<T> = when (responseType) {
+            typeOf<String>() -> Result.success(respData.decodeToString() as T)
```
oxtoacart, line 206: Instead of each endpoint's response handler specifying its own decoder, we now just decode appropriately based on the datatype that the response handler requires.

android/src/main/java/com/tailscale/ipn/ui/viewModel/BugReportViewModel.kt
```diff
@@ -5,24 +5,19 @@ package com.tailscale.ipn.ui.viewModel
...
 
     init {
-        viewModelScope.launch {
-            localAPI.getBugReportId {
-                when (it.successful) {
-                    true -> bugReportID.set(it.success ?: "(Error fetching ID)")
-                    false -> bugReportID.set("(Error fetching ID)")
-                }
-            }
+        Client(viewModelScope).bugReportId { result ->
```
oxtoacart, line 18: Instead of having a stateful reference to a client, we now just instantiate clients when we need to use them, using the appropriate scope at that point.

cmd/localapiservice/localapishim.go
```diff
@@ -76,10 +75,9 @@ func Java_com_tailscale_ipn_ui_localapi_LocalApiClient_doRequest(
 
 	jrespBody := jni.NewByteArray(jenv, resp)
 	respBody := jni.Value(jrespBody)
-	cookie := jni.Value(jcookie)
-	onResponse := jni.GetMethodID(jenv, shim.clientClass, "onResponse", "([BLjava/lang/String;)V")
+	onResponse := jni.GetMethodID(jenv, shim.clientClass, "onResponse", "([B)V")
 
-	jni.CallVoidMethod(jenv, jni.Object(cls), onResponse, respBody, cookie)
+	jni.CallVoidMethod(jenv, jni.Object(cls), onResponse, respBody)
```
oxtoacart, line 80: Note that we no longer need request ids/cookies, each request is its own instance, so Go calls back into that instances' onResponse.

cmd/localapiservice/localapishim.go
```diff
@@ -57,16 +56,16 @@ func Java_com_tailscale_ipn_ui_localapi_LocalApiClient_doRequest(
...
-	pathRef := jni.NewGlobalRef(jenv, jni.Object(jpath))
-	pathStr := jni.GoString(jenv, jni.String(pathRef))
-	defer jni.DeleteGlobalRef(jenv, pathRef)
-
 	// The HTTP verb
 	methodRef := jni.NewGlobalRef(jenv, jni.Object(jmethod))
 	methodStr := jni.GoString(jenv, jni.String(methodRef))
 	defer jni.DeleteGlobalRef(jenv, methodRef)
 
+	// The API Path
```
oxtoacart, line 64: Bikeshedding - idiomatically, method comes before path.

agottardo (approved):

PR #213, opened 2024-03-18
android: ViewModel cleanup

oxtoacart:
- Replace IpnManager, IpnModel and PrefsEditor with IpnViewModel
- Use lazy StateFlows in Notifier
- Manage view model lifecycles using viewModel() function
- Stop watching IPN bus when view model is cleaned up
- Pass IPN notifications as ByteArray instead of string

Updates tailscale/corp#18202


android/src/main/java/com/tailscale/ipn/ui/view/SettingsView.kt
```diff
@@ -44,18 +45,16 @@ import com.tailscale.ipn.ui.util.defaultPaddingModifier
...
-    val onNavigateToAbout: () -> Unit,
-    val onNavigateToMDMSettings: () -> Unit,
-    val onNavigateToManagedBy: () -> Unit,
-)
-
 @Composable
-fun Settings(viewModel: SettingsViewModel) {
+fun Settings(
+    settingsNav: SettingsNav,
+    viewModel: SettingsViewModel = viewModel(factory = SettingsViewModelFactory(settingsNav))
```
oxtoacart, line 56: Notice how we use the `viewModel()` function to manage the lifecycle of the view model, and use `SettingsViewModelFactory` to allow us to inject the `settingsNav`.

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -64,62 +59,49 @@ class MainActivity : ComponentActivity() {
                             onNavigateToMDMSettings = { navController.navigate("mdmSettings") },
                             onNavigateToManagedBy = { navController.navigate("managedBy") })
 
-                    composable("main") {
-                        MainView(
-                            viewModel = MainViewModel(manager.model, manager),
```
oxtoacart, line 69: Notice that we don't construct any view models inside of the MainActivity anymore.

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -64,62 +59,49 @@ class MainActivity : ComponentActivity() {
...
-                            navigation = mainViewNav
+                    val exitNodePickerNav = ExitNodePickerNav(onNavigateHome = {
+                        navController.popBackStack(
```
oxtoacart, line 63: In my OG implementation of the exit node picker, I navigated back to home by using `navigage()`. This would push a new home page onto the stack without freeing the resources used by the old one. Using `popBackStack()` instead unwinds the stack and reuses the existing one, which also has the benefit of it loading faster.

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -0,0 +1,140 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+
+/**
+ * Base model for most models in this application. Provides common facilities for watching IPN
+ * notifications, managing login/logout, updating preferences, etc.
+ */
+open class IpnViewModel : ViewModel() {
+    companion object {
+        val mdmSettings: StateFlow<MDMSettings> = MutableStateFlow(MDMSettings())
```
oxtoacart, line 25: mdmSettings are now modeled as a StateFlow for consistency with how we manage other state.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -4,19 +4,23 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
+import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
-
-typealias NotifierCallback = (Notify) -> Unit
-
-
-class Watcher(
```
oxtoacart, line 17: Didn't need a class for this.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -4,19 +4,23 @@
...
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
+import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
-
-typealias NotifierCallback = (Notify) -> Unit
```
oxtoacart, line 14: No longer part of public API, so no need for type alias.

android/src/main/java/com/tailscale/ipn/ui/util/LoadingIndicator.kt
```diff
@@ -33,8 +33,8 @@ object LoadingIndicator {
             contentAlignment = Alignment.Center,
         ) {
             content()
-            val isLoading = loading.collectAsState()
-            if (isLoading.value) {
+            val isLoading = loading.collectAsState().value
```
oxtoacart, line 36: Drive-by, cleaner syntax.

android/src/main/java/com/tailscale/ipn/ui/viewModel/ExitNodePickerViewModel.kt
```diff
@@ -6,23 +6,30 @@ package com.tailscale.ipn.ui.viewModel
...
 import com.tailscale.ipn.ui.util.set
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import java.util.TreeMap
 
-class ExitNodePickerViewModel(private val model: IpnModel, private val onNavigateHome: () -> Unit) :
-    ViewModel() {
-    companion object {
-        const val TAG = "ExitNodePickerViewModel"
+data class ExitNodePickerNav(
```
oxtoacart, line 20: This now follows the pattern that other models already followed w.r.t. navigation.

android/src/main/java/com/tailscale/ipn/ui/viewModel/SettingsViewModel.kt
```diff
@@ -43,107 +42,116 @@ data class SettingBundle(val title: String? = null, val settings: List<Setting>)
...
+                    )
                 )
-            } else {
-                null
             }
-    )
+        }
+    }
 
-    val settings: List<SettingBundle> = listOf(
```
oxtoacart, line 140: This has been replaced by a StateFlow.

cmd/localapiservice/localapishim.go
```diff
@@ -198,8 +198,8 @@ func Java_com_tailscale_ipn_ui_notifier_Notifier_startIPNBusWatcher(
 			return true
 		}
 		jni.Do(shim.jvm, func(env *jni.Env) error {
-			jjson := jni.JavaString(env, string(js))
-			onNotify := jni.GetMethodID(env, shim.notifierClass, "onNotify", "(Ljava/lang/String;Ljava/lang/String;)V")
+			jjson := jni.NewByteArray(jenv, js)
```
oxtoacart, line 201: Passing ByteArray instead of a String seemed to help quite a bit with memory usage.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -47,95 +41,99 @@ class Notifier(private val scope: CoroutineScope) {
...
             isReady.await()
-            Log.d("Notifier", "Starting IPN Bus watcher for sessionid: ${sessionId}")
+            Log.d(
+                "Notifier",
+                "Starting IPN Bus watcher for sessionid: ${sessionId} in job ${scope.coroutineContext.job}"
+            )
             startIPNBusWatcher(sessionId, mask)
-            watchers.remove(sessionId)
-            Log.d("Notifier", "IPN Bus watcher for sessionid:${sessionId} has halted")
+            scope.coroutineContext.job.invokeOnCompletion {
```
oxtoacart, line 131: This ties us to the scope's lifecycle, so that if the scope's job is canceled/finishes, we'll clean up watcher.

android/src/main/java/com/tailscale/ipn/ui/view/PeerDetails.kt
```diff
@@ -28,55 +28,73 @@ import androidx.compose.ui.platform.LocalClipboardManager
...
+import androidx.lifecycle.viewmodel.compose.viewModel
 import com.tailscale.ipn.R
 import com.tailscale.ipn.ui.util.settingsRowModifier
 import com.tailscale.ipn.ui.viewModel.PeerDetailsViewModel
+import com.tailscale.ipn.ui.viewModel.PeerDetailsViewModelFactory
 
 
 @Composable
-fun PeerDetails(viewModel: PeerDetailsViewModel) {
+fun PeerDetails(
```
barnstar, line 39: This isn't working.  On initial load, the peer is null and the netmap is null.   The netmap value on the model needs to be observed.

android/src/main/java/com/tailscale/ipn/ui/viewModel/PeerDetailsViewModel.kt
```diff
@@ -16,7 +16,13 @@ import com.tailscale.ipn.ui.util.TimeUtil
 data class PeerSettingInfo(val titleRes: Int, val value: ComposableStringFormatter)
 
 
-class PeerDetailsViewModel(val model: IpnModel, val nodeId: StableNodeID) : ViewModel() {
+class PeerDetailsViewModelFactory(private val nodeId: StableNodeID) : ViewModelProvider.Factory {
+    override fun <T : ViewModel> create(modelClass: Class<T>): T {
+        return PeerDetailsViewModel(nodeId) as T
```
barnstar, line 21: The OOM errors can be traced here (partially).  I was using a shared notifier and a shared netmap amongst all of the components and even then, I was running into OOM errors.  You're creating a new notifier for each view model which just compounds the issue.    Each netmap update on corp is like 3Mb of JSON to parse - and you're going to need to fetch that, parse it, just to extract the single node from it here.

We need to share the notifiers and the parsing amongst all the view models.  There is only one netmap and we should observe it in only one place.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -47,95 +41,99 @@ class Notifier(private val scope: CoroutineScope) {
...
     private external fun stopIPNBusWatcher(sessionId: String)
 
-    private var watchers = HashMap<String, Watcher>()
-
     // Callback from jni when a new notification is received
-    fun onNotify(notification: String, sessionId: String) {
-        val notify = decoder.decodeFromString<Notify>(notification)
+    @OptIn(ExperimentalSerializationApi::class)
+    fun onNotify(notification: ByteArray, sessionId: String) {
+        val notify = decoder.decodeFromStream<Notify>(notification.inputStream())
```
barnstar, line 102: Re OOM, I'm wondering if the large allocations on corp are what's killing us.   In go land, this is a Reader, but we're just reading the entire works, then allocating mountains of strings to parse everything, etc.   Chunking onNotify might help with the large allocations in JNI - which is what seems to be failing.

barnstar (changes requested):

android/src/main/java/com/tailscale/ipn/ui/view/PeerDetails.kt
oxtoacart: This seems to be working now.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
barnstar: Mainly habit... Complex closure types look terrible in method declarations, so I generally typealias them, but I'm inconsistent about it....

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -26,116 +26,72 @@ class Watcher(
...
-    private fun unwatchIPNBus(sessionId: String) {
-        stopIPNBusWatcher(sessionId)
-    }
-
-    // Cancels all watchers
-    fun cancelAllWatchers() {
-        for (sessionId in watchers.values.map({ it.sessionId })) {
-            unwatchIPNBus(sessionId)
+            val mask =
+                NotifyWatchOpt.Netmap.value or NotifyWatchOpt.Prefs.value or NotifyWatchOpt.InitialState.value
```
barnstar, line 57: Nit: A comment here explaining the mask choice might be useful, but it should be easy enough for the next guy to figure out.

android/src/main/java/com/tailscale/ipn/ui/view/SettingsView.kt
barnstar: Wouldn't be Java without a Factory!

barnstar (approved):
Much cleaner....

oxtoacart:
Thanks for the review @barnstar !

PR #215, opened 2024-03-18
android: remove unused code to reduce JSON deserialization overhead

oxtoacart:
Updates tailscale/corp#18202


oxtoacart:
In my testing this didn't make a huge difference, but if we're not using types/fields, might as well not bother deserializing them.

android/src/main/java/com/tailscale/ipn/ui/model/IpnState.kt
```diff
@@ -5,114 +5,15 @@ package com.tailscale.ipn.ui.model
 
 import kotlinx.serialization.Serializable
 
-class IpnState {
```
barnstar, line 8: We don't use this now but there are API endpoints that are yet to be implemented that will require us to fetch PeerStatus (for example). 

Removing this doesn't really gain us much.  Commenting out the unused fields will.

android/src/main/java/com/tailscale/ipn/ui/model/NetMap.kt
```diff
@@ -5,51 +5,49 @@ package com.tailscale.ipn.ui.model
...
-class Netmap {
-    @Serializable
-    data class NetworkMap(
-            var SelfNode: Tailcfg.Node,
-            var NodeKey: KeyNodePublic,
-            var Peers: List<Tailcfg.Node>? = null,
-            var Expiry: Time,
-            var Domain: String,
-            var UserProfiles: Map<String, Tailcfg.UserProfile>,
-            var TKAEnabled: Boolean,
```
barnstar, line 17: We'll need some of these fields shortly too.   TAKEnabled before launch.  Domain is used on iOS to disambiguate accounts and tailnet name, etc.

barnstar (changes requested):
We should leave in the (currently) unused types and just comment out the unused fields to reduce the parsing overhead.  A couple of the ones you removed will just be going right back in when we get to tailnet lock and possibly as we refine some of the UI.

PR #225, opened 2024-03-20
android: fix connect

kari-ts:
Kotlinize IPNService and App
Call connect in IPNService
Add observers for readiness to prepare VPN, and quick tile readiness Start Notifier in App, since new state flows need to be observed outside of activity lifecycle

Next: fixing quick tiles

Updates tailscale/corp#18202


android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    override fun newBuilder(): VPNServiceBuilder {
+        val b: Builder = Builder()
+                .setConfigureIntent(configIntent())
+                .allowFamily(OsConstants.AF_INET)
+                .allowFamily(OsConstants.AF_INET6)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) b.setMetered(false) // Inherit the metered status from the underlying networks.
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) b.setUnderlyingNetworks(null) // Use all available networks.
+
+        // RCS/Jibe https://github.com/tailscale/tailscale/issues/2322
```
barnstar, line 82: These should probably be listed in a manifest somewhere - maybe just a file we include in the apk for now.

Really, we should fetch this list, and apply it dynamically since it seems like it grows over time but we can save that for another day.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -26,67 +26,75 @@ import kotlinx.serialization.json.decodeFromStream
...
-  val state: StateFlow<Ipn.State> = MutableStateFlow(Ipn.State.NoState)
-  val netmap: StateFlow<Netmap.NetworkMap?> = MutableStateFlow(null)
-  val prefs: StateFlow<Ipn.Prefs?> = MutableStateFlow(null)
-  val engineStatus: StateFlow<Ipn.EngineStatus?> = MutableStateFlow(null)
-  val tailFSShares: StateFlow<Map<String, String>?> = MutableStateFlow(null)
-  val browseToURL: StateFlow<String?> = MutableStateFlow(null)
-  val loginFinished: StateFlow<String?> = MutableStateFlow(null)
-  val version: StateFlow<String?> = MutableStateFlow(null)
-  val vpnPermissionGranted: StateFlow<Boolean?> = MutableStateFlow(null)
+    val state: StateFlow<Ipn.State> = MutableStateFlow(Ipn.State.NoState)
```
barnstar, line 32: Enable ktfmt... Indents should should all be 2 spaces.

android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -0,0 +1,130 @@
...
+import android.net.VpnService
+import android.os.Build
+import android.system.OsConstants
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import libtailscale.Libtailscale
+import java.util.UUID
+
+open class IPNService : VpnService(), libtailscale.IPNService {
+    private val randomID: String = UUID.randomUUID().toString()
```
barnstar, line 17: ktfmt again... It sets indents at 2 spaces, not 4.

barnstar (changes requested):

android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -0,0 +1,137 @@
...
+
+open class IPNService : VpnService(), libtailscale.IPNService {
+  private val randomID: String = UUID.randomUUID().toString()
+
+  override fun id(): String {
+    return randomID
+  }
+
+  override fun onStartCommand(intent: Intent, flags: Int, startId: Int): Int {
+    if (intent != null && ACTION_STOP_VPN == intent.getAction()) {
```
barnstar, line 24: intent is always non-null here... and below.

kari-ts: the signature implies that intent is non nullable, but [onStartCommand](https://developer.android.com/reference/android/app/Service#onStartCommand(android.content.Intent,%20int,%20int)) can be called with a null intent, like when a service is restarted after being killed by the system and we still have to explicitly check the null case. I think this is just an example of kt's type system not mapped to java's and Android framework methods being defined in Java

barnstar:
This code is all included in https://github.com/tailscale/tailscale-android/pull/228/files

You can just close this, we'll pick it up there.

PR #228, opened 2024-03-20
android: disconnect

kari-ts:
Use localapi to disconnect

Updates https://github.com/tailscale/corp/issues/18202


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -0,0 +1,478 @@
...
+import java.util.Locale
+import java.util.Objects
+import com.tailscale.ipn.App;
+
+class App : Application(), libtailscale.AppContext {
+    private val applicationScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+    var dnsConfigObj = DnsConfig()
+
+    companion object {
+        const val STATUS_CHANNEL_ID = "tailscale-status"
```
oxtoacart, line 71: Looks like it needs a pass of `ktfmt`.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -0,0 +1,478 @@
...
+import java.io.IOException
+import java.net.InetAddress
+import java.net.NetworkInterface
+import java.security.GeneralSecurityException
+import java.util.Locale
+import java.util.Objects
+import com.tailscale.ipn.App;
+
+class App : Application(), libtailscale.AppContext {
+    private val applicationScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
oxtoacart, line 67: Instead of creating our own scope, we can use `kotlinx.coroutines.MainScope()`.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -0,0 +1,478 @@
...
+        Notifier.stop()
+        applicationScope.cancel()
+    }
+
+    fun connect() {
+        val callback: (Result<Ipn.Prefs>) -> Unit = { result ->
+            if (result.isSuccess) {
+                val prefs = result.getOrNull()
+                prefs?.let {
+                    Log.d("App", "Connect: preferences updated successfully: $prefs")
```
oxtoacart, line 135: Style nit: Please move "App" into a `TAG` constant

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -26,67 +26,75 @@ import kotlinx.serialization.json.decodeFromStream
...
-  val state: StateFlow<Ipn.State> = MutableStateFlow(Ipn.State.NoState)
-  val netmap: StateFlow<Netmap.NetworkMap?> = MutableStateFlow(null)
-  val prefs: StateFlow<Ipn.Prefs?> = MutableStateFlow(null)
-  val engineStatus: StateFlow<Ipn.EngineStatus?> = MutableStateFlow(null)
-  val tailFSShares: StateFlow<Map<String, String>?> = MutableStateFlow(null)
-  val browseToURL: StateFlow<String?> = MutableStateFlow(null)
-  val loginFinished: StateFlow<String?> = MutableStateFlow(null)
-  val version: StateFlow<String?> = MutableStateFlow(null)
-  val vpnPermissionGranted: StateFlow<Boolean?> = MutableStateFlow(null)
+    val state: StateFlow<Ipn.State> = MutableStateFlow(Ipn.State.NoState)
```
oxtoacart, line 32: ktmft

android/src/play/java/com/tailscale/ipn/Google.java
```diff
@@ -28,7 +28,7 @@ static void googleSignIn(Activity act, String serverOAuthID, int reqCode) {
 					.build();
 				GoogleSignInClient client = GoogleSignIn.getClient(act, gso);
 				Intent signInIntent = client.getSignInIntent();
-				App.startActivityForResult(act, signInIntent, reqCode);
+				App.Companion.startActivityForResult(act, signInIntent, reqCode);
```
oxtoacart, line 31: If you don't want to have to use Companion here, you could annotate `startActivityForResult` with `@JVMStatic`.

oxtoacart (approved):
Looks great! I just have a few stylistic nits, but nothing functional.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -0,0 +1,478 @@
...
+            } else {
+                val error = result.exceptionOrNull()
+                Log.e("App", "Failed to update preferences: ${error?.message}")
+            }
+        }
+        setWantRunning(true, callback)
+    }
+
+    fun disconnect() {
+        val callback: (Result<Ipn.Prefs>) -> Unit = { result ->
```
barnstar, line 146: Nit.. If all we're going to do is log, we can probably just define a function somewhere that does it.  I don't think you need to log the success case either.

PR #230, opened 2024-03-21
android: add UI to toggle running as exit node on/off

agottardo:
This PR provides UI to let the user toggle `AdvertisedRoutes` by adding/removing the zero routes, with a view to warn the user about battery life impact and potential cellular data charges. 

Language and graphics mimic what we currently have on tvOS, final designs will follow as per @sonovawolf.

The exit node picker is also updated with a Run as Exit Node item at the top of the list.


barnstar:
You can retarget this to main now.

kari-ts (approved):

PR #231, opened 2024-03-21
android: add Tailnet lock setup UI

agottardo:
Adds a view to see the Tailnet lock settings and copy node key and public key, resembling the iOS and macOS ones.


android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
```diff
@@ -0,0 +1,118 @@
...
+                  }
+            }
+
+            item(key = "nodeKey") {
+              Column(
+                  verticalArrangement = Arrangement.spacedBy(8.dp),
+                  modifier = Modifier.padding(horizontal = 16.dp)) {
+                    Spacer(Modifier.size(4.dp))
+                    ClipboardValueView(
+                        value = nodeKey,
```
kari-ts, line 74: this can be null right? since we do nodeKey.set(result.getOrNull()?.NodeKey)
should this check if null and have a fallback value if null?

android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
```diff
@@ -0,0 +1,118 @@
...
+                        subtitle = stringResource(R.string.node_key_explainer))
+                  }
+            }
+
+            item(key = "tailnetLockKey") {
+              Column(
+                  verticalArrangement = Arrangement.spacedBy(8.dp),
+                  modifier = Modifier.padding(horizontal = 16.dp)) {
+                    ClipboardValueView(
+                        value = tailnetLockKey,
```
kari-ts, line 85: same here

android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
agottardo: Fixed

android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
agottardo: Fixed

kari-ts (approved):

PR #232, opened 2024-03-22
android: fix vpn

kari-ts:
-Move most of prepare and establish VPN logic out of Go into Android 
-Fix prepareVPN argument to use request codes to differentiate sign in and prepare VPN 
-Fix missing adapter implementation (setMtu)

Updates tailscale/corp#18202

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/Peer.java
+++ b/android/src/main/java/com/tailscale/ipn/Peer.java
@@ -6,11 +6,32 @@ package com.tailscale.ipn;
 import android.app.Fragment;
 import android.content.Intent;
 
-import libtailscale.Libtailscale;
-
 public class Peer extends Fragment {
+
+    private static int resultOK = -1;
+
+    public class RequestCodes {
+        public static final int requestSignin = 1000;
+        public static final int requestPrepareVPN = 1001;
+    }
+
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        Libtailscale.onActivityResult(requestCode, resultCode, MaybeGoogle.getIdTokenForActivity(getActivity()));
+        switch (requestCode) {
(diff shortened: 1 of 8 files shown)


agottardo (approved):

libtailscale/interfaces.go
```diff
@@ -120,32 +120,10 @@ type InputStream interface {
 // The below are global callbacks that allow the Java application to notify Go
 // of various state changes.
 
-func OnVPNPrepared() {
```
oxtoacart, line 123: Nice!

libtailscale/interfaces.go
```diff
@@ -120,32 +120,10 @@ type InputStream interface {
...
-
 func RequestVPN(service IPNService) {
 	onVPNRequested <- service
 }
 
 func ServiceDisconnect(service IPNService) {
 	onDisconnect <- service
 }
-
-func OnActivityResult(reqCode, resCode int, idToken string) {
```
oxtoacart, line 135: Nice!

android/src/main/java/com/tailscale/ipn/Peer.java
```diff
@@ -6,11 +6,32 @@
...
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        Libtailscale.onActivityResult(requestCode, resultCode, MaybeGoogle.getIdTokenForActivity(getActivity()));
+        switch (requestCode) {
+            case RequestCodes.requestSignin:
+                if (resultCode != resultOK) {
+                    // TODO: send null Google token
+                    break;
+                }
+                // TODO: send Google token
```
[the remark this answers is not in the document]
kari-ts: Browser based, but I do want to investigate what we were using the google token for and make sure that we don't break anything by removing it

PR #233, opened 2024-03-22
android: add DNS Settings view

agottardo:
Updates ENG-2990

This PR adds a DNS Settings view with the same functionality and items as the iOS one. It also moves the 'Use Tailscale DNS Settings' item out of the main settings view into the detail view.

### DIFF
--- a/android/src/main/res/values/strings.xml
+++ b/android/src/main/res/values/strings.xml
@@ -115,5 +115,14 @@
     <string name="tailnet_lock_key_explainer">Used to authorize changes to the Tailnet lock configuration.</string>
     <string name="bug_report_id">Bug Report ID</string>
     <string name="learn_more">Learn more…</string>
+    <string name="dns_settings">DNS Settings</string>
+    <string name="using_tailscale_dns">Using Tailscale DNS</string>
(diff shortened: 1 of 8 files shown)


android/src/main/res/values/strings.xml
```diff
@@ -115,5 +115,15 @@
     <string name="tailnet_lock_key_explainer">Used to authorize changes to the Tailnet lock configuration.</string>
     <string name="bug_report_id">Bug Report ID</string>
     <string name="learn_more">Learn more…</string>
+    <string name="dns_settings">DNS Settings</string>
+    <string name="using_tailscale_dns">Using Tailscale DNS</string>
+    <string name="this_device_is_using_tailscale_to_resolve_dns_names">This device is using Tailscale to resolve DNS names.</string>
```
raggi, line 120: this seems to be a duplicate of the entry on line 125 - do we need two separate strings?

raggi (approved):

android/src/main/res/values/strings.xml
agottardo: nice catch!

PR #238, opened 2024-03-23
Makefile: add help text to make commands

willnorris:
Add `make help` command, which prints available commands and their description.  I'm not 100% sure which commands should be documented, so took an initial attempt at it.

This also makes `make help` the default command, aligning it with the Makefiles in corp and oss.

Updates tailscale/corp#18202


raggi (approved):

raggi:
Thanks - this is a great start. I think we'll do another big refactor once we're down to one app again

oxtoacart:
Thanks @willnorris !

PR #241, opened 2024-03-25
android: add ExitNodeAllowLANAccess toggle in exit node picker

agottardo:
Updates ENG-3011

Just like on iOS, we should show a switch at the bottom of the exit node picker to toggle the `ExitNodeAllowLANAccess` preference.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/ExitNodePickerViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/ExitNodePickerViewModel.kt
@@ -6,6 +6,7 @@ package com.tailscale.ipn.ui.viewModel
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
+import com.tailscale.ipn.R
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/ExitNodePickerViewModel.kt
```diff
@@ -51,15 +52,28 @@ class ExitNodePickerViewModel(private val nav: ExitNodePickerNav) : IpnViewModel
...
+  val allowLANAccessSetting =
+      Setting(
+          R.string.allow_lan_access,
+          SettingType.SWITCH,
+          isOn = MutableStateFlow(Notifier.prefs.value?.ExitNodeAllowLANAccess),
+          onToggle = {
+            LoadingIndicator.start()
+            toggleAllowLANAccess {
+              LoadingIndicator.stop()
+              // (jonathan) TODO: Error handling
```
barnstar, line 66: We can probably swallow the errors for all of these.  There's nothing the user can do, and LocalAPI is not going to fail in the ways it can on iOS.

barnstar (approved):

PR #243, opened 2024-03-25
android: style exit node picker per Material UI and add disable button

oxtoacart:
Updates #ENG-2911

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/util/ClickableOrGrayedOut.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/util/ClickableOrGrayedOut.kt
@@ -1,0 +1,27 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package com.tailscale.ipn.ui.util
+
+import androidx.compose.foundation.clickable
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.semantics.Role
+
+/**
+ * Similar to Modifier.clickable, but if enabled == false, this adds a 75% alpha to make disabled
+ * items appear grayed out.
+ */
+@Composable
+fun Modifier.clickableOrGrayedOut(
+    enabled: Boolean = true,
+    onClickLabel: String? = null,
+    role: Role? = null,
+    onClick: () -> Unit
+) =
+    if (enabled) {
+      clickable(onClickLabel = onClickLabel, role = role, onClick = onClick)
+    } else {
+      alpha(0.75f)
+    }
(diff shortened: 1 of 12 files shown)


oxtoacart:
I need to do a bit more work on this ...

android/src/main/java/com/tailscale/ipn/ui/util/LoadingIndicator.kt
```diff
@@ -35,12 +38,12 @@ object LoadingIndicator {
...
       val isLoading = loading.collectAsState().value
       if (isLoading) {
-        Box(Modifier.matchParentSize().background(Color.Gray.copy(alpha = 0.5f)))
+        Box(Modifier.clickable {}.matchParentSize().background(Color.Gray.copy(alpha = 0.5f)))
 
         Column(
             modifier = Modifier.fillMaxWidth(),
             horizontalAlignment = Alignment.CenterHorizontally) {
-              CircularProgressIndicator()
+              TailscaleLogoView(true, Modifier.size(72.dp))
```
barnstar, line 46: Ha... Nice.

android/src/main/java/com/tailscale/ipn/ui/util/ClickableOrGrayedOut.kt
```diff
@@ -0,0 +1,27 @@
...
+fun Modifier.clickableOrGrayedOut(
+    enabled: Boolean = true,
+    onClickLabel: String? = null,
+    role: Role? = null,
+    onClick: () -> Unit
+) =
+    if (enabled) {
+      clickable(onClickLabel = onClickLabel, role = role, onClick = onClick)
+    } else {
+      alpha(0.75f)
```
barnstar, line 26: This should probably be an explicit colour, but it's fine for now.  We'll see what UX says.

android/src/main/java/com/tailscale/ipn/ui/view/ExitNodePicker.kt
```diff
@@ -116,36 +78,43 @@ fun ExitNodePicker(
...
-              Text(stringResource(R.string.offline))
+              Icon(Icons.Outlined.Check, contentDescription = stringResource(R.string.selected))
             }
           }
         })
   }
 }
 
+@Composable
+fun MullvadItem(nav: ExitNodePickerNav, count: Int, selected: Boolean) {
```
barnstar, line 104: Nit: We should probably use a more generic term than "Mullvad" - considering that these are just exit nodes and, in theory, could be provided by any number of sources.

agottardo: We say `Location-based exit nodes` on other platforms. Mullvad happens to be one of the options for it.

agottardo (approved):

PR #247, opened 2024-03-26
android/mdm: enforce MDM settings in UI items

agottardo:
Fixes ENG-2851

Uses our MDM settings in any place where we are drawing UI that should be hidden/disabled depending on the selected system policy value.

The ForceEnabled implementation lacks a UI component to warn the user when disablement is not allowed, but we can take care of that at a later time.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/DNSSettingsViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/DNSSettingsViewModel.kt
@@ -10,6 +10,8 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
 import com.tailscale.ipn.R
+import com.tailscale.ipn.mdm.AlwaysNeverUserDecidesSetting
+import com.tailscale.ipn.mdm.AlwaysNeverUserDecidesValue
 import com.tailscale.ipn.ui.localapi.Client
 import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.model.Tailcfg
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/ExitNodePickerViewModel.kt
```diff
@@ -62,7 +64,11 @@ class ExitNodePickerViewModel(private val nav: ExitNodePickerNav) : IpnViewModel
           R.string.allow_lan_access,
           SettingType.SWITCH,
           isOn = MutableStateFlow(Notifier.prefs.value?.ExitNodeAllowLANAccess),
-          enabled = MutableStateFlow(true),
+          enabled =
+              MutableStateFlow(
+                  IpnViewModel.mdmSettings.value.get(
```
oxtoacart, line 69: I don't think this will do what you want. If the value of mdmSettings changes, this stateflow will not update. I think you want to use combine, like [here](https://github.com/tailscale/tailscale-android/blob/cf6a203f7ae9faf154541fed036ceec05e6e0b62/android/src/main/java/com/tailscale/ipn/ui/viewModel/DNSSettingsViewModel.kt#L51).

android/src/main/java/com/tailscale/ipn/ui/viewModel/DNSSettingsViewModel.kt
```diff
@@ -41,6 +43,11 @@ class DNSSettingsViewModel() : IpnViewModel() {
           R.string.use_ts_dns,
           SettingType.SWITCH,
           isOn = MutableStateFlow(Notifier.prefs.value?.CorpDNS),
+          enabled =
+              MutableStateFlow(
+                  IpnViewModel.mdmSettings.value.get(
```
oxtoacart, line 48: Same issue here, probably want to use `combine()`.

barnstar (commented):
There are some pending changes to the way settings are rendered which affect this.

agottardo:
> There are some pending changes to the way settings are rendered which affect this.  

Yep, I'll need to revisit this. I also want to optimize performance a bit in the MDMSettings class. 

We're currently reading all values in realtime, but we should really cache them and invalidate when the system notifies us some restriction have changed.

agottardo:
Closing this, will revisit with Percy's new changes.

PR #251, opened 2024-03-27
android/taildrop: support direct mode for incoming taildrop

barnstar:
Updates tailscale/corp#18202

Implements direct mode support for incoming taildrop files.  None of the localAPI endpoints are implemented here but this will get taildrop files to the right places.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -109,7 +109,14 @@ class App : Application(), libtailscale.AppContext {
   override fun onCreate() {
     super.onCreate()
     val dataDir = this.filesDir.absolutePath
-    app = Libtailscale.start(dataDir, this)
+
+    // Set this to enable direct mode for taildrop whereby downloads will be saved directly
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -464,4 +471,27 @@ class App : Application(), libtailscale.AppContext {
       }
     }
   }
+
+  fun prepareDownloadsFolder(): File {
+    var downloads = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS)
```
agottardo, line 476: nit for now: looks like `getExternalStoragePublicDirectory` was deprecated in API level 29. Some discussion [here](https://stackoverflow.com/questions/56468539/getexternalstoragepublicdirectory-deprecated-in-android-q) on replacements.

agottardo (approved):

oxtoacart (commented):
Confirmed, it works for me. I love that this adds functionality and removes 11 lines of code net!

oxtoacart (approved):

PR #252, opened 2024-03-27
android/ui: add back fast user switching status and clean up avatar s…

oxtoacart:
…caling

Updates #cleanup


barnstar (approved):
Looks like a Tailscale client!  🚀

PR #253, opened 2024-03-27
android/ui: add one-time intro screen

barnstar:
fixes ENG-2910

adds a one-time intro screen mostly identical to the one presented in the legacy app.

### DIFF
--- a/android/src/main/res/values/strings.xml
+++ b/android/src/main/res/values/strings.xml
@@ -201,7 +201,7 @@
     <string name="taildrop_share_failed_title">Taildrop Failed</string>
 
     <!-- Strings for the intro screen -->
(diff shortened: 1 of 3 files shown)


android/src/main/res/values/strings.xml
```diff
@@ -200,4 +200,9 @@
     <string name="share_device_not_connected_title">Not Connected</string>
     <string name="taildrop_share_failed_title">Taildrop Failed</string>
 
+    <!-- Strings for the intro screen -->
+    <string name="tailscale">tailscale</string>
```
agottardo, line 204:
```suggestion
    <string name="tailscale">Tailscale</string>
```

agottardo (approved):

PR #262, opened 2024-03-28
android/ui: speed up loading of SettingsView

oxtoacart:
Updates tailscale/corp#18202

@sonovawolf noticed that the Settings view was slow to load. In the logs, I was able to see these kinds of warnings when opening the settings view, a warning that so far I'm not seeing anywhere else.

```
Choreographer  I  Skipped 38 frames!  The application may be doing too much work on its main thread.
```

This PR does a couple of things to address this:

1. The most impactful is to refactor MDMSettings to use StateFlows and only call RestrictionsManager once to initialize those (and do so outside of the UI thread)
2. Simplifies the settings model and takes out some StateFlows and corresponding `collectWithState()` calls

The first part is implemented in reusable fashion as the extension method `ViewModel.launchForInit`

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
+++ b/android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
@@ -5,65 +5,60 @@ package com.tailscale.ipn.mdm
 
 import android.content.RestrictionsManager
 import com.tailscale.ipn.App
+import kotlin.reflect.KVisibility
+import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.isSubclassOf
+import kotlin.reflect.jvm.jvmErasure
 
-class MDMSettings(private val restrictionsManager: RestrictionsManager? = null) {
-  fun get(setting: BooleanSetting): Boolean {
-    restrictionsManager?.let {
-      if (it.applicationRestrictions.containsKey(setting.key)) {
-        return it.applicationRestrictions.getBoolean(setting.key)
-      }
-    }
-    return App.getApplication().getEncryptedPrefs().getBoolean(setting.key, false)
+object MDMSettings {
+  val forceEnabled = BooleanMDMSetting("ForceEnabled", "Force Enabled Connection Toggle")
+
+  val exitNodeID = StringMDMSetting("ExitNodeID", "Forced Exit Node: Stable ID")
+  val keyExpirationNotice = StringMDMSetting("KeyExpirationNotice", "Key Expiration Notice Period")
+  val loginURL = StringMDMSetting("LoginURL", "Custom control server URL")
+  val managedByCaption = StringMDMSetting("ManagedByCaption", "Managed By - Caption")
+  val managedByOrganizationName =
+      StringMDMSetting("ManagedByOrganizationName", "Managed By - Organization Name")
+  val managedByURL = StringMDMSetting("ManagedByURL", "Managed By - Support URL")
+  val tailnet = StringMDMSetting("Tailnet", "Recommended/Required Tailnet Name")
+
+  val hiddenNetworkDevices =
+      StringArrayListMDMSetting("HiddenNetworkDevices", "Hidden Network Device Categories")
+
+  val allowIncomingConnections =
+      AlwaysNeverUserDecidesMDMSetting("AllowIncomingConnections", "Allow Incoming Connections")
+  val detectThirdPartyAppConflicts =
+      AlwaysNeverUserDecidesMDMSetting(
+          "DetectThirdPartyAppConflicts", "Detect potentially problematic third-party apps")
+  val exitNodeAllowLANAccess =
+      AlwaysNeverUserDecidesMDMSetting(
+          "ExitNodeAllowLANAccess", "Allow LAN Access when using an exit node")
+  val postureChecking =
+      AlwaysNeverUserDecidesMDMSetting("PostureChecking", "Enable Posture Checking")
+  val useTailscaleDNSSettings =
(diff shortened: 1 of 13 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/SettingsView.kt
```diff
@@ -45,81 +44,104 @@ fun SettingsView(
...
           if (isAdmin) {
-            item {
-              Spacer(modifier = Modifier.height(4.dp))
-              AdminTextView { handler.openUri(Links.ADMIN_URL) }
-            }
+            Spacer(modifier = Modifier.height(4.dp))
+            AdminTextView { handler.openUri(Links.ADMIN_URL) }
           }
 
-          settingsBundles.forEach { bundle ->
```
oxtoacart, line 68: Making this a bit more literal reduces the number of StateFlows that we have to create and is IMHO more readable.

android/src/main/java/com/tailscale/ipn/ui/view/SettingsView.kt
```diff
@@ -45,81 +44,104 @@ fun SettingsView(
...
+  ListItem(
+      modifier = Modifier.clickable { if (enabled) setting.onClick() },
+      headlineContent = { Text(setting.title ?: stringResource(setting.titleRes)) },
+      trailingContent = {
+        TintedSwitch(checked = swVal, onCheckedChange = setting.onToggle, enabled = enabled)
+      })
+}
+
+@Composable
+private fun NavRow(setting: Setting) {
```
oxtoacart, line 132: Each setting type has its own function now so that we only collectAsState what we need for that type. That's particularly useful for NavRow since it has no dynamic state.

android/src/main/java/com/tailscale/ipn/ui/viewModel/SettingsViewModel.kt
```diff
@@ -3,35 +3,21 @@
...
 import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.launch
 
 enum class SettingType {
   NAV,
   SWITCH,
   TEXT
 }
 
-class ComposableStringFormatter(@StringRes val stringRes: Int, vararg val params: Any) {
```
oxtoacart, line 28: This level of indirection seemed unnecessary, as we don't actually use the params anywhere. It's unlikely that this had a performance effect, but one less thing to worry about.

android/src/main/java/com/tailscale/ipn/ui/viewModel/SettingsViewModel.kt
```diff
@@ -94,66 +63,67 @@ class SettingsViewModel(val navigation: SettingsNav) : IpnViewModel() {
   // Display name for the logged in user
   var isAdmin: StateFlow<Boolean> = MutableStateFlow(false)
 
-  val settings: StateFlow<List<SettingBundle>> = MutableStateFlow(emptyList())
+  val dns =
```
oxtoacart, line 66: The model now only has one instance of each Setting.

android/src/main/java/com/tailscale/ipn/ui/view/SettingsView.kt
barnstar: DNS and tailnetLock will end up getting wrapped in let? block too (they can be disabled via MDM), so it's a state flow per setting, instead of a stateflow for all of the settings... But this does look cleaner.

barnstar (approved):

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -147,7 +155,7 @@ open class IpnViewModel : ViewModel() {
 
   fun deleteProfile(profile: IpnLocal.LoginProfile, completionHandler: (Result<String>) -> Unit) {
     Client(viewModelScope).deleteProfile(profile) {
-      viewModelScope.launch { loadUserProfiles() }
+      launchForInit { loadUserProfiles() }
```
kari-ts, line 158: would we want to use launchForInit in addProfile and switchProfile too?

oxtoacart: I've since removed this, turns out there was a better solution.

android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
```diff
@@ -5,65 +5,61 @@ package com.tailscale.ipn.mdm
...
-class MDMSettings(private val restrictionsManager: RestrictionsManager? = null) {
-  fun get(setting: BooleanSetting): Boolean {
-    restrictionsManager?.let {
-      if (it.applicationRestrictions.containsKey(setting.key)) {
-        return it.applicationRestrictions.getBoolean(setting.key)
-      }
-    }
-    return App.getApplication().getEncryptedPrefs().getBoolean(setting.key, false)
-  }
+object MDMSettings {
```
oxtoacart, line 13: This basically mimics the structure of Notifier now.

android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
```diff
@@ -5,65 +5,61 @@ package com.tailscale.ipn.mdm
...
-            ?: "show"
-    return when (storedString) {
-      "hide" -> {
-        ShowHideValue.Hide
-      }
-      else -> {
-        ShowHideValue.Show
-      }
-    }
+  val allSettings by lazy {
```
oxtoacart, line 51: A little reflection magic.

agottardo (approved):

PR #266, opened 2024-03-28
android/ui: add support for logging in using a custom control server

barnstar:
fixes ENG-2871

Adds a menu option under the FUS for logging in with your custom control server.  Follows the general pattern used on macOS.

### DIFF
--- a/android/src/main/res/values/strings.xml
+++ b/android/src/main/res/values/strings.xml
@@ -84,10 +84,16 @@
     <string name="logout_failed">Unable to logout at this time.  Please try again.</string>
     <string name="error">Error</string>
     <string name="accounts">Accounts</string>
-    <string name="add_account">Add Another Account…</string>
+    <string name="add_account">Add Another Account</string>
+    <string name="add_account_short">Add Account</string>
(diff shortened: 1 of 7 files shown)


android/src/main/res/values/strings.xml
```diff
@@ -84,10 +84,16 @@
     <string name="logout_failed">Unable to logout at this time.  Please try again.</string>
     <string name="error">Error</string>
     <string name="accounts">Accounts</string>
-    <string name="add_account">Add Another Account…</string>
+    <string name="add_account">Add Another Account</string>
```
agottardo, line 87:
```suggestion
    <string name="add_account">Add another account</string>
```

android/src/main/res/values/strings.xml
```diff
@@ -84,10 +84,16 @@
...
+    <string name="add_account_short">Add Account</string>
```
agottardo, line 88:
```suggestion
    <string name="add_account_short">Add account</string>
```

agottardo (approved):

PR #267, opened 2024-03-28
ui: remove clear button if search bar is empty

kari-ts:
Updates tailscale/corp#18202


barnstar (commented):
@Kari I'm going to remove this search bar entirely and replace it with a TextField to get the look that @sonovawolf wants.   This isn't the intended use for SearchBar anyway.

PR #270, opened 2024-03-29
.github/workflows: remove go-licenses action

willnorris:
This is now handled by an action running in corp.

Updates tailscale/corp#18803


creachadair (approved):
👋🏻

PR #272, opened 2024-04-01
android/ui: restyle the search bar

barnstar:
Updates tailscale/corp#18202

restyles the search bar to look a little more iOS like.  switch to a normal text field which gives us much more flexibility over the look and all of the flexbiility of SearchBar.


oxtoacart (commented):
Functionally, this generally works.

W.R.T. [the requirements](https://www.notion.so/tailscale/iOS-Android-differences-f45b6c550e0d4aa197d41ba33e5df4b2?p=b60dd106f0c24d2aa5a3946117b7b1d7&pm=s), there are a couple of issues.

1. The search field should be on the white background rather than gray (i.e. grouped with the list below)

2. The X should only show up if there is text in the search field.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -318,75 +325,94 @@ fun PeerList(
...
-                  })
-            }
+  Box(modifier = Modifier.fillMaxWidth().background(color = MaterialTheme.colorScheme.surface)) {
+    OutlinedTextField(
+        modifier =
+            Modifier.fillMaxWidth().padding(horizontal = 16.dp, vertical = 8.dp).onFocusChanged {
+              isFocussed = it.isFocused
+            },
+        singleLine = true,
+        shape = MaterialTheme.shapes.large,
```
barnstar, line 338: This should match the requirements... Clear immediately clears and defocusses.   Since we track the focus, we can make this a two-step thing, but that's probably overkill.

oxtoacart (approved):
Looking better! I would add some more whitespace above the search box (try 24.dp instead of the 8.dp it is now). Other than that, LGTM.

PR #273, opened 2024-04-01
MainActivity: bring activity to focus on login

kari-ts:
Use CustomTabsIntent to launch login, but fall back to regular intent if Chrome is not installed When logging in, on change in state to login complete, bring activity back to focus

Updates tailscale/corp#18202
Fixes ENG-2928

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/MainActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/MainActivity.kt
@@ -16,9 +16,11 @@ import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.result.ActivityResultLauncher
 import androidx.activity.result.contract.ActivityResultContract
+import androidx.browser.customtabs.CustomTabsIntent
 import androidx.compose.animation.core.tween
 import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
+import androidx.core.net.toUri
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavType
 import androidx.navigation.compose.NavHost
@@ -28,6 +30,7 @@ import androidx.navigation.navArgument
 import androidx.navigation.navigation
 import com.tailscale.ipn.Peer.RequestCodes
 import com.tailscale.ipn.mdm.MDMSettings
+import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.notifier.Notifier
 import com.tailscale.ipn.ui.theme.AppTheme
 import com.tailscale.ipn.ui.view.AboutView
@@ -52,6 +55,7 @@ import com.tailscale.ipn.ui.viewModel.ExitNodePickerNav
 import com.tailscale.ipn.ui.viewModel.SettingsNav
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 
 class MainActivity : ComponentActivity() {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -182,12 +186,38 @@ class MainActivity : ComponentActivity() {
...
-  private fun login(url: String) {
-    // (jonathan) TODO: This is functional, but the navigation doesn't quite work
-    // as expected.  There's probably a better built in way to do this.  This will
-    // unblock in dev for the time being though.
-    val browserIntent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
-    startActivity(browserIntent)
+  private fun login(urlString: String) {
+    lifecycleScope.launch {
+      try {
+        Notifier.state.collect { state ->
```
oxtoacart, line 192: This block deserves a comment as to why it's here (i.e. we're waiting for login to complete and then relaunching the MainActivity.

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -182,12 +186,38 @@ class MainActivity : ComponentActivity() {
...
   }
 
-  private fun login(url: String) {
-    // (jonathan) TODO: This is functional, but the navigation doesn't quite work
-    // as expected.  There's probably a better built in way to do this.  This will
-    // unblock in dev for the time being though.
-    val browserIntent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
-    startActivity(browserIntent)
+  private fun login(urlString: String) {
+    lifecycleScope.launch {
```
oxtoacart, line 190: It's probably better to use our application scope here in case the activity gets destroyed but the app doesn't.

oxtoacart: App.getApplication().applicationScope.launch

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -182,12 +186,38 @@ class MainActivity : ComponentActivity() {
...
+      try {
+        Notifier.state.collect { state ->
+          if (state > Ipn.State.NeedsMachineAuth) {
+            val intent =
+                Intent(applicationContext, MainActivity::class.java).apply {
+                  addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TOP)
+                  action = Intent.ACTION_MAIN
+                  addCategory(Intent.CATEGORY_LAUNCHER)
+                }
+            startActivity(intent)
```
oxtoacart, line 200: I'm pretty sure we need to stop the coroutine here, otherwise this will keep running even after we've launched the activity again? Or maybe the use of `lifecycleScope` above eliminated that need?

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -182,12 +186,38 @@ class MainActivity : ComponentActivity() {
...
+          }
+        }
+      } catch (e: Exception) {
+        Log.e(TAG, "Login: failed to start MainActivity: $e")
+      }
+    }
+
+    val url = urlString.toUri()
+    try {
+      val customTabsIntent = CustomTabsIntent.Builder().build()
```
oxtoacart, line 210: Cool, I didn't know this was a thing. Seems like a nice solution to not open a bunch of tabs in the user's browser.

android/src/main/java/com/tailscale/ipn/MainActivity.kt
oxtoacart: I tested this and this coroutine does indeed keep running.

android/src/main/java/com/tailscale/ipn/MainActivity.kt
kari-ts: good point - the activity gets destroyed and recreated if the screen is rotated, and we want this to still work

android/src/main/java/com/tailscale/ipn/MainActivity.kt
kari-ts: aha. good catch, I thought lifecycleScope took care of it too

oxtoacart (approved):
LGTM!

PR #274, opened 2024-04-02
android/ui: distinguish Tailnet domains in user view

oxtoacart:
Updates tailscale/corp#18202

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
@@ -57,7 +57,6 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
-import androidx.compose.ui.unit.sp
 import androidx.lifecycle.viewmodel.compose.viewModel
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/UserView.kt
```diff
@@ -40,10 +41,12 @@ fun UserView(
           headlineContent = {
             Text(
                 text = profile.UserProfile.DisplayName,
-                style = MaterialTheme.typography.titleMedium)
+                style = MaterialTheme.typography.titleMedium.short)
```
oxtoacart, line 44: Drive-by formatting fix

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -218,7 +207,7 @@ fun ExitNodeStatus(navAction: () -> Unit, viewModel: MainViewModel) {
 }
 
 @Composable
-fun SettingsButton(user: IpnLocal.LoginProfile?, action: () -> Unit) {
```
oxtoacart, line 221: Drive-by, unused parameter

barnstar (approved):
Functionally, this is fine... But it's different.  It does feel more correct.  Design should really weigh in here.

oxtoacart:
Thanks @barnstar , I'll merge and show to @sonovawolf in our walkthrough tomorrow.

PR #276, opened 2024-04-02
android/ui: add key expiry banner

barnstar:
fixes ENG-2912

copies and adds the a key expiry banner identical to the one on iOS.

fixes a couple of small layout issues with the search bar

Screenshot 2024-04-02 at 1 32 24 PM

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
@@ -5,6 +5,7 @@ package com.tailscale.ipn.ui.model
 
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
 import com.tailscale.ipn.R
 import com.tailscale.ipn.ui.theme.off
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
```diff
@@ -97,7 +99,7 @@ class Tailcfg {
       get() = Name.endsWith(".mullvad.ts.net.")
 
     val displayName: String
-      get() = ComputedName ?: ""
+      get() = ComputedName
```
kari-ts, line 102: this field might be omitted sometimes right? https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L357

barnstar: Good catch.  That was an auto correction.  It's not optional in the schema and should be.  I will audit the other fields too.

barnstar: Fixed.  Switched to node.displayName throughout, which uses ComputedName ?: Name...

kari-ts (approved):

PR #279, opened 2024-04-02
IPNService: on close, edit prefs with WantRunning=false

kari-ts:
This fixes the issue where when the VPN was turned off by system settings, the toggle was showing the user as conencted

Updates tailscale/corp#18202
Fixes tailscale/corp#18863


libtailscale/backend.go
```diff
@@ -212,7 +212,7 @@ func (a *App) runBackend(ctx context.Context) error {
 			if cfg.rcfg != nil && state >= ipn.Starting {
```
barnstar, line 212: Is it idiomatic go practice to algebraically compare 'enums' like this instead of:

switch val {
   case a,b,c,e,f: doThing()
   case d: doOtherThing()
}

//add g, compiler failure

This seems very dangerous...

oxtoacart (approved):
Working for me!

barnstar (approved):

PR #280, opened 2024-04-03
android/ui: show the switching-to user immediately on FUS switch

barnstar:
Updates tailscale/corp#18202

This moves the logged-in-user to global state (which is is) and sets it immediately when you switch tailnets.  On failure, we will set things back to the prior user.

This fixes the issue of seeing the "old" user while the transition is happening.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
@@ -38,11 +38,8 @@ open class IpnViewModel : ViewModel() {
   init {
     viewModelScope.launch {
       Notifier.state.collect {
-        // Refresh the user profiles if we're transitioning out of the
-        // NeedsLogin state.
-        if (it == Ipn.State.NeedsLogin) {
-          viewModelScope.launch { loadUserProfiles() }
-        }
+        // Reload the user profiles on all state transitions to ensure loggedInUser is correct
+        viewModelScope.launch { loadUserProfiles() }
       }
     }
 


android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -124,9 +127,15 @@ open class IpnViewModel : ViewModel() {
   }
 
   fun switchProfile(profile: IpnLocal.LoginProfile, completionHandler: (Result<String>) -> Unit) {
-    Client(viewModelScope).switchProfile(profile) {
-      startVPN()
-      completionHandler(it)
+    val currentUser = loggedInUser.value
+    loggedInUser.set(profile)
```
oxtoacart, line 131: Notifier is really meant to be a channel to get updates from the backend's state. I don't like side-effecting it based on in-progress operations in a view model. Is there a different approach we can take?

oxtoacart: Here's a clue ...

If you comment out the condition on line 43 of IpnViewModel, then there's no delay in picking up the correct user.

```
viewModelScope.launch {
      Notifier.state.collect {
        // Refresh the user profiles if we're transitioning out of the
        // NeedsLogin state.
        //        if (it == Ipn.State.NeedsLogin) {
        viewModelScope.launch { loadUserProfiles() }
        //        }
      }
    }
 ```

oxtoacart: Honestly, refreshing the user profile on every state change doesn't seem too bad, so maybe we can just take out the `if (it == Ipn.State.NeedsLogin)` condition?

barnstar: Ack... Done.  That's enough to get it to work.  This should really come in a notifier blob instead of requiring the UI to track it via localAPI calls... but localAPI is cheap enough.

oxtoacart (approved):

PR #281, opened 2024-04-03
android/ui: change exit node disable string to 'stop'

barnstar:
Updates tailscale/corp#18202

Updates the exit node stop button to 'stop' instead of 'disable' pending the required back end changes to remember the preferred exit node.


android/src/main/res/values/strings.xml
```diff
@@ -125,6 +125,7 @@
     <string name="enabled">Enabled</string>
     <string name="disabled">Disabled</string>
     <string name="disable">Disable</string>
```
oxtoacart, line 127: Should we get rid of this as it's now unused?

oxtoacart (approved):

PR #283, opened 2024-04-03
android/ui: add compose-style splash screen

barnstar:
Updates tailscale/corp#18202

Updates the splash screen to the modern themed jetpack compose variant.


oxtoacart (approved):
Code LGTM and it works. I still don't understand what it's supposed to look like, but with that caveat, I approve!

PR #292, opened 2024-04-04
android/ui: cleanup unused code

barnstar:
Updates tailscale/corp#18202

Removes unused code from various legacy ported classes.  Fixes a number of soft warnings.

Removes/renames a few composables we were no longer using.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -1,39 +1,25 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
+
 package com.tailscale.ipn
 
-import android.Manifest
 import android.app.Activity
 import android.app.Application
-import android.app.DownloadManager
-import android.app.Fragment
-import android.app.FragmentTransaction
 import android.app.NotificationChannel
-import android.app.PendingIntent
-import android.app.UiModeManager
-import android.content.ContentResolver
-import android.content.ContentValues
 import android.content.Context
 import android.content.Intent
 import android.content.SharedPreferences
-import android.content.pm.PackageInfo
-import android.content.pm.PackageManager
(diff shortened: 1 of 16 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -62,22 +48,17 @@ class App : Application(), libtailscale.AppContext {
...
+
     private val networkConnectivityRequest =
         NetworkRequest.Builder()
             .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
             .addCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN)
             .build()
-    lateinit var appInstance: App
 
-    @JvmStatic
-    fun startActivityForResult(act: Activity, intent: Intent?, request: Int) {
```
kari-ts, line 77: I think we still need this. We could definitely refactor this, but right now this launches Peer with the request VPN intent

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -265,18 +243,6 @@ class App : Application(), libtailscale.AppContext {
     return null
   }
 
-  // attachPeer adds a Peer fragment for tracking the Activity
-  // lifecycle.
-  fun attachPeer(act: Activity) {
```
kari-ts, line 270: aha. I see, requestVpnPermission calls VpnService.prepare, so when App.prepareVpn calls VpnService.prepare, this is almost always null (vpn has been prepared), so we never need to call startActivityForResult right now. but if we're planning on moving requestVpnPermission() out of onStart, then we would still need this (and to actually call this function too)

kari-ts: we should probably refactor the vpn logic in general, it's overly confusing as is

barnstar:
This has gone completely stale... I will redo.

PR #294, opened 2024-04-05
Stop treating settings as data

oxtoacart:
The first commit fixes the proximate [issue](https://github.com/tailscale/corp/issues/18983). The second commit just generally cleans up how we do settings. Having them as data unnecessarily complicated things and kept us from doing more normal state management.


barnstar (approved):

android/src/main/java/com/tailscale/ipn/ui/view/ExitNodePicker.kt
```diff
@@ -71,7 +73,11 @@ fun ExitNodePicker(
         // TODO: make sure this actually works, and if not, leave it out for now
```
kari-ts, line 73: I think we can take this comment out :)

PR #297, opened 2024-04-05
build.gradle, Makefile: remove custom fdroid build

kari-ts:
We're no longer using GoogleSignIn, so there's no need for separate product flavors Updates tailscale/corp#18202

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,13 +171,9 @@ $(LIBTAILSCALE): Makefile android/libs $(LIBTAILSCALE_SOURCES) $(GOBIN)/gomobile
 
 libtailscale: $(LIBTAILSCALE)
 
-tailscale-new-fdroid.apk: $(LIBTAILSCALE)
-	(cd android && ./gradlew test assembleFdroidDebug)
-	mv android/build/outputs/apk/fdroid/debug/android-fdroid-debug.apk $@
-
 tailscale-new-debug.apk: $(LIBTAILSCALE)
-	(cd android && ./gradlew test assemblePlayDebug)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -171,10 +171,6 @@ $(LIBTAILSCALE): Makefile android/libs $(LIBTAILSCALE_SOURCES) $(GOBIN)/gomobile
 
 libtailscale: $(LIBTAILSCALE)
 
-tailscale-new-fdroid.apk: $(LIBTAILSCALE)
-	(cd android && ./gradlew test assembleFdroidDebug)
-	mv android/build/outputs/apk/fdroid/debug/android-fdroid-debug.apk $@
-
 tailscale-new-debug.apk: $(LIBTAILSCALE)
 	(cd android && ./gradlew test assemblePlayDebug)
```
bradfitz, line 175: but this says "PlayDebug" .... is it using any Play Store / Google APIs?

kari-ts: nope, thanks for pointing this out!

linsui:
The navigation-dynamic-features-fragment lib depends on gms. Maybe I should simply remove it in the fdroid build metadata?

bradfitz:
@kari-ts, have you verified the build works on a non-Google-ified device?

oxtoacart:
> navigation-dynamic-features-fragment

AFAICT, we're not actually using that library for anything, so we can just remove that dependency across the board.

oxtoacart (approved):

kari-ts:
removed some additional unused dependencies.

verified the build mostly works on an AOSP emulator

PR #298, opened 2024-04-08
android/ui: navigation improvements

oxtoacart:
1. More careful back navigation to avoid navigating to blank screen
2. After adding an account, navigate back to main view

Updates tailscale/corp#18202

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
@@ -30,8 +30,9 @@ import com.tailscale.ipn.R
 import com.tailscale.ipn.ui.Links
 
 @Composable
-fun AboutView(nav: BackNavigation) {
-  Scaffold(topBar = { Header(R.string.about_view_title, onBack = nav.onBack) }) { innerPadding ->
+fun AboutView(backToSettings: BackNavigation) {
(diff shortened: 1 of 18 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
```diff
@@ -30,8 +30,9 @@ import com.tailscale.ipn.R
 import com.tailscale.ipn.ui.Links
 
 @Composable
-fun AboutView(nav: BackNavigation) {
-  Scaffold(topBar = { Header(R.string.about_view_title, onBack = nav.onBack) }) { innerPadding ->
+fun AboutView(backToSettings: BackNavigation) {
```
barnstar, line 33: nit:  There's an assumption in the naming here.  The caller sets up the back nav stack.  It does not have to be "Settings".

barnstar (approved):
Minor nit about the naming of the navigation callbacks in the constructors (they're only accurate with the current way the view stacks are set up) but otherwise, looks good.

android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
oxtoacart: This was intentional - the caller is actually explicitly making this navigate back to settings, not just generally "back".

PR #304, opened 2024-04-10
ui: port syspolicy handler code to new app

kari-ts:
port over https://github.com/tailscale/tailscale-android/pull/199 from cmd/tailscale and legacy_android to libtailscale and android/

fix issue where we were creating MDMSettings every time we were retrieving policy settings 

Updates tailscale/corp#18202

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/mdm/MDMSettingsDefinitions.kt
+++ b/android/src/main/java/com/tailscale/ipn/mdm/MDMSettingsDefinitions.kt
@@ -4,31 +4,13 @@
 package com.tailscale.ipn.mdm
 
 import android.os.Bundle
-import android.util.Log
 import com.tailscale.ipn.App
 import com.tailscale.ipn.ui.util.set
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 
-// Don't keep track of the key if shouldRegisterKey is false - shouldRegisterKey should be false when creating an MDMSetting to check whether a key has been registered with that setting.
-abstract class MDMSetting<T>(defaultValue: T, val key: String, val localizedTitle: String, private val shouldRegisterKey: Boolean = true) {
-  private val keys = mutableSetOf<String>()
-
-    init {
-      if (shouldRegisterKey) { registerKey(key)
-    }
-  }
+abstract class MDMSetting<T>(defaultValue: T, val key: String, val localizedTitle: String) {
   val flow: StateFlow<T> = MutableStateFlow<T>(defaultValue)
-
-  private fun registerKey(key: String) {
-    if (!keyExists(key)) {
-        keys.add(key)
-    }
-}
-
-fun keyExists(key: String): Boolean {
-    return keys.contains(key)
-}
 
   fun setFrom(bundle: Bundle?, app: App) {
     val v = getFrom(bundle, app)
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/tailscale/ipn/mdm/MDMSettingsDefinitions.kt
```diff
@@ -4,14 +4,32 @@
...
+  }
   val flow: StateFlow<T> = MutableStateFlow<T>(defaultValue)
 
+  private fun registerKey(key: String) {
+    if (!keyExists(key)) {
+        keys.add(key)
+    }
+}
+
+fun keyExists(key: String): Boolean {
```
barnstar, line 29: nit: the one line method here isn't really needed (unless it's called by something else I'm not seeing.)

barnstar (approved):
Looks good, but @agottardo should take a look as well.

android/src/main/java/com/tailscale/ipn/mdm/MDMSettingsDefinitions.kt
```diff
@@ -4,14 +4,32 @@
...
 
 import android.os.Bundle
+import android.util.Log
 import com.tailscale.ipn.App
 import com.tailscale.ipn.ui.util.set
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 
-abstract class MDMSetting<T>(defaultValue: T, val key: String, val localizedTitle: String) {
+// Don't keep track of the key if shouldRegisterKey is false - shouldRegisterKey should be false when creating an MDMSetting for retrieving a setting.
```
agottardo, line 13: Does this mean that all initializer calls in MDMSettings.kt need to be updated with `shouldRegisterKey: false`?. The default value here is `true`.

android/src/main/java/com/tailscale/ipn/mdm/MDMSettingsDefinitions.kt
kari-ts: we are using this in App.getSyspolicyStringValue - MDMSettings has all the MDMSettings we've created, and we want to check to see if there is one with the key we're given

android/src/main/java/com/tailscale/ipn/mdm/MDMSettingsDefinitions.kt
kari-ts: no, they should register keys. when we are checking to see if there exists an MDMSetting for a certain key, shouldRegisterKey should be false. I updated the comment to hopefully be a little clearer

libtailscale/syspolicy_handler.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"log"
+
+	"tailscale.com/util/syspolicy"
+)
+
+// androidHandler is a syspolicy handler for the Android version of the Tailscale client,
+// which lets the main networking code read values set via the Android RestrictionsManager.
+type androidHandler struct {
```
oxtoacart, line 14: nit: since this isn't in its own package, a more specific name would be good, like `syspolicyHandler`.

agottardo (approved):

oxtoacart (commented):
[This](https://github.com/tailscale/tailscale-android/pull/308) is a simpler way to do it.

oxtoacart (approved):

PR #305, opened 2024-04-10
android/ui: add support for remembering the last used exit node

barnstar:
Updates ENG-2911

Disabling an exit node is now temporary and you can re-enable it without re-selecting it from the picker.

Selecting "none" will clear the current selection as well as the previously remembered node.


android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -60,17 +60,30 @@ class MainViewModel : IpnViewModel() {
...
-    Client(viewModelScope).editPrefs(prefsOut) { LoadingIndicator.stop() }
+    if (prefs.activeExitNodeID != null) {
+      // We have an active exit node so we should keep it, but disable it
+      Client(viewModelScope).setUseExitNode(false) { LoadingIndicator.stop() }
+    } else if (prefs.selectedExitNodeID != null) {
+      // We have a prior exit node to enable
+      Client(viewModelScope).setUseExitNode(true) { LoadingIndicator.stop() }
+    } else {
+      // This should not be possible.  In this state the button is hidden
+      Log.e(TAG, "No exit node to disable an no prior exit node to enable")
```
agottardo, line 85:
```suggestion
      Log.e(TAG, "No exit node to disable and no prior exit node to enable")
```

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
agottardo, line 1: We're starting to add a lot of exit node-specific logic in here, perhaps we can move it out to a separate component one day?

android/src/main/java/com/tailscale/ipn/ui/theme/Theme.kt
```diff
@@ -333,6 +333,17 @@ val ColorScheme.onBackgroundLogoDotDisabled: Color
         Color(0x66FFFFFF)
       }
 
+val ColorScheme.button: ButtonColors
```
agottardo, line 336: Should we call this `exitNodeBannerButton`?

agottardo (approved):
Some nits

kari-ts (approved):

PR #301, opened 2024-04-09
android/ui: show count of Mullvad countries, also fix navigation-rela…

oxtoacart:
…ted crash when adding new account

Updates tailscale/corp#18202

PR #306, opened 2024-04-10
android/ui: prevent navigation before we've added content

barnstar:
Fixes tailscale/corp#19070

If the activity hadn't yet been created, we can still get an onIntent which was assuming the navController had been instantiated.  Switched that to and optional so that we can null check it.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/MainActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/MainActivity.kt
@@ -74,11 +74,10 @@ import kotlinx.coroutines.launch
 
 class MainActivity : ComponentActivity() {
   private lateinit var requestVpnPermission: ActivityResultLauncher<Unit>
-  private lateinit var navController: NavHostController
+  private var navController: NavHostController? = null
 
   companion object {
     private const val TAG = "Main Activity"
-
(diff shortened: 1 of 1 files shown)


agottardo (approved):

kari-ts:
if an activity doesn't exist, doesn't it get created before an intent gets sent to it? I wonder if the navigation issue was caused by the bug fixed here: https://github.com/tailscale/tailscale-android/pull/301/commits/45cb242e02f96e85dbe92d05aeb6c03543ab9022 (line 109 in MainActivity)

barnstar:
> if an activity doesn't exist, doesn't it get created before an intent gets sent to it? I wonder if the navigation issue was caused by the bug fixed here: [45cb242](https://github.com/tailscale/tailscale-android/commit/45cb242e02f96e85dbe92d05aeb6c03543ab9022) (line 109 in MainActivity)

It was caused by that yeah.  I was only able to repro this on my Chromebook, but it definitely blew up in onNewIntent when we tried to navigate back from the login flow.  I'm assuming if there's no content, sorting out the navigation doesn't matter.

oxtoacart:
I wish we'd talked, that was fixed [here](https://github.com/tailscale/tailscale-android/pull/301). I'll rebase and refix.

PR #308, opened 2024-04-10
Ox/syspol

oxtoacart:


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -459,29 +455,15 @@ class App : Application(), libtailscale.AppContext {
...
-      showHideSetting.keyExists(key) ->
-          showHideSetting.getFrom(manager.applicationRestrictions, this).value
-      stringSetting.keyExists(key) ->
-          stringSetting.getFrom(manager.applicationRestrictions, this) ?: ""
-      else -> {
-        Log.d("MDM", "$key is not defined on Android. Returning empty.")
-        ""
-      }
-    }
+    return MDMSettings.allSettingsByKey[key]?.flow?.value?.toString()
```
oxtoacart, line 463: Since we already have the latest values for these settings in stateflows, we can just get them from there (which is also more performant).

kari-ts (approved):

PR #309, opened 2024-04-10
ui: show 'No results' when search returns empty

kari-ts:
Updates tailscale/corp#18202
Fixes tailscale/corp#19062
Screenshot (Apr 10, 2024 4_35_37 PM)

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
@@ -40,6 +40,7 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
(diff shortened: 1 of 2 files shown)


agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -321,6 +322,7 @@ fun PeerList(
 ) {
   val peerList = viewModel.peers.collectAsState(initial = emptyList<PeerSet>())
   val searchTermStr by viewModel.searchTerm.collectAsState(initial = "")
+  val showNoResults = derivedStateOf { searchTermStr.isNotEmpty() && peerList.value.isEmpty() }
```
barnstar, line 325: nit: we're usually calling .value when collect/derive the state.  Then you see a bool elsewhere, and not a State<Boolean>..  It's only used once here, so no big deal, but change it if you have sec.

```
  val showNoResults = derivedStateOf { searchTermStr.isNotEmpty() && peerList.value.isEmpty() }.value
```

PR #311, opened 2024-04-11
android/ui: add state subheadings to settings rows

barnstar:
Fixes tailscale/corp#19044

Add version, dns state, and tailnet lock status as settings option subtitles.

Screenshot 2024-04-11 at 12 38 00 PM


agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/view/SettingsView.kt
```diff
@@ -110,7 +127,14 @@ object Setting {
               style = MaterialTheme.typography.bodyMedium,
               color = if (destructive) MaterialTheme.colorScheme.error else Color.Unspecified)
         },
-    )
+        supportingContent = {
```
oxtoacart, line 130: Please change this to the below to avoid having empty supporting content.

```
supportingContent = subtitle?.let {
            Text(
                it,
                style = MaterialTheme.typography.bodySmall,
                color = MaterialTheme.colorScheme.onSurfaceVariant)
        })
```

barnstar: That needs to be a closure, but yes, done!

PR #312, opened 2024-04-11
android/ui: sort self peer first in list of peers

oxtoacart:
Closes tailscale/corp#19111

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/util/PeerHelper.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/util/PeerHelper.kt
@@ -39,7 +39,18 @@ class PeerCategorizer {
         grouped
             .map { (userId, peers) ->
               val profile = netmap.userProfile(userId)
-              PeerSet(profile, peers.sortedBy { it.ComputedName })
+              PeerSet(
+                  profile,
+                  peers.sortedWith { a, b ->
+                    when {
+                      a.StableID == b.StableID -> 0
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/ui/util/PeerHelper.kt
```diff
@@ -39,7 +39,18 @@ class PeerCategorizer {
...
-              PeerSet(profile, peers.sortedBy { it.ComputedName })
+              PeerSet(
+                  profile,
+                  peers.sortedWith { a, b ->
+                    when {
+                      a.StableID == b.StableID -> 0
+                      a.isSelfNode(netmap) -> -1
+                      b.isSelfNode(netmap) -> 1
+                      else ->
+                          (a.ComputedName?.lowercase() ?: "").compareTo(
```
oxtoacart, line 50: I went ahead and made this case insensitive to match how we sort users.

barnstar (approved):
This is a strange UX request, because SelfPeer is probably the one I care *least* about.   I know it's online.  I don't need to know it's IP.

oxtoacart:
> This is a strange UX request, because SelfPeer is probably the one I care _least_ about. I know it's online. I don't need to know it's IP.

Yeah, if we weren't already doing it like this on iOS I'd leave it alone. Really, self peer probably shouldn't even be in the list and be accessed differentlly.

PR #313, opened 2024-04-11
android/ui: improve dev QOL by adding support for compose previews

barnstar:
Updates corp#19117

This adds @Previews for many of the primary views.  We can expand upon this over time to include different data sets, states, etc.  Some are not included here due to issues with dependency injection/singletons, lack of fake data etc... But this gets the ball rolling.

The ErrorDialogs were not wrapped in the AppTheme either - so one bug fixed.


agottardo (approved):
Looks like fun! Reusable components <3

oxtoacart (approved):
This is super cool!

PR #315, opened 2024-04-11
ui: reintroduce dark mode theme

agottardo:


android/src/main/java/com/tailscale/ipn/ui/view/TailscaleLogoView.kt
```diff
@@ -31,10 +35,24 @@ val logoDotsMatrix: DotsMatrix =
     )
 
 @Composable
-fun TailscaleLogoView(animated: Boolean = false, modifier: Modifier) {
+fun TailscaleLogoView(
+    animated: Boolean = false,
+    usesOnBackgroundColors: Boolean = false,
```
oxtoacart, line 40: Why make this a parameter instead of just checking `isSystemInDarkTheme()`?

agottardo: The colours we use to draw each dot in the logo are different depending on whether you're drawing the logo on top of a generic app surface (such as for a progress indicator, or the welcome view), or over `MaterialTheme.colorScheme.logoBackground`, which you're going to use to draw the app icon (we use that in the about view). See what the iOS implementation does with the `logoStyle` enum parameter.

oxtoacart (approved):
Code LGTM and visually things look in order. I'll defer to @sonovawolf for final design review.

PR #316, opened 2024-04-11
Makefile: clean up legacy builds

kari-ts:
-Remove legacy builds
-Update version number and name using last unstable

Updates tailscale/corp#18202
Fixes tailscale/corp#19001


android/build.gradle
```diff
@@ -37,8 +37,8 @@ android {
     defaultConfig {
         minSdkVersion 26
         targetSdkVersion 34
-        versionCode 198
-        versionName "1.59.53-t0f042b981-g1017015de26"
+        versionCode 203
+        versionName "1.63.93-t1fbaf2610-ge0020f8c4bf"
```
agottardo, line 41: Hmm, this should say 1.65.x now that 1.64 is out

kari-ts: thanks for pointing this out! I updated it with the latest tag. the next unstable that we build for the beta will be 1.65.x

agottardo (approved):
Approved with a word of caution that I'm not too familiar with the Android build infrastructure... 😬

raggi (approved):

oxtoacart (approved):

PR #317, opened 2024-04-12
android/ui: handle NeedsMachineAuth state

barnstar:
Fixes corp#19119

Adds a variation on the ConnectView to render a header and explainer text for the NeedsMachineAuth state.  A button to take you directly to the admin page is presented if you are an admin.

Screenshot 2024-04-12 at 10 50 23 AM

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
@@ -7,6 +7,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.res.stringResource
 import com.tailscale.ipn.R
+import com.tailscale.ipn.ui.Links
 import com.tailscale.ipn.ui.theme.off
 import com.tailscale.ipn.ui.theme.on
 import com.tailscale.ipn.ui.util.ComposableStringFormatter
@@ -90,6 +91,16 @@ class Tailcfg {
       get() =
           Capabilities?.contains("https://tailscale.com/cap/is-admin") == true ||
               CapMap?.contains("https://tailscale.com/cap/is-admin") == true
+
+    // Derives the url to directly administer a node
+    val nodeAdminUrl: String
+      get() = primaryIPv4Address?.let { "${Links.ADMIN_URL}/machines/${it}" } ?: Links.ADMIN_URL
+
+    val primaryIPv4Address: String?
+      get() = displayAddresses.firstOrNull { it.type == DisplayAddress.addrType.V4 }?.address
+
+    val primaryIPv6Address: String?
+      get() = displayAddresses.firstOrNull { it.type == DisplayAddress.addrType.V6 }?.address
 
     // isExitNode reproduces the Go logic in local.go peerStatusFromNode
     val isExitNode: Boolean =
--- a/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
@@ -26,6 +26,7 @@ import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.ArrowDropDown
 import androidx.compose.material.icons.outlined.Clear
 import androidx.compose.material.icons.outlined.Close
+import androidx.compose.material.icons.outlined.Lock
 import androidx.compose.material.icons.outlined.Search
 import androidx.compose.material.icons.outlined.Settings
 import androidx.compose.material3.Button
@@ -93,7 +94,7 @@ data class MainViewNavigation(
 
 @OptIn(ExperimentalPermissionsApi::class)
 @Composable
-fun MainView(navigation: MainViewNavigation, viewModel: MainViewModel = viewModel()) {
+fun MainView(loginAtUrl: (String) -> Unit, navigation: MainViewNavigation, viewModel: MainViewModel = viewModel()) {
   LoadingIndicator.Wrap {
     Scaffold(contentWindowInsets = WindowInsets.Companion.statusBars) { paddingInsets ->
       Column(
@@ -103,7 +104,7 @@ fun MainView(navigation: MainViewNavigation, viewModel: MainViewModel = viewMode
             val user = viewModel.loggedInUser.collectAsState(initial = null).value
             val stateVal = viewModel.stateRes.collectAsState(initial = R.string.placeholder).value
             val stateStr = stringResource(id = stateVal)
-            val netmap = viewModel.netmap.collectAsState(initial = null)
+            val netmap = viewModel.netmap.collectAsState(initial = null).value
 
             ListItem(
                 colors = MaterialTheme.colorScheme.surfaceContainerListItem,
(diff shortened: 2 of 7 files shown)


bradfitz:
Fixes tailscale/corp

(Needs the org too for be recognized)

agottardo (approved):
Note the iOS equivalent also shows a "lock" symbol. Not a big deal for v0 and we can fix it later!

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -53,6 +55,7 @@ class MainViewModel : IpnViewModel() {
         it?.let { netmap ->
           peerCategorizer.regenerateGroupedPeers(netmap)
           peers.set(peerCategorizer.groupedAndFilteredPeers(searchTerm.value))
+          isAdmin.set(netmap.SelfNode.isAdmin)
```
oxtoacart, line 58: This doesn't quite work. For example, on tailscale.com, I don't have `https://tailscale.com/cap/is-admin`, but I am able to approve the device because I'm an [IT Admin](https://tailscale.com/kb/1138/user-roles#it-admin). I should see the link, but I don't.

oxtoacart: AFAICT, we don't have a way of knowing whether or not someone is an IT Admin. Given that, I'd probably err on the side of showing the link.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -252,16 +257,35 @@ fun ConnectView(
...
+          Text(
+              text = stringResource(id = R.string.machine_auth_required),
+              style = MaterialTheme.typography.titleMedium,
+              textAlign = TextAlign.Center)
+          Text(
+              text = stringResource(id = R.string.machine_auth_explainer),
+              style = MaterialTheme.typography.bodyMedium,
+              textAlign = TextAlign.Center)
+           if(isAdmin) {
+            PrimaryActionButton(onClick = {  handler.openUri(Links.ADMIN_URL) }) {
```
oxtoacart, line 282: It would be nice to link directly to the machine in question. It would also be nice to use our login flow so that once the machine is approved, we automatically come back to the app.

oxtoacart (commented):
I feel like we could use a little more whitespace above the "Open admin console" button.

I'm going to submit a PR with suggestions.

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
bradfitz: I guess we never considered that. The cap/is-admin came first and then we added new admin types later:

```
commit 6a87f4dd42bdd547ef6363ac4beb07920adf55e7
Author: Brad Fitzpatrick <...>
Date:   Fri Apr 9 08:25:24 2021 -0700

    control: tell nodes in MapResponse whether they're an admin


commit 15bb5e2adaf7e09af2e838f3fb5e484cb759261c
Author: Sonia Appasamy <...>
Date:   Fri Nov 5 17:16:52 2021 -0400

    control/cfgdb,adminhttp: add new admin console roles
    
    This commit adds three new admin console roles to domains: Network
    Admin, IT Admin, and Auditor. This change also uses these new roles
    to permission what different users can access from the admin api.
    
    Change-Id: Ib4143842699c2d9e74b7f219b8927a8ef30dc6a4
```

BTW, what we'd want to advertise to nodes is "can you authorized a device" and not "are you an IT admin".

The whole "IT Admin" vs "Network Admin" roles I kinda view as a mistake. They conflate titles and capabilities. Maybe they should've been templates people could then rename & tweak permissions on. (Maybe we'll do that later). Related: https://tailscale.com/blog/rbac-like-it-was-meant-to-be

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -324,8 +356,7 @@ fun PeerList(
 ) {
   val peerList = viewModel.peers.collectAsState(initial = emptyList<PeerSet>())
   val searchTermStr by viewModel.searchTerm.collectAsState(initial = "")
-  val showNoResults =
-      derivedStateOf { searchTermStr.isNotEmpty() && peerList.value.isEmpty() }.value
+  val showNoResults = remember { derivedStateOf { searchTermStr.isNotEmpty() && peerList.value.isEmpty() } }.value
```
barnstar, line 359: Drive by warning fix...

android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
```diff
@@ -91,6 +92,14 @@ class Tailcfg {
           Capabilities?.contains("https://tailscale.com/cap/is-admin") == true ||
               CapMap?.contains("https://tailscale.com/cap/is-admin") == true
 
+    // Derives the url to directly administer a node
+    val nodeAdminUrl: String
+      get() {
+        val address =
+                displayAddresses.firstOrNull { it.type == DisplayAddress.addrType.V4 }?.address
+        return if (address != null) "${Links.ADMIN_URL}/machines/${address}" else Links.ADMIN_URL
+      }
```
barnstar, line 101: displayAddresses already removes the mask and categorizes all of our IPs... The adminURL is the first ipV4.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -103,7 +104,7 @@ fun MainView(navigation: MainViewNavigation, viewModel: MainViewModel = viewMode
             val user = viewModel.loggedInUser.collectAsState(initial = null).value
             val stateVal = viewModel.stateRes.collectAsState(initial = R.string.placeholder).value
             val stateStr = stringResource(id = stateVal)
-            val netmap = viewModel.netmap.collectAsState(initial = null)
+            val netmap = viewModel.netmap.collectAsState(initial = null).value
```
barnstar, line 107: Drive by...

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
barnstar: Nice suggestion btw...  @agottardo One more thing to add to the growing "Things Android does better that we need to fix on Apple" backlog...

android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
oxtoacart: Might as well use this same logic for extracting the primary IP for display on the peer list.

oxtoacart (approved):
Just one comment about reusing the new nodeAdminUrl logic on the peer list.

android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
barnstar: Done.  With a fallback to the ipV6 address...

android/src/main/java/com/tailscale/ipn/ui/view/PeerView.kt
```diff
@@ -28,7 +28,7 @@ fun PeerView(
     peer: Tailcfg.Node,
     selfPeer: String? = null,
     stateVal: Ipn.State? = null,
-    subtitle: () -> String = { peer.Addresses?.first()?.split("/")?.first() ?: "" },
+    subtitle: () -> String = { peer.primaryIPv4Address ?: peer.primaryIPv6Address ?: "" },
```
oxtoacart, line 31: Thank you!

PR #318, opened 2024-04-12
android/ui: handle device approval like usual login

oxtoacart:


oxtoacart:
This also has the benefit of opening the admin panel in the same CustomTabs as where the user logged in.

barnstar:
I was half way through my own version of these changes, but I ended up copying most of this.  I extracted the IP for the URL in a slightly safer way (I hope).

PR #320, opened 2024-04-12
ui: don't show key expiry warning if key doesn't expire

agottardo:
Fixes tailscale/tailscale#11701

We should not show a key expiry warning if the KeyExpiry is a zero-value time.Time


barnstar (approved):
Good catch...

PR #321, opened 2024-04-12
ui: fix missing character in date format string for years in strings.xml

MDrollette:
Fix missing character in date format string


agottardo (approved):
Thanks!

PR #324, opened 2024-04-12
libtailscale, android: fix allow LAN access

kari-ts:
-Exclude local routes in VPNServiceBuilder
-Maybe update TUN on new state update where state >= starting -Clean up updateTUN

Updates tailscale/corp#18984
Updates tailscale/corp#18202

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/VPNServiceBuilder.kt
+++ b/android/src/main/java/com/tailscale/ipn/VPNServiceBuilder.kt
@@ -5,6 +5,8 @@ package com.tailscale.ipn
 
 import android.net.VpnService
 import libtailscale.ParcelFileDescriptor
+import java.net.InetAddress
+import android.net.IpPrefix as AndroidIpPrefix
 
 class VPNServiceBuilder(private val builder: VpnService.Builder) : libtailscale.VPNServiceBuilder {
   override fun addAddress(p0: String, p1: Int) {
@@ -17,6 +19,12 @@ class VPNServiceBuilder(private val builder: VpnService.Builder) : libtailscale.
 
   override fun addRoute(p0: String, p1: Int) {
     builder.addRoute(p0, p1)
+  }
+
+  override fun excludeRoute(p0: String, p1: Int) {
+    val inetAddress = InetAddress.getByName(p0)
+    val prefix = AndroidIpPrefix(inetAddress, p1)
+    builder.excludeRoute(prefix)
   }
 
   override fun addSearchDomain(p0: String) {
--- a/libtailscale/backend.go
+++ b/libtailscale/backend.go
@@ -5,6 +5,7 @@ package libtailscale
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
@@ -166,6 +167,27 @@ func (a *App) runBackend(ctx context.Context) error {
 		select {
 		case s := <-stateCh:
 			state = s
+			if cfg.rcfg != nil && state >= ipn.Starting && service != nil {
+				// On state change, check if there are router or config changes requiring an update to VPNBuilder
+				if err := b.updateTUN(service, cfg.rcfg, cfg.dcfg); err != nil {
+					if errors.Is(err, errMultipleUsers) {
+						// TODO: surface error to user
+					}
+					log.Printf("VPN update failed: %v", err)
+
+					mp := new(ipn.MaskedPrefs)
+					mp.WantRunning = false
+					mp.WantRunningSet = true
+
+					_, err := a.EditPrefs(*mp)
+					if err != nil {
+						log.Printf("localapi edit prefs error %v", err)
+					}
+
+					b.lastCfg = nil
+					b.CloseTUNs()
+				}
+			}
 		case n := <-netmapCh:
 			networkMap = n
 		case c := <-configs:
@@ -214,6 +236,8 @@ func (a *App) runBackend(ctx context.Context) error {
 				if err := b.updateTUN(service, cfg.rcfg, cfg.dcfg); err != nil {
 					log.Printf("VPN update failed: %v", err)
 					service.Close()
+					b.lastCfg = nil
+					b.CloseTUNs()
 				}
 			}
 		case s := <-onDisconnect:
--- a/libtailscale/interfaces.go
+++ b/libtailscale/interfaces.go
@@ -77,6 +77,7 @@ type VPNServiceBuilder interface {
 	AddDNSServer(string) error
 	AddSearchDomain(string) error
 	AddRoute(string, int32) error
+	ExcludeRoute(string, int32) error
 	AddAddress(string, int32) error
 	Establish() (ParcelFileDescriptor, error)
 }
--- a/libtailscale/localapi.go
+++ b/libtailscale/localapi.go
@@ -5,6 +5,7 @@ package libtailscale
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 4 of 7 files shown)


android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -83,6 +84,10 @@ object Notifier {
       state.collect { currstate ->
         readyToPrepareVPN.set(currstate > Ipn.State.Stopped)
         tileReady.set(currstate >= Ipn.State.Stopped)
+
+        if (currstate >= Ipn.State.Starting) {
+          Libtailscale.maybeUpdateTUN()
```
agottardo, line 89: Can we add a comment to explain why we're calling into this?

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -146,6 +146,10 @@ class App : Application(), libtailscale.AppContext {
         .editPrefs(Ipn.MaskedPrefs().apply { WantRunning = wantRunning }, callback)
   }
 
+  override fun setStopRunning(){
```
bradfitz, line 149: this isn't formatted the same way as line 155

Does Kotlin not have an auto-formatter?

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
barnstar: It feels to me like the UI shouldn't even do this.  If we always need to call this method when the state changes to !Starting, it should be done in goland and explained there.

barnstar (changes requested):

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
kari-ts: makes sense! moved this in to go.

libtailscale/backend.go
```diff
@@ -174,6 +175,18 @@ func (a *App) runBackend(ctx context.Context) error {
...
 			}
 			configErrs <- b.updateTUN(service, cfg.rcfg, cfg.dcfg)
+		case <-maybeUpdateTUN:
+			if cfg.rcfg != nil {
+				if err := b.updateTUN(service, cfg.rcfg, cfg.dcfg); err != nil {
+					if errors.Is(err, errMultipleUsers) {
+						log.Printf("Update TUN error: %v", errMultipleUsers)
+					}
+					log.Printf("VPN update failed: %v", err)
+					a.appCtx.SetStopRunning()
```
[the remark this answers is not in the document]
kari-ts: I've added a todo here to add in messaging to the user for the multiple users error

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -17,6 +17,7 @@ import kotlinx.coroutines.launch
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
+import libtailscale.Libtailscale
```
agottardo, line 20: can we drop this import now?

libtailscale/localapi.go
```diff
@@ -103,6 +108,12 @@ func (app *App) CallLocalAPIMultipart(timeoutMillis int, method, endpoint string
 	}
 }
 
+func (app *App) EditPrefs(prefs ipn.MaskedPrefs) (LocalAPIResponse, error) {
+	r, w := io.Pipe()
```
barnstar, line 112: You didn't write the masked prefs to the pipe here so the patch body will be empty.

barnstar (changes requested):

go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	tailscale.com v1.65.0-pre.0.20240418182715-02c6af2a69a8
 )
 
+replace tailscale.com => /Users/karilam/tailscale
```
agottardo, line 13: drop

barnstar (approved):
It looks fine, but seek a second opinion on the rcfg.LocalRoutes stuff.

libtailscale/net.go
```diff
@@ -170,6 +170,22 @@ func (b *backend) updateTUN(service IPNService, rcfg *router.Config, dcfg *dns.O
 			return err
 		}
 	}
+
+	for _, route := range rcfg.LocalRoutes {
+		// If route is localhost, do not exclude.
```
oxtoacart, line 175: Why do we not exclude localhost? And what about IPv6's loopback address?

libtailscale/localapi.go
```diff
@@ -18,9 +19,14 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"tailscale.com/ipn"
 )
 
-// CallLocalAPI calls the given endpoint on the local API using the given HTTP method
+const timeoutMillis = 30000
```
oxtoacart, line 26: On the Kotlin side, we specify the timeout on a per request basis, we just happen to default to 30000 millis. If this is really just the timeout for the `EditPrefs` call, it might make sense to hardcode that inside of `EditPrefs`.

libtailscale/interfaces.go
```diff
@@ -77,10 +77,16 @@ type VPNServiceBuilder interface {
 	AddDNSServer(string) error
 	AddSearchDomain(string) error
 	AddRoute(string, int32) error
+	ExcludeRoute(string, int32) error
 	AddAddress(string, int32) error
 	Establish() (ParcelFileDescriptor, error)
 }
 
+// IpPrefix corresponds to Android's IpPrefix
+type IpPrefix interface {
```
oxtoacart, line 86: IIUC, this doesn't need to be defined here because we don't use it anywhere on the Go side.

android/src/main/java/com/tailscale/ipn/VPNServiceBuilder.kt
```diff
@@ -38,3 +47,10 @@ class ParcelFileDescriptor(private val fd: android.os.ParcelFileDescriptor) :
     return fd.detachFd()
   }
 }
+
+class IpPrefix(private val str: String) : libtailscale.IpPrefix {
```
oxtoacart, line 51: I don't think you need this class, I think you can just move the function `createIpPrefix` from below into `VPNServiceBuilder`.

oxtoacart (approved):
Code generally LGTM and more importantly, it works in my testing :)

I left a few comments with suggested improvements.

libtailscale/localapi.go
bradfitz: use time.Duration

const timeout = 30 * time.Second

libtailscale/interfaces.go
bradfitz: IPPrefix in Go per https://go.dev/wiki/CodeReviewComments#initialisms

libtailscale/net.go
kari-ts: VpnService throws an exception if it's a loopback address - https://android.googlesource.com/platform/frameworks/base/+/c741553/core/java/android/net/VpnService.java#303. Updated to check for IPv6!

kari-ts: Upated to check IPv6 too! VpnService throws an exception for loopback addresses: https://android.googlesource.com/platform/frameworks/base/+/c741553/core/java/android/net/VpnService.java#303

PR #327, opened 2024-04-15
android/ui: support receiving Taildrop files directly to Downloads fo…

oxtoacart:
…lder

Closes tailscale/tailscale#11705

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -7,7 +7,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
-        android:maxSdkVersion="28" />
+        android:maxSdkVersion="29" />
     <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 
@@ -28,6 +28,7 @@
         android:name=".App"
         android:allowBackup="false"
         android:banner="@drawable/tv_banner"
+        android:requestLegacyExternalStorage="true"
         android:icon="@mipmap/ic_launcher"
         android:label="Tailscale"
         android:roundIcon="@mipmap/ic_launcher_round"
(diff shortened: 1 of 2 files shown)


android/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
-        android:maxSdkVersion="28" />
+        android:maxSdkVersion="29" />
```
oxtoacart, line 10: This makes sure we also request this permission on Android 10

android/src/main/AndroidManifest.xml
```diff
@@ -28,6 +28,7 @@
         android:name=".App"
         android:allowBackup="false"
         android:banner="@drawable/tv_banner"
+        android:requestLegacyExternalStorage="true"
```
oxtoacart, line 31: We need this in order to be able to use `Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS)` on Android 10.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -317,25 +315,6 @@ class App : Application(), libtailscale.AppContext {
       return null
     }
 
-  @Throws(IOException::class)
```
oxtoacart, line 320: Unused vestigial code related to Taildrop.

barnstar:
The issue ID in the commit appears to be incorrect...

barnstar:
> The issue ID in the commit appears to be incorrect...

tailscale/tailscale#11705

oxtoacart:
I tested this on an Android 10 emulator and verified the following:

1. Without this fix, we do in fact encounter the OG [issue](https://github.com/tailscale/tailscale/issues/11705)
2. With this fix, the permission prompt works as expected
3. With this fix, files that are Taildropped to the Android device successfully make it into the `Downloads` folder

I had originally hoped to be able to just not request this permission on Android 10, but I tried that and Taildrop was unable to save incoming files to `Downloads`.

barnstar (approved):

barnstar (approved):
🚀

PR #331, opened 2024-04-15
android/ui: support login via auth key

barnstar:
updates ENG-3269

Adds support for joining a tailnet with an auth key in the UI.

Refactors some of the look to put the different custom login options in on their own screens instead of the menu itself.

Moves the login flow logic to the base class for the viewModel where it belongs.  removes some vestigial code.

There is no failure feedback for invalid auth keys or broken control servers.  That will require some fixes to provide better feedback from localAPI/notifier, but the feature is otherwise fully operational.
Screenshot 2024-04-15 at 3 47 34 PM
Screenshot 2024-04-15 at 3 47 20 PM


android/src/main/res/values/strings.xml
```diff
@@ -105,16 +105,24 @@
...
+    <string name="custom_control_menu">Use an alternate server</string>
     <string name="custom_control_menu_desc">Enter the URL of an alternate Tailscale server or your self-hosted Headscale server.</string>
     <string name="custom_control_placeholder">https://my.custom.server.com</string>
+    <string name="auth_key_menu">Use an auth key</string>
+    <string name="auth_key_title">Add an account using an auth key</string>
+    <string name="auth_key_explanation">Enter a valid auth key generated by the Tailscale admin console</string>
+    <string name="auth_key_placeholder">ie: tskey-auth-mYta1ln3tCNTRL-s3cr3tauthk3yFr0madM1n</string>
+    <string name="invalidAuthKey">The provided auth key is not in the correct format.</string>
+    <string name="invalidAuthKeyTitle">Invalid Key</string>
+    <string name="custom_control_url_title">Custom control server url</string>
```
agottardo, line 124:
```suggestion
    <string name="custom_control_url_title">Custom control server URL</string>
```

android/src/main/res/values/strings.xml
```diff
@@ -105,16 +105,24 @@
...
-    <string name="custom_control_menu">Add account using alternate server</string>
+    <string name="custom_control_menu">Use an alternate server</string>
     <string name="custom_control_menu_desc">Enter the URL of an alternate Tailscale server or your self-hosted Headscale server.</string>
     <string name="custom_control_placeholder">https://my.custom.server.com</string>
+    <string name="auth_key_menu">Use an auth key</string>
+    <string name="auth_key_title">Add an account using an auth key</string>
+    <string name="auth_key_explanation">Enter a valid auth key generated by the Tailscale admin console</string>
+    <string name="auth_key_placeholder">ie: tskey-auth-mYta1ln3tCNTRL-s3cr3tauthk3yFr0madM1n</string>
+    <string name="invalidAuthKey">The provided auth key is not in the correct format.</string>
+    <string name="invalidAuthKeyTitle">Invalid Key</string>
```
kari-ts, line 123: lowercase key

kari-ts (approved):

android/src/main/res/values/strings.xml
barnstar: Fixed

PR #333, opened 2024-04-16
MainView: fix toggle animation

kari-ts:
Use persistent isOn state across recompositions

Updates tailscale/corp#18202
Fixes tailscale/corp#19194


barnstar (approved):

sonovawolf:
That was fast!

PR #336, opened 2024-04-17
android: pass interface name to go

kari-ts:
Use Android API to pass interface name to Tailscale on network updates

Fixes tailscale/corp#19215
Fixes tailscale/corp#19124

PR #339, opened 2024-04-18
Revert "android: pass interface name to go"

kari-ts:
Reverts tailscale/tailscale-android#336


bradfitz:
Reverts should always say why a bit.

Like "Reverts XXXXX. Merged accidentally/prematurely and broke the build."

bradfitz (approved):

PR #340, opened 2024-04-18
android: pass interface name to go

kari-ts:
Use Android API to pass interface name to Tailscale on network updates
Pull in OSS changes to include https://github.com/tailscale/tailscale/pull/11784 

Fixes tailscale/corp#19215

### DIFF
--- a/libtailscale/callbacks.go
+++ b/libtailscale/callbacks.go
@@ -23,13 +23,14 @@ var (
 	// onGoogleToken receives google ID tokens.
 	onGoogleToken = make(chan string)
 
-	// onDNSConfigChanged is notified when the network changes and the DNS config needs to be updated.
-	onDNSConfigChanged = make(chan struct{}, 1)
+	// onDNSConfigChanged is notified when the network changes and the DNS config needs to be updated. It receives the updated interface name.
+	onDNSConfigChanged = make(chan string, 1)
 )
 
-func OnDnsConfigChanged() {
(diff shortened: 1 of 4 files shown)


libtailscale/net.go
```diff
@@ -239,6 +240,7 @@ func (b *backend) NetworkChanged() {
 		if nm, ok := b.sys.NetMon.GetOK(); ok {
 			nm.InjectEvent()
```
bradfitz, line 241: per the docs you just wrote in the oss CL, you want to set the atomic first and THEN call this InjectEvent, so your line 243 should be up at 240, no?

libtailscale/callbacks.go
```diff
@@ -23,13 +23,14 @@ var (
...
 
-	// onDNSConfigChanged is notified when the network changes and the DNS config needs to be updated.
-	onDNSConfigChanged = make(chan struct{}, 1)
+	// onDNSConfigChanged is notified when the network changes and the DNS config needs to be updated. It receives the updated interface name.
+	onDNSConfigChanged = make(chan string, 1)
 )
 
-func OnDnsConfigChanged() {
+// ifname is the interface name retrieved from LinkProperties on network change, if any.
+func OnDnsConfigChanged(ifname string) {
```
bradfitz, line 31: Unrelated to this PR, but the style is OnDNSConfigChanged per https://go.dev/wiki/CodeReviewComments#initialisms and matching the variable name `onDNSConfigChanged` in backend.go

barnstar (approved):

PR #134, opened 2023-09-21
Makefile: add an emulator command to start an emulator

raggi:

PR #343, opened 2024-04-18
Makefile: add command to start emulator

kari-ts:


raggi:
Why prefer an arm image? That'll be slow on x86 machines?

See https://github.com/tailscale/tailscale-android/pull/134/files for another method, this also explicitly picks a google apis image

kari-ts:
Updated to not prefer ARM. https://github.com/tailscale/tailscale-android/pull/134/files didn't quite work for me since uname -u returns arm64 for me

raggi (approved):

PR #358, opened 2024-04-30
android: fix Quick Settings tailscale

kari-ts:
-Get rid of unused stopVPN() function
-Get rid of unused ACTION_STOP_VPN intent handling; this is redundant with DISCONNECT_VPN intent -Tile active state should only depend on ipn state, and not the results of editing the prefs with wantRunning set. It should be active iff ipn.State > Stopped

Fixes tailscale/tailscale#11920


oxtoacart:
I confirmed that this fixes the issue I had with inconsistent state of the quick settings tile after restarting Tailscale.

barnstar (approved):

agottardo (approved):

PR #359, opened 2024-04-30
android: accessibility fixes

kari-ts:
Updates tailscale/corp#18976


android/src/main/java/com/tailscale/ipn/ui/view/Avatar.kt
```diff
@@ -34,7 +34,7 @@ fun Avatar(profile: IpnLocal.LoginProfile?, size: Int = 50, action: (() -> Unit)
               indication = rememberRipple(bounded = false),
               onClick = action)
     }
-    Icon(imageVector = Icons.Default.Person, contentDescription = null, modifier = modifier)
+    Icon(imageVector = Icons.Default.Person, contentDescription = "Settings", modifier = modifier)
```
agottardo, line 37: Localizable?

agottardo (approved):

PR #360, opened 2024-05-01
android/makefile: build and sign standalone apk

barnstar:
Builds and signs a raw apk suitable for posting on packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,21 +86,26 @@ apk: $(DEBUG_APK) ## Build the debug APK
 tailscale-debug: $(DEBUG_APK) ## Build the debug APK
 
 .PHONY: release
-release: tailscale.jks $(RELEASE_AAB) $(RELEASE_APK) signaab signapk ## Build and sign the release AAB and APK
+release: signingkey $(RELEASE_AAB) $(RELEASE_APK) signaab signapk ## Build and sign the release AAB and APK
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -81,8 +86,17 @@ apk: $(DEBUG_APK) ## Build the debug APK
...
+release: tailscale.jks $(RELEASE_AAB) $(RELEASE_APK) signaab signapk ## Build and sign the release AAB and APK
+
+.PHONY: signaab
+signaab: tailscale.jks $(RELEASE_AAB) ## Sign the release AAB
 	jarsigner -sigalg SHA256withRSA -digestalg SHA-256 -keystore tailscale.jks $(RELEASE_AAB) tailscale
+	jarsigner -verify -certs $(RELEASE_AAB)
+
+.PHONY: signapk
+signapk: tailscale.jks $(RELEASE_APK) ## Sign the release APK
+	jarsigner -sigalg SHA256withRSA -digestalg SHA-256 -keystore tailscale.jks $(RELEASE_APK) tailscale
```
barnstar, line 98: @kari-ts I have only a cursory understanding of Android signing, but I think simply signing the apk with the same keys we sign the aab with *should* be sufficient for something we want to distribute directly.  Correct me if I'm completely wrong here.

PR #362, opened 2024-05-01
android: add persistent notification with VPN status

kari-ts:
-When connected, tapping on the notification disconnects 
-When disconnected, tapping on the notification connects 
-Navigate to system notifications instead of app info when tapping on 'Notifications' 
-Add text to notification permissions request explaining that the persistent notification is off by default
-Clean up unused notification channel and methods

-Follow up will be to update logo depending on connection status

Fixes tailscale/tailscale#10104
share_5306084378167387858
Screenshot (May 1, 2024 11_47_22 AM)

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -61,11 +61,9 @@ class App : Application(), libtailscale.AppContext {
   companion object {
     const val STATUS_CHANNEL_ID = "tailscale-status"
     const val STATUS_NOTIFICATION_ID = 1
-    const val NOTIFY_CHANNEL_ID = "tailscale-notify"
-    const val NOTIFY_NOTIFICATION_ID = 2
     private const val PEER_TAG = "peer"
     private const val FILE_CHANNEL_ID = "tailscale-files"
-    private const val FILE_NOTIFICATION_ID = 3
+    private const val FILE_NOTIFICATION_ID = 2
     private const val TAG = "App"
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -212,15 +208,33 @@ class App : Application(), libtailscale.AppContext {
...
+    val pendingIntent : PendingIntent = PendingIntent.getBroadcast(
+      this,
+      0,
+      intent,
+      PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+    )
     if (ready){
       startVPN()
     }
+    val notificationMessage = if (ready) "Connected" else "Disconnected"
```
agottardo, line 230: use string resources for localization

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -323,12 +337,22 @@ class App : Application(), libtailscale.AppContext {
     }
     val pending: PendingIntent =
         PendingIntent.getActivity(this, 0, viewIntent, PendingIntent.FLAG_UPDATE_CURRENT)
+    notify("File received", msg, FILE_CHANNEL_ID, pending, FILE_NOTIFICATION_ID)
```
agottardo, line 340: needs localization?

agottardo (approved):

agottardo:
It might be desirable to move the notifications logic out of `App.kt`, which is becoming larger and larger.

PR #365, opened 2024-05-03
android/ui: add mdm key expiry notification window

barnstar:
Updates tailscale/corp#19743

Adjust the key expiry window and it's related notification based on the keyExpiry MDM setting.  Default remains 24 hours.


agottardo (approved):

agottardo (approved):
Go Tests!!

agottardo (approved):

PR #368, opened 2024-05-03
App: add action button to persistent notification

kari-ts:
-Only show notification if connected
-Tapping persistent notification takes you to the app
-Tapping action button disconnects without taking you to the app
-Disconnecting dismisses the notification

Updates tailscale/tailscale#10104

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -65,6 +65,9 @@ class App : Application(), libtailscale.AppContext {
     private const val FILE_CHANNEL_ID = "tailscale-files"
     private const val FILE_NOTIFICATION_ID = 2
     private const val TAG = "App"
+    private const val NOTIF_REQ_CODE = 0
+    private const val NOTIF_ACTION_REQ_CODE = 1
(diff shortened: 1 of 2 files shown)


oxtoacart:
I found a few issues in testing.

1. When I tap on the notification, it takes me back to whatever view I had open in the Tailscale app. I think it would be best to open the home screen.

2. When I tap on the notification, it dismisses the notification, even if I didn't disconnect. I think the notification should remain as long as Tailscale is on.

3. When I tap "disconnect", it does disconnect Tailscale, but the notification is not dismissed.

4. If I disconnect Tailscale from the Tailscale UI, the notification remains. It should disappear.

oxtoacart:
IIRC, what I did the last time I had something like this, I ended up using a bound foreground service along with the VPN. Something like [this](https://medium.com/@manuaravindpta/is-it-possible-foreground-service-can-be-a-bound-service-2fedbfaac6dc#:~:text=A%20service%20can%20be%20started,device%20is%20under%20memory%20pressure.). The foreground service gets a persistent notification that sticks around as long as the service is running. Binding it to the VPNService makes the service go away when the VPN is stopped. And the nice thing about having a foreground service running is that it should make it less likely for the our app (and hence the VPN) to be killed.

PR #370, opened 2024-05-07
android: use EditPrefs instead of passing UpdatePrefs to start

oxtoacart:
This allows us to precisely set the options we need during login and avoid wiping away defaults like AllowSingleHost.

Updats tailscale/tailscale#11731

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
@@ -85,8 +85,11 @@ open class IpnViewModel : ViewModel() {
 
   // Login/Logout
 
-  fun login(options: Ipn.Options = Ipn.Options(), completionHandler: (Result<Unit>) -> Unit = {}) {
-
+  fun login(
+      maskedPrefs: Ipn.MaskedPrefs? = null,
+      authKey: String? = null,
+      completionHandler: (Result<Unit>) -> Unit = {}
+  ) {
     MDMSettings.loginURL.flow.value?.let {
       Log.d(TAG, "Using MDM derived control URL: $it")
       loginWithCustomControlURL(it, completionHandler)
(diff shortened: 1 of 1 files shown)


agottardo (approved):

MulverineX:
"Updats", whoops! 😝

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -85,14 +85,21 @@ open class IpnViewModel : ViewModel() {
 
   // Login/Logout
 
-  fun login(options: Ipn.Options = Ipn.Options(), completionHandler: (Result<Unit>) -> Unit = {}) {
-
+  fun login(
+      options: Ipn.Options = Ipn.Options(),
```
bradfitz, line 89: maybe you should just take the authKey string only? You don't have separate frontend-vs-backend log IDs on Android now, do you?

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -102,21 +109,24 @@ open class IpnViewModel : ViewModel() {
       }
     }
 
-    Client(viewModelScope).start(options) { start ->
-      start.onFailure { completionHandler(Result.failure(it)) }.onSuccess { loginAction() }
+    val startAction = {
+      Client(viewModelScope).start(options) { start ->
+        start.onFailure { completionHandler(Result.failure(it)) }.onSuccess { loginAction() }
```
bradfitz, line 114: if this is the style, that's fine, but I personally think it's sad that onFailure and onSuccess's bodies don't line up in any way.

bradfitz (approved):

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
oxtoacart: Yeah, that's the auto-formatter.

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
oxtoacart: Good call.

oxtoacart:
Thanks for the review @agottardo, @bradfitz and @MulverineX

PR #377, opened 2024-05-08
android: fix quick settings tile status

kari-ts:
https://github.com/tailscale/tailscale-android/pull/358 updated the Quick Settings tile to only depend on ipn state. This was only partially correct in the sense that we made changes to only check for whether the state was > stopped and not whether Tailscale was on.

This checks for two states, whether Tailscale is on, and whether the tile is ready to be used. The former requires ipn state to be >= Starting, and the latter checks whether ipn state is > RequiresMachineAuth. Tile readiness determines whether an intent is to open MainActivity or whether an intent to connect/disconnect VPN is sent. Whether Tailscale is on or off determines whether the tile status is active or not.

Updates tailscale/tailscale#11920

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -46,6 +46,9 @@
                 <category android:name="android.intent.category.LAUNCHER" />
                 <category android:name="android.intent.category.LEANBACK_LAUNCHER" />
             </intent-filter>
+            <intent-filter>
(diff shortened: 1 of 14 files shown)


agottardo:
@oxtoacart removed the tile from the manifest in https://github.com/tailscale/tailscale-android/pull/380, so I believe we'll need to cherry-pick a revert of that change into this PR.

oxtoacart (approved):

PR #378, opened 2024-05-08
android: lazily initialize App to avoid starting Tailscale when unnec…

oxtoacart:
…essary

For example, when viewing the QuickSettings tile, there's no need to start Tailscale's backend.

Updates #cleanup

Note - this does not solve the issue with Tailscale wanting to connect whenever you open up the activity, but it keeps the QuickSettings tile from starting the backend. That's good both because it keeps it from connecting, and also because it saves resources since we're not starting up background activity just for displaying the Quick Settings tile.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -70,7 +70,7 @@ class App : Application(), libtailscale.AppContext {
             .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
             .addCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN)
             .build()
-    lateinit var appInstance: App
+    private lateinit var appInstance: App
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -215,26 +225,17 @@ class App : Application(), libtailscale.AppContext {
...
-    this.action = action
-    }
-    val pendingIntent : PendingIntent = PendingIntent.getBroadcast(
-      this,
-      0,
-      intent,
-      PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
-    )
-    if (ready){
+    val intent = Intent(this, IPNReceiver::class.java).apply { this.action = action }
```
oxtoacart, line 228: Sorry, I don't know why this section got reformatted.

PR #389, opened 2024-05-13
android: make IPNService a foreground service

oxtoacart:
Prevents BackgroundServiceStartNotAllowedException.

Updates #cleanup

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -4,14 +4,13 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
         android:maxSdkVersion="29" />
     <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
-    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_SPECIAL_USE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_SYSTEM_EXEMPTED"/>
 
     <!-- Disable input emulation on ChromeOS -->
     <uses-feature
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -441,6 +415,30 @@ open class UninitializedApp : Application() {
...
+        PendingIntent.getBroadcast(
+            this, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE)
+    return NotificationCompat.Builder(this, STATUS_CHANNEL_ID)
+        .setSmallIcon(icon)
+        .setContentTitle("Tailscale")
+        .setContentText(message)
+        .setAutoCancel(!vpnRunning)
+        .setOnlyAlertOnce(!vpnRunning)
+        .setOngoing(vpnRunning)
+        .setSilent(true)
```
oxtoacart, line 439: Drive-by, make this silent.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -441,6 +415,30 @@ open class UninitializedApp : Application() {
...
+        .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+        .addAction(NotificationCompat.Action.Builder(0, actionLabel, pendingIntent).build())
```
oxtoacart, line 441: Drive-by, linter didn't like us using a content intent, wants us to use actions instead.

agottardo (approved):

bradfitz:
See http://go/cleanup :)

BackgroundServiceStartNotAllowedException sounds not good... was this crashing?

android/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,8 @@
         android:maxSdkVersion="29" />
     <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_SPECIAL_USE"/>
```
bradfitz, line 14: https://developer.android.com/about/versions/14/changes/fgs-types-required says:

> In addition to declaring the FOREGROUND_SERVICE_TYPE_SPECIAL_USE foreground service type, developers should declare use cases in the manifest. To do so, they specify the <property> element within the <service> element. These values and corresponding use cases are reviewed when you submit your app in the Google Play Console. The use cases you provide are free-form, and you should make sure to provide enough information to let the reviewer see why you need to use the specialUse type.

Will this get rejected by the play store without a reason?

agottardo: Sounds like we should use https://developer.android.com/about/versions/14/changes/fgs-types-required#system-exempted given it specifically mentions VPN apps as a use-case.

kari-ts (approved):

PR #391, opened 2024-05-13
android: make disconnected notification non-foreground

kari-ts:
Follow-up to https://github.com/tailscale/tailscale-android/pull/389
Use foreground notifications only when the service was started as a foreground service

Updates #cleanup

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -99,6 +99,7 @@ class App : UninitializedApp(), libtailscale.AppContext {
   override fun onTerminate() {
     super.onTerminate()
     Notifier.stop()
+    notificationManager.cancelAll()
     applicationScope.cancel()
   }
 
@@ -127,8 +128,12 @@ class App : UninitializedApp(), libtailscale.AppContext {
     applicationScope.launch {
       Notifier.state.collect { state ->
         val ableToStartVPN = state > Ipn.State.NeedsMachineAuth
+        // If VPN is stopped, show a disconnected notification. If it is running as a foregrround service, IPNService will show a connected notification.
+        if (state == Ipn.State.Stopped){
+          notifyStatus(false)
+        }
         val vpnRunning = state == Ipn.State.Starting || state == Ipn.State.Running
-        updateConnStatus(ableToStartVPN, vpnRunning)
+        updateConnStatus(ableToStartVPN)
         QuickToggleService.setVPNRunning(vpnRunning)
       }
     }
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -412,26 +414,23 @@ open class UninitializedApp : Application() {
...
     return NotificationCompat.Builder(this, STATUS_CHANNEL_ID)
-        .setSmallIcon(icon)
+        .setSmallIcon(R.drawable.ic_notification)
         .setContentTitle("Tailscale")
-        .setContentText(message)
-        .setAutoCancel(!vpnRunning)
-        .setOnlyAlertOnce(!vpnRunning)
-        .setOngoing(vpnRunning)
+        .setContentText(getString(R.string.connected))
+        .setAutoCancel(false)
```
oxtoacart, line 431: I think this is the default, so call can be removed

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -412,26 +414,23 @@ open class UninitializedApp : Application() {
...
+        .setOnlyAlertOnce(false)
```
oxtoacart, line 432: I think this is the default, so call can be removed

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -412,26 +414,23 @@ open class UninitializedApp : Application() {
       return
     }
     val nm: NotificationManagerCompat = NotificationManagerCompat.from(this)
-    nm.notify(STATUS_NOTIFICATION_ID, buildStatusNotification(vpnRunning))
+    nm.notify(STATUS_NOTIFICATION_ID, buildStatusNotification())
   }
 
-  fun buildStatusNotification(vpnRunning: Boolean): Notification {
-    val message = getString(if (vpnRunning) R.string.connected else R.string.not_connected)
-    val icon = if (vpnRunning) R.drawable.ic_notification else R.drawable.ic_notification_disabled
```
oxtoacart, line 420: The drawables and strings for the disconnected state can possibly be removed from resources, since they're no longer used.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -399,7 +401,7 @@ open class UninitializedApp : Application() {
     nm.createNotificationChannel(channel)
   }
 
-  protected fun notifyStatus(vpnRunning: Boolean) {
+  protected fun notifyStatus() {
```
oxtoacart, line 404: Maybe rename to "notifyRunning" or something like that.

oxtoacart (commented):
I left a few comments.

I'm not 100% sure we want to do this. Dropping this means that we lose the disconnected state status bar icon.

kari-ts:
Put disconnected notification back in, but it is no longer a persistent notification. Also fix issue so that the correct notification is used for always-on VPN and when IPNService is restarted after being killed

android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -41,6 +41,7 @@ open class IPNService : VpnService(), libtailscale.IPNService {
           // This means we were started by Android due to Always On VPN.
           // Get the application to make sure it's been initialized, then
           // request the VPN.
+          showForegroundNotification()
```
oxtoacart, line 44: Have you tested this? I think it might not work because I'm not sure that `startForeground()` has been called in this case.

kari-ts: Using non-foreground notification. Also canceling notifications when App closes to clear any lingering notifications, and only showing a disconnected notification when in the Stopped state.

oxtoacart (approved):
LGTM!

PR #394, opened 2024-05-14
android/docker: update makefile and dockerfile for build automation

barnstar:
Updates tailsale/corp#19670

Added a dockerfile where can run full release build in addition to  dropping into a shell.

The build will now look for JKS_PASSWORD in the environment for completing the signing step without user interaction.

Several smaller recipes added to the makefile for building the docker builder image, running and cleaning it up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
+## For signed release build JKS_PASSWORD must be set to the password for the jks keystore
+## and tailscale.jks must be present in the repo root.
+
 DEBUG_APK=tailscale-debug.apk
 RELEASE_AAB=tailscale-release.aab
 LIBTAILSCALE=android/libs/libtailscale.aar
@@ -82,7 +85,7 @@ tailscale-debug: $(DEBUG_APK) ## Build the debug APK
 
 .PHONY: release
 release: tailscale.jks $(RELEASE_AAB) ## Build the release AAB
-	jarsigner -sigalg SHA256withRSA -digestalg SHA-256 -keystore tailscale.jks $(RELEASE_AAB) tailscale
+	jarsigner -sigalg SHA256withRSA -digestalg SHA-256 -keystore tailscale.jks -storepass $(JKS_PASSWORD) $(RELEASE_AAB) tailscale
 
 # gradle-dependencies groups together the android sources and libtailscale needed to assemble tests/debug/release builds.
 .PHONY: gradle-dependencies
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -196,13 +199,24 @@ install: $(DEBUG_APK) ## Install the debug APK on a connected device
 run: install ## Run the debug APK on a connected device
 	adb shell am start -n com.tailscale.ipn/com.tailscale.ipn.MainActivity
 
-.PHONY: dockershell
-dockershell: ## Run a shell in the Docker build container
-	docker build -t tailscale-android .
-	docker run -v $(CURDIR):/build/tailscale-android -it --rm tailscale-android
+.PHONY: dockerEnv
+dockerEnv: ## Builds the docker image for the android build environment
```
raggi, line 203: we have kebab case and snake case and uncased concatenations in this makefile already, the traditional choice is kebab case or uncased concatenation - but I'd really love it if we don't introduce a fourth style. CamelCase is extremely rare in make because it breaks down on case insensitive filesystems.

raggi: Perhaps:

docker-image:

docker-build:

docker-remove-image:

Makefile
```diff
@@ -196,13 +199,24 @@ install: $(DEBUG_APK) ## Install the debug APK on a connected device
 run: install ## Run the debug APK on a connected device
 	adb shell am start -n com.tailscale.ipn/com.tailscale.ipn.MainActivity
 
-.PHONY: dockershell
-dockershell: ## Run a shell in the Docker build container
```
raggi, line 200: What is the replacement for this command? does it have no users anymore?

Makefile
```diff
@@ -196,13 +199,24 @@ install: $(DEBUG_APK) ## Install the debug APK on a connected device
...
-	docker build -t tailscale-android .
-	docker run -v $(CURDIR):/build/tailscale-android -it --rm tailscale-android
+.PHONY: dockerEnv
+dockerEnv: ## Builds the docker image for the android build environment
+	docker build  -f docker/AndroidBuild-amd64.DockerFile -t tailscale-android-build-amd64 .
+
+.PHONY: tailscale.jks dockerBuild
+dockerBuild: ## Runs the docker image for the android build environment and builds release
```
raggi, line 207: this should depend on the image build

Makefile
```diff
@@ -82,7 +85,7 @@ tailscale-debug: $(DEBUG_APK) ## Build the debug APK
 
 .PHONY: release
 release: tailscale.jks $(RELEASE_AAB) ## Build the release AAB
-	jarsigner -sigalg SHA256withRSA -digestalg SHA-256 -keystore tailscale.jks $(RELEASE_AAB) tailscale
+	jarsigner -sigalg SHA256withRSA -digestalg SHA-256 -keystore tailscale.jks -storepass $(JKS_PASSWORD) $(RELEASE_AAB) tailscale
```
raggi, line 88: does this fail if i make a release build but I don't have a password?

barnstar: The release build always required that you enter JKS password to enter manually.  There will be  a check in the corp builder to ensure JKS_PASSWORD is set and we'll error out before we try the build (and defer the remove-image job).  I'll also add a post build check to make sure things are properly signed.

Makefile
barnstar: Done.

Makefile
barnstar: Added back.  There's a legit use for it.

Makefile
barnstar: Fixed.

agottardo (approved):

raggi (approved):

PR #395, opened 2024-05-14
ui: add descriptions to notification channels

agottardo:
Provide descriptions that will be displayed in the Android system notification settings to describe the purpose of each notification channel to the user.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/StartVPNWorker.java
+++ b/android/src/main/java/com/tailscale/ipn/StartVPNWorker.java
@@ -45,7 +45,7 @@ public final class StartVPNWorker extends Worker {
         String channelId = "start_vpn_channel";
 
         // Use createNotificationChannel method from App.java
-        app.createNotificationChannel(channelId, "Start VPN Channel", NotificationManager.IMPORTANCE_DEFAULT);
+        app.createNotificationChannel(channelId, getApplicationContext().getString(R.string.vpn_start), getApplicationContext().getString(R.string.notifications_delivered_when_user_interaction_is_required_to_establish_the_vpn_tunnel), NotificationManager.IMPORTANCE_MAX);
 
         // Use prepareIntent if available.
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tailscale/ipn/StartVPNWorker.java
```diff
@@ -45,7 +45,7 @@ public Result doWork() {
         String channelId = "start_vpn_channel";
 
         // Use createNotificationChannel method from App.java
-        app.createNotificationChannel(channelId, "Start VPN Channel", NotificationManager.IMPORTANCE_DEFAULT);
+        app.createNotificationChannel(channelId, getApplicationContext().getString(R.string.vpn_start), getApplicationContext().getString(R.string.notifications_delivered_when_user_interaction_is_required_to_establish_the_vpn_tunnel), NotificationManager.IMPORTANCE_MAX);
```
kari-ts, line 48: https://developer.android.com/reference/android/app/NotificationManager#IMPORTANCE_MAX - this says it's unused?

agottardo: Yes, but we do want the behaviour of `IMPORTANCE_HIGH`: "Higher notification importance: shows everywhere, makes noise and peeks. May use full screen intents.". 

`IMPORTANCE_MAX` will get us that, in addition to whatever they will ever come up with next to make these notifications more prominent.

agottardo: Didn't notice it was _deprecated_. Fixed.

kari-ts: https://android.googlesource.com/platform/frameworks/base.git/+/master/core/java/android/app/NotificationManager.java#454 IMPORTANCE_MAX not being used right now

kari-ts (approved):

PR #396, opened 2024-05-14
MainViewModel: treat NoState -> Starting as starting

kari-ts:
https://github.com/tailscale/tailscale-android/assets/135075563/6c50e05f-93ba-4929-951e-2cca0054627f

Toggle should be on when user transitions from NoState -> Starting, but not when the user is in NoState. This change uses a placeholder string "--" when the user is in NoState.

Fixes tailscale/corp#19961

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
@@ -13,6 +13,7 @@ import com.tailscale.ipn.ui.util.PeerCategorizer
 import com.tailscale.ipn.ui.util.PeerSet
 import com.tailscale.ipn.ui.util.TimeUtil
 import com.tailscale.ipn.ui.util.set
+import com.tailscale.ipn.App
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -87,15 +96,18 @@ class MainViewModel : IpnViewModel() {
...
+      this == State.NoState -> R.string.placeholder
+      this == State.InUseOtherUser -> R.string.placeholder
+      this == State.NeedsLogin -> R.string.please_login
+      this == State.NeedsMachineAuth -> R.string.needs_machine_auth
+      this == State.Stopped -> R.string.stopped
+      this == State.Starting -> R.string.starting
+      this == State.Running -> R.string.connected
+      else -> R.string.placeholder
   }
+  val string = App.get().getString(resId)
```
agottardo, line 111:
```suggestion
```
Looks like `string` is not used?

agottardo (approved):

PR #397, opened 2024-05-15
Update README with Play Store beta testing track, Amazon link, F-Droid info

agottardo:


izelnakri:
Hi @agottardo & @andrew-d , I wanted to download this app on F-Droid given on Meta Quest 3, there is no Google Play store, however F-Droid can be sideloaded along with most app. I would **really** like to have this app, since `tailscaled` process can't be run on termux without rooting. F-Droid apk is the only way I can connect certain android devices to my Tailscale network it seems. Correct me if I'm wrong ;)

agottardo:
We distribute our own APKs as well:

https://pkgs.tailscale.com/stable/#android

PR #398, opened 2024-05-16
android: fix connect VPN permissions

kari-ts:
https://github.com/tailscale/tailscale-android/assets/135075563/64897cae-265d-4df4-a2c4-658879dec98b

-show VPN connection permissions after intro screen -make toggle state and main view take VPN preparedness into consideration

Fixes tailscale/tailscale#12148


android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -82,20 +105,47 @@ class MainViewModel : IpnViewModel() {
     }
   }
 
+  fun maybeShowVpnPermissionLauncher() {
```
agottardo, line 108: Perhaps rename to "showVPNPermissionLauncherIfUnauthorized"? Or whatever makes it clearer what it does. :)

agottardo (approved):

PR #399, opened 2024-05-16
android: enable proguard

kari-ts:


raggi (commented):
Does this still retain everything needed over JNI?

agottardo (approved):

kari-ts:
@raggi I think so, but I'm not sure. I included 
```
-keepclasseswithmembernames class * {
    native <methods>;
} 
```
and then I built the app and bashed on it a bit. Other than that, is there anything I can do to check to make sure that I'm keeping everything that we need?

raggi (approved):

raggi:
> @raggi I think so, but I'm not sure. I included
> 
> ```
> -keepclasseswithmembernames class * {
>     native <methods>;
> } 
> ```
> 
> and then I built the app and bashed on it a bit. Other than that, is there anything I can do to check to make sure that I'm keeping everything that we need?

Ah neat, I'd missed this in proguard in the past (or it's new).

PR #404, opened 2024-05-17
Notifier: init app if uninitialized

kari-ts:
Fixes tailscale/corp#20087

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
@@ -4,6 +4,7 @@
 package com.tailscale.ipn.ui.notifier
 
 import android.util.Log
+import com.tailscale.ipn.App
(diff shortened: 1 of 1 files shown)


agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -56,6 +57,9 @@ object Notifier {
   @OptIn(ExperimentalSerializationApi::class)
   fun start(scope: CoroutineScope) {
     Log.d(TAG, "Starting")
+    if (!::app.isInitialized) {
+      App.get()
+    }
```
barnstar, line 62: I think it would be better if this throws and the caller gets a slap on the wrist for creating a notifier without an app.

This is likely my bad for not calling App.start in the shareActivity onStart which is the only way I can see this happening?

PR #406, opened 2024-05-17
android/ui: unhide accounts if VPN is prepared

barnstar:
Updates tailscale/tailscale#12148

There was a small bug where we weren't rechecking the vpn permissions so the FUS would never show.  We'll now do that in the view model base case on initialization and again on vpn-enable.


agottardo (approved):

kari-ts:
Thanks for the fix!

PR #408, opened 2024-05-21
android: exit node banner ui improvements

kari-ts:
Screenshot_20240522-134127
Screenshot_20240522-134221

-show if device is running as an exit node
-if device is running as exit node, show button to stop running as exit node
-show exit node connection status
-for exit nodes, tapping 'disable' should result in the exit node being set to none but the banner still showing this exit node, with an 'enable' button which sets the exit node to this one

Updates https://github.com/tailscale/corp/issues/19122

Follow up will include:
-prevent user from running as exit node if it is using an exit node and vice versa instead of silently failing 
-further handle MDM state - no buttons for MDM
-add explanation box for MDM offline state


android/src/main/java/com/tailscale/ipn/ui/theme/Theme.kt
```diff
@@ -152,6 +152,18 @@ val ColorScheme.off: Color
 val ColorScheme.link: Color
   get() = onPrimaryContainer
 
+  val ColorScheme.exitOff: Color
```
agottardo, line 155: When surfaced within auto-completion, these names are unclear. Can we use more extended strings such as `exitNodeOffline`, `runAsExitNodeContainer`, etc.?

android/src/main/java/com/tailscale/ipn/ui/theme/Theme.kt
```diff
@@ -287,6 +334,44 @@ val ColorScheme.secondaryButton: ButtonColors
     }
   }
 
+  val ColorScheme.offButton: ButtonColors
```
agottardo, line 337: Same: `exitNodeOfflineButton`

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -208,6 +220,24 @@ open class IpnViewModel : ViewModel() {
 
   // Exit Node Manipulation
 
+  class AdvertisedRoutesHelper() {
```
agottardo, line 223: Can you please move `AdvertisedRoutesHelper` out to a dedicated file?

android/src/main/java/com/tailscale/ipn/ui/theme/Theme.kt
kari-ts: @sonovawolf curious what your thoughts are here. I tried to give them fairly generic names in case we'd ever want to reuse them. Should we plan for these colors to be very specific, and name them accordingly?

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
kari-ts: Can you explain why? `AdvertisedRoutesHelper` is only used in `IpnViewModel`, so keeping it here makes it easier to understand the context it's used in

agottardo: It makes it easier to search for things across the repo. For the iOS project we follow a "one type for one file" rule, which means you can look up the file containing a given type without using full search.

android/src/main/java/com/tailscale/ipn/ui/theme/Theme.kt
sonovawolf: If you want to go totally generic you can use the [standard color aliases](https://login.tailscale.com/admin/design/colors) like `red-500`.

If you need a semantic one, I'd either use the Android ones like `surfaceContainer` or make a very specific one like `exitNodeOffline`. Android variables by default don't really have a concept of "intent" or "role" like many other design systems, but I defined some [here](https://www.figma.com/design/QgX0rndA763pwggR7ctHLh/Mobile-Apps?node-id=5981-1437&t=aQeNJ3KK5TnSh2mG-4). I'm not totally sure how many of them are in the code already.

For the card backgrounds I'd use `error` and `warning` respectively. For the text I'd use `onError` and `onWarning`, and for the button I'd just use custom colors since it's a one-off use case.

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
kari-ts: Done.

android/src/main/java/com/tailscale/ipn/ui/theme/Theme.kt
kari-ts: Thanks! I'm going to go ahead and use the ones defined in the Figma so we can potentially reuse.

sonovawolf (approved):

PR #413, opened 2024-05-29
IpnViewModel: fix NPE

kari-ts:
Fixes tailscale/tailscale#12281

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
@@ -75,8 +75,10 @@ open class IpnViewModel : ViewModel() {
 
     viewModelScope.launch {
       Notifier.prefs.collect {
-        it?.let { lastPrefs = it }
-        isRunningExitNode.set(it?.let { AdvertisedRoutesHelper.exitNodeOnFromPrefs(it) })
+        it?.let {
+          lastPrefs = it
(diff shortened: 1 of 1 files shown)


agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -76,7 +76,9 @@ open class IpnViewModel : ViewModel() {
     viewModelScope.launch {
       Notifier.prefs.collect {
         it?.let { lastPrefs = it }
-        isRunningExitNode.set(it?.let { AdvertisedRoutesHelper.exitNodeOnFromPrefs(it) })
+        if (it != null) {
+          isRunningExitNode.set(AdvertisedRoutesHelper.exitNodeOnFromPrefs(it))
+        }
       }
```
oxtoacart, line 82: One could also write this whole block as

```
it?.let {
    lastPrefs = it
    isRunningExitNode.set(AdvertisedRoutesHelper.exitNodeOnFromPrefs(it))
}
```

oxtoacart: Also, do we need to set `isRunningExitNode` to `false` if there are no prefs?

kari-ts: There are several places that check `isRunningExitNode`, so we could either have a default value or enforce that all checks do a null check

oxtoacart (approved):

PR #426, opened 2024-06-19
ui: deliver health notifications to user

agottardo:
This PR adds support for notifying the user when health warnings are sent down coming from LocalAPI. We remove duplicates and debounce updates; then deliver a notification for each health warning are they are sent down. Just like on macOS, notifications are removed when a Warnable becomes healthy again.

Notifications are delivered on a separate notification channel, so they can be disabled if needed.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -28,6 +28,7 @@ import com.tailscale.ipn.mdm.MDMSettings
 import com.tailscale.ipn.ui.localapi.Client
 import com.tailscale.ipn.ui.localapi.Request
 import com.tailscale.ipn.ui.model.Ipn
+import com.tailscale.ipn.ui.notifier.HealthNotifier
 import com.tailscale.ipn.ui.notifier.Notifier
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -71,6 +72,7 @@ class App : UninitializedApp(), libtailscale.AppContext {
   val dns = DnsConfig()
   private lateinit var connectivityManager: ConnectivityManager
   private lateinit var app: libtailscale.Application
+  private var healthNotifier: HealthNotifier? = null
 
   override fun getPlatformDNSConfig(): String = dns.dnsConfigAsString
 
@@ -92,6 +94,11 @@ class App : UninitializedApp(), libtailscale.AppContext {
         getString(R.string.taildrop_file_transfers),
         getString(R.string.notifications_delivered_when_a_file_is_received_using_taildrop),
         NotificationManagerCompat.IMPORTANCE_DEFAULT)
+    createNotificationChannel(
+        HealthNotifier.HEALTH_CHANNEL_ID,
+        getString(R.string.health_channel_name),
+        getString(R.string.health_channel_description),
+        NotificationManagerCompat.IMPORTANCE_HIGH)
     appInstance = this
     setUnprotectedInstance(this)
   }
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/tailscale/ipn/ui/notifier/HealthNotifier.kt
```diff
@@ -0,0 +1,117 @@
...
+    val currentWarnableCodes = warnings.map { it.WarnableCode }.toSet()
+
+    val addedWarnings: MutableSet<String> = mutableSetOf()
+    for (warning in warnings) {
+      if (ignoredWarnableCodes.contains(warning.WarnableCode)) {
+        continue
+      }
+
+      addedWarnings.add(warning.WarnableCode)
+      if (this.currentWarnings.contains(warning.WarnableCode)) {
```
barnstar, line 69: A when clause would be more kotlin-esque here.

android/src/main/java/com/tailscale/ipn/ui/notifier/HealthNotifier.kt
```diff
@@ -0,0 +1,117 @@
...
+        // Already notified, skip
+        continue
+      } else if (warning.DependsOn?.let {
+        it.any { depWarnableCode -> currentWarnableCodes.contains(depWarnableCode) }
```
barnstar, line 73: Maybe make this a function on the warning so the logic for this check doesn't live here... 
```
else if (warning.dependsOn(currentWarnableCodes)) {

}
```

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -92,6 +94,12 @@ class App : UninitializedApp(), libtailscale.AppContext {
         getString(R.string.taildrop_file_transfers),
         getString(R.string.notifications_delivered_when_a_file_is_received_using_taildrop),
         NotificationManagerCompat.IMPORTANCE_DEFAULT)
```
barnstar, line 96: This whole file could use a reformat... The indentation is all over the place....

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -71,6 +72,7 @@ class App : UninitializedApp(), libtailscale.AppContext {
   val dns = DnsConfig()
   private lateinit var connectivityManager: ConnectivityManager
   private lateinit var app: libtailscale.Application
+  private lateinit var healthNotifier: HealthNotifier
```
barnstar, line 75: It doesn't look like this is referenced anywhere but if it *is*, optional >> lateinit...   Lots of NPE bugs in need of fixing due to when onCreate() runs...

android/src/main/res/values/strings.xml
```diff
@@ -263,4 +263,6 @@
     <string name="notifications_delivered_when_user_interaction_is_required_to_establish_the_vpn_tunnel">Notifications delivered when user interaction is required to establish the VPN tunnel.</string>
     <string name="optional_notifications_which_display_the_status_of_the_vpn_tunnel">Optional notifications which display the status of the VPN tunnel.</string>
     <string name="notifications_delivered_when_a_file_is_received_using_taildrop">Notifications delivered when a file is received using Taildrop.</string>
+    <string name="health_channel_name">Health</string>
+    <string name="health_channel_description">Notifications about the health of the VPN connection.</string>
```
barnstar, line 267: I'm not convinced we want to expose this to the user as "Health"...  It has a slightly negative context.  @sonovawolf should probably weigh in here...

barnstar (approved):
Noted a couple of stylistic changes to clean things up a bit, but logically this looks great.

android/src/main/java/com/tailscale/ipn/App.kt
agottardo: Done, good call

android/src/main/java/com/tailscale/ipn/App.kt
agottardo: It actually looks nice in Android Studio (ktfmt is taking care of this). There's something off with the rendering on GitHub.

android/src/main/res/values/strings.xml
agottardo: Changed this for now to be a bit more generic:

> **Errors and warnings**
> This notification category is used to deliver important status notifications and should be left enabled. For instance, it is used to notify you about errors or warnings that affect Internet connectivity.

android/src/main/java/com/tailscale/ipn/ui/notifier/HealthNotifier.kt
agottardo: done!

android/src/main/res/values/strings.xml
agottardo: Will review tomorrow with @sonovawolf

PR #431, opened 2024-06-24
ui: add sheet to ping devices and see relay status

agottardo:
This PR adds the ability to ping other devices in your tailnet from the Android app, similarly to the current functionality on iOS. The ping view displays the current latency value, a chart with latency over time, and whether you are using a direct/relayed connection.


android/build.gradle
```diff
@@ -120,6 +120,7 @@ dependencies {
     // Supporting libraries.
     implementation("io.coil-kt:coil-compose:2.6.0")
     implementation("com.google.zxing:core:3.5.1")
+    implementation("com.patrykandpatrick.vico:compose:2.0.0-alpha.21")
```
barnstar, line 123: There is a stable 1.15 version of this.  Is that suitable over the alpha?

There is also Apache licenses so we need to include the proper attributions.

android/src/main/java/com/tailscale/ipn/ui/view/PingView.kt
```diff
@@ -0,0 +1,211 @@
...
+                      tint = Color.Red)
+                  Text(
+                      stringResource(id = R.string.pingFailed),
+                      fontSize = MaterialTheme.typography.titleMedium.fontSize,
+                      fontFamily = MaterialTheme.typography.titleMedium.fontFamily,
+                      fontWeight = FontWeight.Bold,
+                      textAlign = TextAlign.Center,
+                      color = Color.Red)
+                  Text(
+                      error,
```
barnstar, line 192: The error that comes back here isn't really human readable when you ping an offline device (and I will assume for other localAPI errors).  We should render something canned "Unable to reach device" or just elide the detailed error altogether.

go.Universe$proxyError<blah> or something.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -597,6 +615,38 @@ fun PeerList(
                                     shape = RoundedCornerShape(percent = 50))) {}
                     Spacer(modifier = Modifier.size(8.dp))
                     Text(text = peer.displayName, style = MaterialTheme.typography.titleMedium)
+                    DropdownMenu(
```
barnstar, line 618: Long press doesn't work on Android TV.   We should provide a way to access ping from the details view or just embed the composable right in the details view and show your ping times automatically.

android/src/main/java/com/tailscale/ipn/ui/viewModel/PingViewModel.kt
```diff
@@ -0,0 +1,115 @@
+package com.tailscale.ipn.ui.viewModel
+
```
barnstar, line 2: Missing the license headers on a few files.

android/src/main/java/com/tailscale/ipn/ui/viewModel/PingViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+            }
+          }
+        }
+      }
+      statusResult.onFailure { Log.d(TAG, "Failed to fetch status: $it") }
+    }
+  }
+}
+
+class PingViewModelFactory(private val peer: Tailcfg.Node) : ViewModelProvider.Factory {
```
barnstar, line 111: For consistency, put this up at the top of the file.

barnstar (changes requested):
Works great.  Some issues with Android TV though.

android/build.gradle
agottardo: Downgraded. https://github.com/tailscale/corp/issues/21114 tracks updating the LICENSES

android/src/main/java/com/tailscale/ipn/ui/viewModel/PingViewModel.kt
agottardo: done

android/src/main/java/com/tailscale/ipn/ui/view/PingView.kt
agottardo: Fixed this, errors now get remapped to either of two strings (depending on whether "timeout" is returned within the error string)

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
agottardo: Added to the peer details view as well, with a button in the top bar.

barnstar (approved):

PR #441, opened 2024-07-08
mdm: support split tunneling configuration via syspolicy

agottardo:
Updates tailscale/tailscale#6912

Adds two new Android-only MDM policies: IncludedPackageNames and ExcludedPackageNames. These are comma-separated string values that contain Android package names to configure app-based split tunneling programmatically.

If ExcludedPackageNames is non-empty, Tailscale will exclude the given apps from the VPN tunnel.

If IncludedPackageNames is non-empty, Tailscale will configure the VPN tunnel to only route the given apps via Tailscale.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/IPNService.kt
+++ b/android/src/main/java/com/tailscale/ipn/IPNService.kt
@@ -9,6 +9,7 @@ import android.net.VpnService
 import android.os.Build
 import android.system.OsConstants
 import android.util.Log
+import com.tailscale.ipn.mdm.MDMSettings
 import libtailscale.Libtailscale
 import java.util.UUID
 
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -114,12 +115,24 @@ open class IPNService : VpnService(), libtailscale.IPNService {
...
     b.setUnderlyingNetworks(null) // Use all available networks.
 
-    // Prevent certain apps from getting their traffic + DNS routed via Tailscale:
-    // - any app that the user manually disallowed in the GUI
-    // - any app that we disallowed via hard-coding
-    for (disallowedPackageName in UninitializedApp.get().disallowedPackageNames()) {
-      Log.d(TAG, "Disallowing app: $disallowedPackageName")
-      disallowApp(b, disallowedPackageName)
+    val includedPackages: List<String> =
+        MDMSettings.includedPackages.flow.value?.split(",") ?: emptyList()
```
barnstar, line 119: You should trim these strings too (remove whitespace and newlines), in case somebody tosses in a space or a linefeed.

android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -114,12 +115,24 @@ open class IPNService : VpnService(), libtailscale.IPNService {
...
+        b.addAllowedApplication(packageName)
+      }
+    } else {
+      // Otherwise, prevent certain apps from getting their traffic + DNS routed via Tailscale:
+      // - any app that the user manually disallowed in the GUI
+      // - any app that we disallowed via hard-coding
+      for (disallowedPackageName in UninitializedApp.get().disallowedPackageNames()) {
+        Log.d(TAG, "Disallowing app: $disallowedPackageName")
+        disallowApp(b, disallowedPackageName)
+      }
```
barnstar, line 135: If both lists are non-empty, log an error that you're ignoring the excluded list.

barnstar (changes requested):

android/src/main/java/com/tailscale/ipn/IPNService.kt
agottardo: disallowedPackageNames is never going to be empty. We have hardcoded exclusions.

barnstar (approved):

PR #451, opened 2024-07-24
android: prepare VPN when quick tile is clicked

kari-ts:
Currently, the VPN is prepared when MainActivity is launched. If Tailscale is enabled by a quick tile, the VPN is not prepared.

Fixes tailscale/tailscale#12489

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -9,7 +9,6 @@ import android.app.NotificationChannel
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
-import android.content.SharedPreferences
 import android.content.pm.PackageManager
 import android.net.ConnectivityManager
 import android.net.LinkProperties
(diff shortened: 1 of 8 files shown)


agottardo (approved):

android/src/main/java/com/tailscale/ipn/QuickToggleService.java
```diff
@@ -91,7 +85,22 @@ private void onTileClick() {
...
+            if (intent != null) {
+                // VPN permission is not granted, start MainActivity to request it
+                startMainActivity();
+            }
             app.startVPN();
         }
     }
+
+    private void startMainActivity() {
+        Intent i = getPackageManager().getLaunchIntentForPackage(getPackageName());
```
barnstar, line 98: Is there a broader issue here in that some of the VPN prep is dependent on the main activity getting started?   Can we decouple those things so that the burden is not on all intents or other activities to launch the main activity?

kari-ts: Thanks for pointing this out. Moved VPN prep into App

barnstar (approved):

barnstar (approved):

PR #458, opened 2024-08-02
android: add HealthView

agottardo:
Updates tailscale/tailscale#4136

This PR adds a proper health warnings viewer for the Android client, like we already do on iOS and macOS. Since 1.68, we have been displaying health warnings as notification bubbles in the HomeView, but this does not scale and does not allow us to surface low-priority warnings in a subtle way. 

Now, a small info circle or exclamation mark icon is displayed next to the connection status when one or more warnings are found. Tapping on the icon, a details view provides visibility into the full list of health warnings, sorted by severity level.


android/src/main/java/com/tailscale/ipn/ui/view/HealthView.kt
```diff
@@ -0,0 +1,97 @@
...
+    Box(
+        modifier =
+            Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp)
+                .clip(shape = RoundedCornerShape(10.dp, 10.dp, 10.dp, 10.dp))
+                .fillMaxWidth()) {
+          ListItem(
+              colors = warning.Severity.listItemColors(),
+              headlineContent = {
+                Text(
+                    warning.Title,
```
nickkhyl, line 88: The `Title` field appears to be optional. As an example, I think it is not used with [legacy errors](https://github.com/tailscale/tailscale/blob/main/health/healthmsg/healthmsg.go).

Should we hide the headline if `warning.Title` is empty?

android/src/main/java/com/tailscale/ipn/ui/view/HealthView.kt
```diff
@@ -0,0 +1,97 @@
...
+                    style = MaterialTheme.typography.titleMedium,
+                )
+              },
+              supportingContent = {
+                Text(warning.Text, style = MaterialTheme.typography.bodyMedium)
```
nickkhyl, line 93: Minor, kind of related to the above, and perhaps should be fixed in oss.. But it also appears that some legacy error messages might start with a lower-case word, as error strings [should not be capitalized](https://go.dev/wiki/CodeReviewComments#error-strings) in go. I think it would be nice to capitalize them for warnables, but likely not here.

nickkhyl (approved):
LGTM. Just two minor things, one of which should probably be fixed in oss.

android/src/main/java/com/tailscale/ipn/ui/view/HealthView.kt
agottardo: Fixed.

android/src/main/java/com/tailscale/ipn/ui/view/HealthView.kt
agottardo: +1; I'll open a separate issue for this, it should definitely be handled in OSS (health package, before sending stuff over to LocalAPI).

PR #461, opened 2024-08-06
mdm: improve handling and returning of not configured policy settings

nickkhyl:
We should distinguish between unconfigured policy settings and those configured with the default values. In the first case, the `syspolicyHandler` should return `syspolicy.ErrNoSuchKey` instead of the default value, while in the latter case, it should return the actual setting value, even if that value happens to be the default value such as `user-decides`. This distinction should also be reflected in the "Current MDM settings" view.

In this PR, we update `MDMSetting.flow` to hold both the value to be used by the app and a flag indicating whether the policy setting is configured or not. If the policy setting is not configured, the value is the default value for the setting type. We then use this new flag to decide whether to throw a `NoSuchKeyException` from the Kotlin-side of the `syspolicyHandler` implementation and how to display the policy setting in the "Current MDM settings" view.

Additionally, we update `MDMSettings.update` and `MDMSetting.setFrom` methods to avoid calling `app.getEncryptedPrefs` (and reading/decrypting the prefs) for every defined MDM setting.

Updates tailscale/tailscale#12687

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -316,21 +316,25 @@ class App : UninitializedApp(), libtailscale.AppContext {
   @Throws(
       IOException::class, GeneralSecurityException::class, MDMSettings.NoSuchKeyException::class)
   override fun getSyspolicyStringValue(key: String): String {
-    return MDMSettings.allSettingsByKey[key]?.flow?.value?.toString()
-        ?: run {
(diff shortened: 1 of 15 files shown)


nickkhyl:
The main objective was to correct the values and errors returned by `syspolicyHandler` to meet the backend expectations. But here is the updated debug view with the unconfigured policy settings marked as `[not set]`:

agottardo (approved):

PR #465, opened 2024-08-14
android: fix paddings and headers of Taildrop destination picker

agottardo:
Updates tailscale/corp#22362

First round of polish for the Taildrop device picker, to use more consistent metrics and SectionDivider resembling the rest of the app. We'll follow up with device icons like the ones we have on iOS in a later PR.


agottardo:
To be merged after the release.

barnstar (approved):

PR #466, opened 2024-08-15
taildrop: use a random filename if real filename cannot be determined

oxtoacart:
Also pull in latest oss to avoid crashing if sharing fails.

Updates tailscale/corp#22357

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ShareActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/ShareActivity.kt
@@ -9,6 +9,7 @@ import android.os.Build
 import android.os.Bundle
 import android.provider.OpenableColumns
 import android.util.Log
+import android.webkit.MimeTypeMap
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tailscale/ipn/ShareActivity.kt
```diff
@@ -92,7 +94,20 @@ class ShareActivity : ComponentActivity() {
             val nameCol = c.getColumnIndex(OpenableColumns.DISPLAY_NAME)
             val sizeCol = c.getColumnIndex(OpenableColumns.SIZE)
             c.moveToFirst()
-            val name = c.getString(nameCol)
+            val name: String =
+                c.getString(nameCol)
+                    ?: run {
+                      // For some reason, some content resolvers don't return a name.
+                      // Try to build a name from a random integer plus file extension
+                      // (if type can be determined), else just a random integer.
```
oxtoacart, line 102: Maybe there's a better algorithm we can use than just some random integer, I'm not sure.

agottardo (approved):

PR #471, opened 2024-08-16
Revert "VPNServiceBuilder: document excludeRoute not supported on pre-33 API"

agottardo:
Reverts tailscale/tailscale-android#467


kelivel (approved):

bradfitz:
Revert commit messages should always say why.

PR #473, opened 2024-08-19
android: prepare VPN when quick tile is clicked

kari-ts:
Currently, the VPN is prepared when MainActivity is launched. If Tailscale is enabled by a quick tile, the VPN is not prepared. This change creates an application scoped view model and moves the VPN prep to the application class so that it is not dependent on MainActivity.

Fixes tailscale/tailscale#12489

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -22,6 +22,9 @@ import android.util.Log
 import androidx.core.app.ActivityCompat
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.ViewModelStore
+import androidx.lifecycle.ViewModelStoreOwner
 import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKey
 import com.tailscale.ipn.mdm.MDMSettings
@@ -30,6 +33,8 @@ import com.tailscale.ipn.ui.localapi.Request
 import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.notifier.HealthNotifier
 import com.tailscale.ipn.ui.notifier.Notifier
+import com.tailscale.ipn.ui.viewModel.VpnViewModel
+import com.tailscale.ipn.ui.viewModel.VpnViewModelFactory
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/VpnViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+  @Suppress("UNCHECKED_CAST")
+  override fun <T : ViewModel> create(modelClass: Class<T>): T {
+    if (modelClass.isAssignableFrom(VpnViewModel::class.java)) {
+      return VpnViewModel(application) as T
+    }
+    throw IllegalArgumentException("Unknown ViewModel class")
+  }
+}
+
+class VpnViewModel(application: Application) : AndroidViewModel(application) {
```
agottardo, line 24: Can we document what this does, and why it was added?

android/src/main/java/com/tailscale/ipn/ui/viewModel/VpnViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+  init {
+    prepareVpn()
+  }
+
+  private fun prepareVpn() {
+    // Check if the user has granted permission yet.
+    if (!vpnPrepared.value) {
+      val vpnIntent = VpnService.prepare(getApplication())
+      if (vpnIntent != null) {
+        setVpnPrepared(false)
```
agottardo, line 38: I'd add `Log.d` with an appropriate `TAG` in both cases. It's helpful to have plenty of debug logs in all this initialization logic, to aid debugging in case issues come up.

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -81,7 +94,7 @@ class MainViewModel : IpnViewModel() {
     viewModelScope.launch {
       var previousState: State? = null
 
-      combine(Notifier.state, vpnPrepared) { state, prepared -> state to prepared }
+      combine(Notifier.state, isVpnPrepared) { state, prepared -> state to prepared }
```
agottardo, line 97: Can we just reference `vpnViewModel.vpnPrepared` instead of adding an extra new `val` which only gets used once?

agottardo (approved):
Some minor suggestions for clarity.

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -72,6 +76,15 @@ class App : UninitializedApp(), libtailscale.AppContext {
   val dns = DnsConfig()
   private lateinit var connectivityManager: ConnectivityManager
   private lateinit var app: libtailscale.Application
+
+  override val viewModelStore: ViewModelStore
+    get() = appViewModelStore
+
+  lateinit var vpnViewModel: VpnViewModel
+    private set
```
agottardo, line 84: This `private set` is probably causing the build failure.

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
kari-ts: mainViewModel.isVpnPrepared is also used by MainView

PR #475, opened 2024-08-20
android: check if other VPN is active

kari-ts:
Detect when another VPN is active and launch dialog giving user the option to navigate to settings to disable. Update state string and toggle to require successful VPN preparation

To do in a follow-up: monitor VPN connection, and if Tailscale VPN disconnects due to another VPN connecting, update toggle and text

Updates tailscale/tailscale#12850

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/MainActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/MainActivity.kt
@@ -5,12 +5,15 @@ package com.tailscale.ipn
 
 import android.annotation.SuppressLint
 import android.app.Activity
+import android.app.AlertDialog
 import android.content.Context
 import android.content.Intent
 import android.content.RestrictionsManager
 import android.content.pm.ActivityInfo
 import android.content.res.Configuration.SCREENLAYOUT_SIZE_LARGE
 import android.content.res.Configuration.SCREENLAYOUT_SIZE_MASK
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
 import android.os.Bundle
 import android.provider.Settings
 import android.util.Log
@@ -129,8 +132,13 @@ class MainActivity : ComponentActivity() {
             vpnViewModel.setVpnPrepared(true)
             App.get().startVPN()
           } else {
-            Log.d("VpnPermission", "VPN permission denied")
-            vpnViewModel.setVpnPrepared(false)
+            if (isAnotherVpnActive(this)) {
+              Log.d("VpnPermission", "Another VPN is likely active")
+              showOtherVPNConflictDialog()
+            } else {
+              Log.d("VpnPermission", "Permission was denied by the user")
+              viewModel.setVpnPrepared(false)
+            }
           }
         }
     viewModel.setVpnPermissionLauncher(vpnPermissionLauncher)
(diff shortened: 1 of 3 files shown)


android/src/main/res/values/strings.xml
```diff
@@ -292,4 +292,11 @@
     <string name="health_warnings">Health warnings</string>
     <string name="no_issues_found">No issues found</string>
     <string name="tailscale_is_operating_normally">Tailscale is operating normally.</string>
+
+    <!-- Strings for the multiple VPNs dialog -->
+    <string name="vpn_permission_denied">VPN permission denied</string>
+    <string name="multiple_vpn_explainer">Only one VPN can be active, and it appears another is already running. Before starting Tailscale, disable the other VPN.</string>
+    <string name="go_to_settings">Go to settings</string>
```
agottardo, line 299:
```suggestion
    <string name="go_to_settings">Go to Settings</string>
```

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -285,6 +293,34 @@ class MainActivity : ComponentActivity() {
     lifecycleScope.launch { Notifier.loginFinished.collect { _ -> loginQRCode.set(null) } }
   }
 
+  private fun showPermissionDeniedDialog() {
```
agottardo, line 296: Perhaps, use a function name more relevant to the purpose? For instance, "showOtherVPNConflictDialog()".

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -285,6 +293,34 @@ class MainActivity : ComponentActivity() {
...
+  fun isAnotherVpnActive(context: Context): Boolean {
+    val connectivityManager =
+        context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+
+    val activeNetwork = connectivityManager.activeNetwork
+    if (activeNetwork != null) {
+      val networkCapabilities = connectivityManager.getNetworkCapabilities(activeNetwork)
+      if (networkCapabilities != null &&
+          networkCapabilities.hasTransport(NetworkCapabilities.TRANSPORT_VPN)) {
+        return true
```
agottardo, line 318: Future improvement: I don't know if there is a way to know this, but it would be extremely useful to log the name of the other VPN app here. It would allow the support team to point the user to the app that is conflicting. Perhaps the active network contains a field with more information about the app that created it.

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -285,6 +293,34 @@ class MainActivity : ComponentActivity() {
...
+        .setPositiveButton(R.string.go_to_settings) { _, _ ->
+          // Intent to open the VPN settings
+          val intent = Intent(Settings.ACTION_VPN_SETTINGS)
+          startActivity(intent)
+        }
+        .setNegativeButton(R.string.cancel, null)
+        .show()
+  }
+
+  fun isAnotherVpnActive(context: Context): Boolean {
```
agottardo, line 309: Future: Is there a way to move this logic out of the main activity to prevent it from growing infinitely? On iOS, we have a dedicated, centralized class called ConflictDetector to perform this kind of system environment checks, instead of adding logic in different places.

agottardo (approved):

android/src/main/java/com/tailscale/ipn/MainActivity.kt
kari-ts: agreed this would be useful, but unfortunately I don't think there is a good way to do this

PR #478, opened 2024-08-22
android: Fix search not filtering machines from input

joshvocal:
Resolves: https://github.com/tailscale/tailscale/issues/13218

* Typing into the textfield to filter machines properly works now
  * `searchTerm.collect {}` was unreachable and caused it not to collect input

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/20eea88c-0458-409c-b114-17517bb1c9d7" /> | <video src="https://github.com/user-attachments/assets/c74c8f1c-6454-4247-851a-6869719d84c6" /> |

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
@@ -23,9 +23,11 @@ import com.tailscale.ipn.ui.util.PeerCategorizer
 import com.tailscale.ipn.ui.util.PeerSet
 import com.tailscale.ipn.ui.util.TimeUtil
 import com.tailscale.ipn.ui.util.set
+import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.launch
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -132,10 +143,6 @@ class MainViewModel(private val vpnViewModel: VpnViewModel) : IpnViewModel() {
         }
       }
 
-      viewModelScope.launch {
-        searchTerm.collect { term -> peers.set(peerCategorizer.groupedAndFilteredPeers(term)) }
-      }
-
       viewModelScope.launch {
```
kari-ts, line 146: I believe the bug was introduced when a } was deleted and this block and the above block launching the searchTerm.collect became unreachable. Could you also move the `viewModelScope.launch {
      App.get().healthNotifier?.currentIcon?.collect { icon -> healthIcon.set(icon) }
    }` as well please?

kari-ts (changes requested):
Thanks for finding the bug and creating a fix for it! We really appreciate it!

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
joshvocal: I've made those changes @kari-ts

kari-ts (approved):
Thanks for the fix!

PR #479, opened 2024-08-22
android: Add Voicemail apps to Android Split Tunneling settings by de…

joshvocal:
Resolves: https://github.com/tailscale/tailscale/issues/13199

Adds this list of voicemail apps to the default list of split tunneling apps:
* com.att.mobile.android.vvm
* com.tmobile.vvm.application
* com.metropcs.service.vvm
* com.mizmowireless.vvm
* com.vna.service.vvm
* com.dish.vvm
* com.comcast.modesto.vvm.client

Explored using a regex pattern for carriers, but it looks like the `VPNServiceBuilder` [needs the full package name](https://developer.android.com/reference/android/net/VpnService.Builder#addDisallowedApplication(java.lang.String)) in [this function](https://github.com/tailscale/tailscale-android/blob/main/android/src/main/java/com/tailscale/ipn/IPNService.kt#L99-L105).


agottardo (approved):
Thank you. This is quite US-centric, but we can add more entries in the future.

PR #480, opened 2024-08-26
android: stop tailscaled when VPN has been revoked

kari-ts:
-add new Ipn UI state 'Stopping' to handle the case where the VPN is no longer active and a request to stop Tailscale has been issued (but is not complete yet) and use for optimistic UI 
-when VPN has been revoked, stop tailscaled and set the state to Stopping 
-this fixes the race condition where when we tell tailscaled to stop, stopping races against the netmap state updating as a result of the VPN being revoked 
-add isActive state and use instead of isPrepared for UI showing whether we are connected - we were previously using isPrepared as a proxy for connection, but sometimes the VPN has been prepared but is not active (eg when VPN permissions have been given and VPN has been connected previously, but has been revoked) 
-refactor network callbacks into its own class for readability

Fixes tailscale/tailscale#12850

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -12,10 +12,6 @@ import android.content.Intent
 import android.content.SharedPreferences
 import android.content.pm.PackageManager
 import android.net.ConnectivityManager
-import android.net.LinkProperties
-import android.net.Network
-import android.net.NetworkCapabilities
-import android.net.NetworkRequest
 import android.os.Build
 import android.os.Environment
 import android.util.Log
@@ -45,7 +41,6 @@ import kotlinx.serialization.json.Json
 import libtailscale.Libtailscale
 import java.io.File
 import java.io.IOException
-import java.net.InetAddress
 import java.net.NetworkInterface
 import java.security.GeneralSecurityException
 import java.util.Locale
@@ -56,11 +51,6 @@ class App : UninitializedApp(), libtailscale.AppContext, ViewModelStoreOwner {
   companion object {
     private const val FILE_CHANNEL_ID = "tailscale-files"
     private const val TAG = "App"
-    private val networkConnectivityRequest =
-        NetworkRequest.Builder()
-            .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
-            .addCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN)
-            .build()
     private lateinit var appInstance: App
 
     /**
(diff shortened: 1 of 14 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/VpnViewModel.kt
```diff
@@ -29,6 +29,8 @@ class VpnViewModel(application: Application) : AndroidViewModel(application) {
   // Whether the VPN is prepared
```
agottardo, line 29: Can we provide comments to explain the difference between `vpnPrepared` and `vpnActive`, and their purpose? Who sets these, and who reads them?

libtailscale/net.go
```diff
@@ -118,7 +127,7 @@ var googleDNSServers = []netip.Addr{
 
 func (b *backend) updateTUN(service IPNService, rcfg *router.Config, dcfg *dns.OSConfig) error {
 	if reflect.DeepEqual(rcfg, b.lastCfg) && reflect.DeepEqual(dcfg, b.lastDNSCfg) {
-		b.logger.Logf("updateTUN: no change to Routes or DNS, ignore")
+		fmt.Printf("updateTUN: no change to Routes or DNS, ignore")
```
agottardo, line 130: did you mean to drop the logger?

libtailscale/net.go
```diff
@@ -193,10 +203,15 @@ func (b *backend) updateTUN(service IPNService, rcfg *router.Config, dcfg *dns.O
...
 	if err != nil {
 		if strings.Contains(err.Error(), "INTERACT_ACROSS_USERS") {
+			// Update VPN status if VPN interface cannot be created
+			b.logger.Logf("updateTUN: could not establish VPN because %v", err)
+			vpnService.service.UpdateVpnStatus(false)
 			return errMultipleUsers
 		}
 		return fmt.Errorf("VpnService.Builder.establish: %v", err)
 	}
+	log.Printf("KARI, setting vpn status to true")
```
agottardo, line 213: debugging leftover

agottardo: no harm in leaving it, just drop the Kari part

agottardo (changes requested):
Some nits

libtailscale/net.go
kari-ts: reverting change!

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -3,6 +3,7 @@
 
 package com.tailscale.ipn.ui.view
 
+import android.util.Log
```
agottardo, line 6: unused import?

android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -550,6 +509,10 @@ open class UninitializedApp : Application() {
     return builtInDisallowedPackageNames + userDisallowed
   }
 
+  fun getAppsScopedViewModel(): VpnViewModel {
```
agottardo, line 512:
```suggestion
  fun getAppScopedViewModel(): VpnViewModel {
```

agottardo (approved):

PR #482, opened 2024-08-28
android: fix Hostname type in MaskedPrefs, it should be String type

yinkaisheng:
set Hostname's type to String


agottardo:
@yinkaisheng can you please fix the DCO check? Thanks for the contribution.

agottardo (approved):

PR #483, opened 2024-08-28
android: make clipboard values clickable and focusable

kari-ts:
also, use Column isntead of LazyColumn since the Tailnet lock view is a short list and doesn't require lazy rendering

Fixes tailscale/corp#21737

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
@@ -42,89 +42,96 @@ fun TailnetLockSetupView(
     backToSettings: BackNavigation,
     model: TailnetLockSetupViewModel = viewModel(factory = TailnetLockSetupViewModelFactory())
 ) {
-  val statusItems by model.statusItems.collectAsState()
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
```diff
@@ -49,40 +54,44 @@ fun TailnetLockSetupView(
...
+                }
+
+            // Tailnet lock key
+            Lists.SectionDivider()
+            ClipboardValueView(
+                value = tailnetLockTlPubKey,
+                title = stringResource(R.string.tailnet_lock_key),
+                subtitle = stringResource(R.string.tailnet_lock_key_explainer))
+
+            // Optionally add more content here
```
agottardo, line 93: ?

agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/view/TailnetLockSetupView.kt
kari-ts: leftover from debugging, thank you!

PR #485, opened 2024-08-29
android: bump OSS to 1.73.73, use Go 1.23

agottardo:
OSS and Version updated to 1.73.73


kari-ts:
I think we need to also update the Makefile and Dockerfiles

agottardo:
> I think we need to also update the Makefile and Dockerfiles

There's no need to. The Go version in the Dockerfile is only used to bootstrap. We use the Go version used in the main OSS repo.

kari-ts:
> > I think we need to also update the Makefile and Dockerfiles
> 
> There's no need to. The Go version in the Dockerfile is only used to bootstrap. We use the Go version used in the main OSS repo.

don't you need to update `go mod tidy -compat=1.22` in `update-oss` in the Makefile?

kari-ts (approved):

PR #493, opened 2024-09-11
android: fix network callback race

kari-ts:
ConnectivityManager doesn't make guarantees about the order of network updates. Only use network updates for currently active network.


android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -28,29 +30,46 @@ object NetworkChangeCallback {
         networkConnectivityRequest,
         object : ConnectivityManager.NetworkCallback() {
           override fun onAvailable(network: Network) {
-            super.onAvailable(network)
+            // Only use if the network in the callback is the current active network. Network
+            // updates are not guaranteed to come in order.
+            if (connectivityManager.getActiveNetwork() == network) {
+              activeNetwork = network
```
andrew-d, line 36: Do we need to synchronize access to this field?

andrew-d (commented):
I wasn't able to find any information about whether at `NetworkCallback` can have methods called synchronously, so I think we might need to protect `activeNetwork` with something?

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -28,29 +30,46 @@ object NetworkChangeCallback {
...
+              if (dns.updateDNSFromNetwork(sb.toString())) {
+                Libtailscale.onDNSConfigChanged(linkProperties?.interfaceName)
+              }
+            } else {
 
-            if (dns.updateDNSFromNetwork(sb.toString())) {
-              Libtailscale.onDNSConfigChanged(linkProperties?.interfaceName)
+              Log.d(
+                  TAG,
+                  "monitor DNS changes, not sending DNS config because available network does not match currently active network")
```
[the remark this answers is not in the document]
agottardo: Yes, please make sure we always log the name of the network we received an update for, and the network that we currently believe to be active.

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,55 +2,85 @@
...
 
+  // Track the current active network
+  @Volatile private var activeNetwork: Network? = null
+  private val TAG = "NetworkChangeCallback"
+
+  // Create a lock object for synchronizing access to activeNetwork
+  private val activeNetworkLock = Any()
+
   // requestNetwork attempts to find the best network that matches the passed NetworkRequest. It is
   // possible that this might return an unusuable network, eg a captive portal.
```
andrew-d, line 23: This comment is out-of-date

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,55 +2,85 @@
...
-            .build()
-
-    connectivityManager.registerNetworkCallback(
-        networkConnectivityRequest,
+
+    connectivityManager.registerDefaultNetworkCallback(
         object : ConnectivityManager.NetworkCallback() {
           override fun onAvailable(network: Network) {
-            super.onAvailable(network)
+            val networkCapabilities = connectivityManager.getNetworkCapabilities(network)
```
andrew-d, line 29: I don't think this is safe; from [the docs](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onAvailable(android.net.Network)):

> Do NOT call ConnectivityManager.getNetworkCapabilities(android.net.Network) or ConnectivityManager.getLinkProperties(android.net.Network) or other synchronous ConnectivityManager methods in this callback as this is prone to race conditions (there is no guarantee the objects returned by these methods will be current). Instead, wait for a call to onCapabilitiesChanged(android.net.Network, android.net.NetworkCapabilities) and onLinkPropertiesChanged(android.net.Network, android.net.LinkProperties) whose arguments are guaranteed to be well-ordered with respect to other callbacks.

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,55 +2,85 @@
...
+                  val searchDomains: String? = linkProperties?.domains
+                  if (searchDomains != null) {
+                    sb.append("\n")
+                    sb.append(searchDomains)
+                  }
+
+                  if (dns.updateDNSFromNetwork(sb.toString())) {
+                    Libtailscale.onDNSConfigChanged(linkProperties?.interfaceName)
+                  } else {
+                    Log.d(TAG, "DNS config not updated")
```
andrew-d, line 57:
```suggestion
                    Log.d(TAG, "DNS config not updated because active network $connectivityManager.activeNetwork.toString() does not match network $network")
```

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,55 +2,85 @@
...
           override fun onAvailable(network: Network) {
-            super.onAvailable(network)
+            val networkCapabilities = connectivityManager.getNetworkCapabilities(network)
+            if (networkCapabilities != null &&
+                networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN) &&
+                networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)) {
+              // Only use if the network in the callback is the current active network. Network
+              // updates are not guaranteed to come in order.
+              synchronized(activeNetworkLock) {
+                if (connectivityManager.activeNetwork == network) {
```
andrew-d, line 36: Is this required now that we're only getting notified about the system's default network? The docs aren't super clear I think...

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,55 +2,85 @@
...
-            if (dns.updateDNSFromNetwork("")) {
-              Libtailscale.onDNSConfigChanged("")
+            // Proceed only if the network being lost is the most recently active network
+            synchronized(activeNetworkLock) {
+              if (network == activeNetwork) {
+                super.onLost(network)
+                if (dns.updateDNSFromNetwork("")) {
+                  Libtailscale.onDNSConfigChanged("")
+                } else {
+                  Log.d(TAG, "DNS config not updated")
```
andrew-d, line 77: (same log changes as above)

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
kari-ts: thanks for calling this out!!!

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
kari-ts: iiuc it is not necessary

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,48 +2,49 @@
...
-              sb.append(searchDomains)
+          override fun onCapabilitiesChanged(network: Network, capabilities: NetworkCapabilities) {
+            super.onCapabilitiesChanged(network, capabilities)
+            networkCapabilitiesCache[network] = capabilities
+            val linkProperties = linkPropertiesCache[network]
+            if (linkProperties != null &&
+                capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN) &&
+                capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)) {
+              maybeUpdateDNSConfig(linkProperties, dns)
             }
```
andrew-d, line 37: Could you add a log message here, stating why we're not updating the DNS config?

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -2,48 +2,49 @@
...
-              Libtailscale.onDNSConfigChanged(linkProperties?.interfaceName)
+          override fun onLinkPropertiesChanged(network: Network, linkProperties: LinkProperties) {
+            super.onLinkPropertiesChanged(network, linkProperties)
+            linkPropertiesCache[network] = linkProperties
+            val capabilities = networkCapabilitiesCache[network]
+            if (capabilities != null &&
+                capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN) &&
+                capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)) {
+              maybeUpdateDNSConfig(linkProperties, dns)
             }
```
andrew-d, line 48: And here?

andrew-d (approved):
I _think_ this is right, though we should be careful to test this on a few devices/networks IMO. Two requests for log messages, but otherwise LGTM!

PR #495, opened 2024-09-11
android: fix BuildConfig infinite loop

kari-ts:
Rather than create a Go struct that is set by Android, have Go call into Android to fetch build BuildConfig Updates tailscale/tailscale#13431


andrew-d (approved):
Thanks! We can always make this generic again later 😃

PR #498, opened 2024-09-12
android: rework NetworkChangeCallback to track all networks

andrew-d:
Instead of just tracking our default network, track all of them and decide upon each change which is the "best" option.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
+++ b/android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
@@ -6,94 +6,162 @@ import android.net.ConnectivityManager
 import android.net.LinkProperties
 import android.net.Network
 import android.net.NetworkCapabilities
+import android.net.NetworkRequest
 import android.util.Log
 import libtailscale.Libtailscale
 import java.net.InetAddress
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 
 object NetworkChangeCallback {
 
   private const val TAG = "NetworkChangeCallback"
 
-  // Cache LinkProperties and NetworkCapabilities since synchronous ConnectivityManager calls are
-  // prone to races.
-  // Since there is no guarantee for which update might come first, maybe update DNS configs on
-  // both.
-  val networkCapabilitiesCache = mutableMapOf<Network, NetworkCapabilities>()
-  val linkPropertiesCache = mutableMapOf<Network, LinkProperties>()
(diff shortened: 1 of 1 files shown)


agottardo (approved):

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -6,94 +6,159 @@ import android.net.ConnectivityManager
...
+    // default network used by Tailscale will always show up with capability
+    // NOT_VPN=false, and we must filter out NOT_VPN networks to avoid routing
+    // loops.
+    connectivityManager.registerNetworkCallback(
+        networkConnectivityRequest,
         object : ConnectivityManager.NetworkCallback() {
+          override fun onAvailable(network: Network) {
+            super.onAvailable(network)
+
+            Log.d(TAG, "onAvailable: network ${network.toString()}")
```
kari-ts, line 51: I learned that you don't need to explicitly call .toString() in string interpolation, you can just reference the object directly and Kotlin automatically calls .toString() for you

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -6,94 +6,159 @@ import android.net.ConnectivityManager
...
+              activeNetworks.remove(network)
+              maybeUpdateDNSConfigLocked("onLost", dns)
             }
           }
         })
   }
 
-  fun maybeUpdateDNSConfig(linkProperties: LinkProperties, dns: DnsConfig) {
+  // pickNonMetered returns the first non-metered network in the list of
+  // networks, or the first network if none are non-metered.
```
kari-ts, line 87: Can there be a situation where there are multiple non-metered networks, so picking the first one might result in the wrong one being selected?

kari-ts: Sorry, we discussed this earlier. Resolving.

kari-ts: It is possible, but should be ok to just pick one.

kari-ts (approved):

android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
andrew-d: Huh, TIL; thanks!

PR #502, opened 2024-09-13
android: include hex in LoginQRView

kari-ts:
Screenshot 2024-09-13 at 2 24 16 PM

Updates tailscale/tailscale#13277

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/LoginQRViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/LoginQRViewModel.kt
@@ -20,12 +20,28 @@ import kotlinx.coroutines.launch
 
 class LoginQRViewModel : IpnViewModel() {
 
+  val numCode: StateFlow<String?> = MutableStateFlow(null)
   val qrCode: StateFlow<ImageBitmap?> = MutableStateFlow(null)
 
   init {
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/LoginQRViewModel.kt
```diff
@@ -20,12 +20,28 @@ import kotlinx.coroutines.launch
...
   val qrCode: StateFlow<ImageBitmap?> = MutableStateFlow(null)
 
   init {
     viewModelScope.launch {
       Notifier.browseToURL.collect { url ->
-        url?.let { qrCode.set(generateQRCode(url, 200, 0)) } ?: run { qrCode.set(null) }
+        url?.let {
+          qrCode.set(generateQRCode(url, 200, 0))
+          // Extract the string after "https://login.tailscale.com/a/"
+          val prefix = "https://login.tailscale.com/a/"
```
agottardo, line 32: This is not going to work with a custom coordination server.

kari-ts: If there's a custom server, then we aren't showing the code

agottardo (approved):

PR #503, opened 2024-09-13
Makefile: disable GOTOOLCHAIN from dynamic switching

raggi:
Go has a new build facility that can utilize other toolchains if a module says so, but we manage the toolchain in our own way, so disable it.

Updates #501


bradfitz (approved):
This is fine, but I suspect more's also going on.

raggi:
> This is fine, but I suspect more's also going on.

yeah, agreed, I don't think this is involved, but I want to hard rule it out.

PR #511, opened 2024-09-16
android: only update DNS configs on LinkProperties changes

kari-ts:
We were updating DNS configs when capabilities changed, without LinkProperties having been filled in. Because onAvailable always happened first, LinkProperties were created with default value, and onCapabilitiesChanged sent a DNS update using those LinkProperties. This change only updates DNS configs on LinkProperties, which is the last update sent on a network change.


android/src/main/java/com/tailscale/ipn/NetworkChangeCallback.kt
```diff
@@ -9,34 +9,29 @@ import android.net.NetworkCapabilities
...
 
   private const val TAG = "NetworkChangeCallback"
 
-  private data class NetworkInfo(
-    var caps: NetworkCapabilities,
-    var linkProps: LinkProperties
-  )
+  private data class NetworkInfo(var caps: NetworkCapabilities, var linkProps: LinkProperties)
 
-  private val lock = ReentrantLock()
```
andrew-d, line 25: I think we still need this; upon re-reading the docs, it looks like a call to `onCapabilitiesChanged` will not happen concurrently with `onLinkPropertiesChanged`, but only for the same interface; we could still see concurrent calls for two different interfaces. IMO, it's cheap enough to keep this lock and might save us from a subtle bug.

andrew-d (commented):
Good catch on the ordering there! I'd still prefer to keep the lock, but only updating in `onLinkPropertiesChanged` looks right to me ✅

andrew-d (approved):
Thank you!

PR #512, opened 2024-09-17
android: bump OSS to 1.75.11

barnstar:
udpates #cleanup

OSS and Version updated to 1.75.11-t8b962f23d-gf07d419a125

The toolchain hash is being incorrectly by bumpOSS. Reverting it back to the correct value for 1.74/1.75


kari-ts:
Thanks for the fix!

kari-ts (approved):

PR #513, opened 2024-09-18
android: make custom url check case-insensitive

kari-ts:
Fixes tailscale/corp#23210

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/CustomLoginViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/CustomLoginViewModel.kt
@@ -36,7 +36,9 @@ class LoginWithCustomControlURLViewModel : CustomLoginViewModel() {
     // localAPIClient will use the default server if we give it a broken URL,
     // but we can make sure we can construct a URL from the input string and
     // ensure it has an http/https scheme
-    when (urlStr.startsWith("http") && urlStr.contains("://") && urlStr.length > 7) {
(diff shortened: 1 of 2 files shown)


agottardo (approved):

android/src/main/java/com/tailscale/ipn/ui/viewModel/CustomLoginViewModel.kt
```diff
@@ -36,7 +36,9 @@ class LoginWithCustomControlURLViewModel : CustomLoginViewModel() {
     // localAPIClient will use the default server if we give it a broken URL,
     // but we can make sure we can construct a URL from the input string and
     // ensure it has an http/https scheme
-    when (urlStr.startsWith("http") && urlStr.contains("://") && urlStr.length > 7) {
+    when (urlStr.startsWith("http", ignoreCase = true) &&
```
barnstar, line 39: Double checking - will Http:// work though?  (on all browsers?)

I probably copied this from Apple, where it does not.  If it doesn't, we could be nice and  just lowercase the scheme for them.

barnstar: Or if it's possible, modify the input field so it doesn't auto-capitalize

kari-ts: This worked for me. I like this suggestion, done!

barnstar (approved):

PR #515, opened 2024-09-19
android: send Android logs to logz

kari-ts:
ExtendedLog sends log messages to Android's logcat and Tailscale's logger Libtailscale wrapper is a Kotlin wrapper that allows us to get around the problems with mocking a native library

Fixes tailscale/corp#23191

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
@@ -19,6 +19,7 @@ import kotlinx.coroutines.launch
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
+import com.tailscale.ipn.util.TSLog
 
 // Notifier is a wrapper around the IPN Bus notifier.  It provides a way to watch
 // for changes in various parts of the Tailscale engine.  You will typically only use
@@ -59,7 +60,7 @@ object Notifier {
 
   @OptIn(ExperimentalSerializationApi::class)
   fun start(scope: CoroutineScope) {
-    Log.d(TAG, "Starting")
+    TSLog.d(TAG, "Starting Notifier")
     if (!::app.isInitialized) {
       App.get()
     }
@@ -89,7 +90,7 @@ object Notifier {
   }
 
   fun stop() {
-    Log.d(TAG, "Stopping")
+    TSLog.d(TAG, "Stopping Notifier")
     manager?.let {
       it.stop()
       manager = null
--- a/android/src/main/java/com/tailscale/ipn/util/TSLog.kt
+++ b/android/src/main/java/com/tailscale/ipn/util/TSLog.kt
@@ -1,0 +1,44 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+package com.tailscale.ipn.util
+
+import android.util.Log
+import libtailscale.Libtailscale
+
+object TSLog {
+  var libtailscaleWrapper = LibtailscaleWrapper()
+
+  fun d(tag: String?, message: String) {
+    Log.d(tag, message)
+    libtailscaleWrapper.sendLog(tag, message)
+  }
+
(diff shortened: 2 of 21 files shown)


barnstar (approved):

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -59,7 +60,7 @@ object Notifier {
 
   @OptIn(ExperimentalSerializationApi::class)
   fun start(scope: CoroutineScope) {
-    Log.d(TAG, "Starting")
+    TSLog.d(TAG, "Starting")
```
agottardo, line 63:
```suggestion
    TSLog.d(TAG, "Starting Notifier")
```

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -89,7 +90,7 @@ object Notifier {
   }
 
   fun stop() {
-    Log.d(TAG, "Stopping")
+    TSLog.d(TAG, "Stopping")
```
agottardo, line 93:
```suggestion
    TSLog.d(TAG, "Stopping Notifier")
```

android/src/main/java/com/tailscale/ipn/util/TSLog.kt
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+package com.tailscale.ipn.util
+
+import android.util.Log
+import libtailscale.Libtailscale
+
+object TSLog {
+  var libtailscaleWrapper = LibtailscaleWrapper()
```
agottardo, line 9: private?

android/src/main/java/com/tailscale/ipn/util/TSLog.kt
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+package com.tailscale.ipn.util
+
+import android.util.Log
+import libtailscale.Libtailscale
+
+object TSLog {
```
agottardo, line 8: Let's document why this exists and when it should be used in place of the util.Log.

agottardo (approved):

android/src/main/java/com/tailscale/ipn/util/TSLog.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  fun e(tag: String?, message: String, throwable: Throwable? = null) {
+    if (throwable == null) {
+      Log.e(tag, message)
+      libtailscaleWrapper.sendLog(tag, message)
+    } else {
+      Log.e(tag, message, throwable)
+      libtailscaleWrapper.sendLog(tag, "$message ${throwable?.localizedMessage}")
+    }
+  }
```
barnstar, line 36: In swift, we have the notion of debug logs (which only show up when you do local debug builds) and verbose logs, which sit behind a flag you can toggle at build or runtime.

android/src/main/java/com/tailscale/ipn/util/TSLog.kt
kari-ts: this can't be private because we use it in TimeUtilTest

android/src/main/java/com/tailscale/ipn/util/TSLog.kt
kari-ts: oh nice, this is a good idea. I think we can use Proguard to strip out logs for release builds, so maybe we could use that to strip verbose/debug logs? or we can specify in build flavors in our build.gradle. I'll do this in a follow up

libtailscale/interfaces.go
```diff
@@ -169,3 +171,13 @@ func RequestVPN(service IPNService) {
 func ServiceDisconnect(service IPNService) {
 	onDisconnect <- service
 }
+
+func SendLog(logstr string) {
```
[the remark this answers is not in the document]
kari-ts: very good idea. done.

PR #517, opened 2024-09-20
android: use PackageManager to determine install AppSourceChecker

kari-ts:
We were using MaybeGoogle to determine whether the app was installed from the Play Store, but this has not worked since the refactor. Fixes tailscale/tailscale#13442
Fixes tailscale/corp#23283


bradfitz:
Thanks!

Let's change `Fixes tailscale/corp#23283` to `Updates`, as this doesn't track F-Droid vs Amazon separately in Hostinfo.Package yet, so there's a bit more to do.

android/src/main/java/com/tailscale/ipn/AppSourceChecker.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.content.Context
+import android.os.Build
+import android.util.Log
+
+object AppSourceChecker {
+
+  const val TAG = "AppSourceChecker"
+
+  fun isInstalledFromGooglePlay(context: Context): Boolean {
```
agottardo, line 13: Could we cache the resulting value of this function in a variable somewhere, so we don't run this logic multiple times per app execution? Calls into PackageManager are slow, and the result is guaranteed to not change across calls here.

kari-ts: It's called once by runBackend

android/src/main/java/com/tailscale/ipn/AppSourceChecker.kt
```diff
@@ -0,0 +1,34 @@
...
+        packageManager.getInstallerPackageName(packageName)
+    }
+
+    Log.d(TAG, "Installer package name: $installerPackageName")
+
+    return when (installerPackageName) {
+        "com.android.vending" -> "googleplay"
+        "org.fdroid.fdroid" -> "fdroid"
+        "com.amazon.venezia" -> "amazon"
+        null -> "sideloaded"
```
bradfitz, line 30: how about "unknown" instead of "sideloaded"? IIUC, "null" is also used by Samsung or other Android variants that don't implement this.

So we don't know 100% that it's sideloaded.

android/src/main/java/com/tailscale/ipn/AppSourceChecker.kt
```diff
@@ -0,0 +1,34 @@
...
+    } else {
+        packageManager.getInstallerPackageName(packageName)
+    }
+
+    Log.d(TAG, "Installer package name: $installerPackageName")
+
+    return when (installerPackageName) {
+        "com.android.vending" -> "googleplay"
+        "org.fdroid.fdroid" -> "fdroid"
+        "com.amazon.venezia" -> "amazon"
```
bradfitz, line 29: nice!

bradfitz (approved):

android/src/main/java/com/tailscale/ipn/AppSourceChecker.kt
agottardo: 👍🏻

agottardo: I was concerned about the backend calling this repeatedly in the background every time hostinfo gets created, but this is good.

PR #518, opened 2024-09-20
Makefile: use "tailscale_go" build tag when using Tailscale's Go toolchain

bradfitz:
Updates tailscale/tailscale#13527

### DIFF
--- a/build-tags.sh
+++ b/build-tags.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+if [[ -z "$TOOLCHAIN_DIR" ]]; then
+    # By default, if TOOLCHAIN_DIR is unset, we assume we're
+    # using the Tailscale Go toolchain (github.com/tailscale/go)
+    # at the revision specified by go.toolchain.rev. If so,
+    # we tell our caller to use the "tailscale_go" build tag.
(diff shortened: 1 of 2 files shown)


raggi (approved):

build-tags.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [[ -z "$TOOLCHAIN_DIR" ]]; then
+    # By default, if TOOLCHAIN_DIR is unset, we assume we're
+    # using the Tailscale Go toolchain (github.com/tailscale/go)
+    # at the revision specified by go.toolchain.rev. If so,
+    # we tell our caller to use the "tailscale_go" build tag.
+    echo "tailscale_go"
+else
+    # Otherwise, if TOOLCHAIN_DIR is specified, we assume
+    # we're F-Droid or something using a stock Go toolchain.
```
agottardo, line 11: Does this mean we're allowing F-Droid to put out Tailscale APKs that will crash on older Android versions due to not using our toolchain?

bradfitz: We won't assert the toolchain version is correct without the tailscale_go build tag. So f-droid is unaffected.

PR #520, opened 2024-09-21
Login with custom server & preauth key

cmueller-tp:
Hi!
This is just a hacky PoC - works but not prod ready (I'm not a Kotlin dev).

As I see it there are 2 general options to accomplish this:

1.  adding another entry in the menu with another menu to enter both
2. unifying the menus with default URL and optional preauth key (kind of what I did but what I currently have does only work with both)

What do you think?

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/MainActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/MainActivity.kt
@@ -55,7 +55,6 @@ import com.tailscale.ipn.ui.view.ExitNodePicker
 import com.tailscale.ipn.ui.view.HealthView
 import com.tailscale.ipn.ui.view.IntroView
 import com.tailscale.ipn.ui.view.LoginQRView
-import com.tailscale.ipn.ui.view.LoginWithAuthKeyView
 import com.tailscale.ipn.ui.view.LoginWithCustomControlURLView
 import com.tailscale.ipn.ui.view.MDMSettingsDebugView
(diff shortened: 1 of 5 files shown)


agottardo:
Thank you for the contribution. We will discuss this with our UX design team to decide the correct location and flow for this feature.

cc @sonovawolf @kari-ts

cmueller-tp:
Let me know if the suggested flow is to your liking, I can do a quick implementation of the intended behavior, if that helps.

cmueller-tp:
What do you think about logging in via a token that consists of the preauth key and a server URI optionally encoded in a QR code? That would make logging in via smartphone very convenient and seamless for users and would also come in handy for onboarding servers with just a single token (b64) that needs to be copied over to the client.

farshadjanu1 (approved):

farshadjanu1 (approved):

farshadjanu1 (approved):

SoyDiego:
any news bout this PR @agottardo. Or any solution if I have custom server, because I want to login with tailscale and I don't have the option in android :(

Thanks

lsadehaan:
Same problem here...

lantrix:
Why is this not a default feature on all mobile clients 😮‍💨

robertofierimonte:
Hi @agottardo, not sure you're still looking at this, but if you're not, can you please point to someone else from the team who can review and bring this forward? 

This is much desired and needed functionality in the Android app.

PR #524, opened 2024-09-30
android: bumping OSS

kari-ts:
OSS and Version updated to 1.75.51-ta70287d32-gc10aca720b8


barnstar (approved):
Double check that toolchain rev against the repo - just to make sure it’s picking up the most recent commit.

kari-ts:
> Double check that toolchain rev against the repo - just to make sure it’s picking up the most recent commit.

yep! per https://github.com/tailscale/go/tags

PR #533, opened 2024-10-04
android: show locked out view

kari-ts:
MainViewModel listens for tailnet lock status, and MainView shows "admin signing required" messaging for devices that require signing. This also refactors ConnectView a bit for readability.

Fixes tailscale/corp#23657

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
@@ -14,6 +14,7 @@ import androidx.lifecycle.viewModelScope
 import com.tailscale.ipn.App
 import com.tailscale.ipn.R
 import com.tailscale.ipn.mdm.MDMSettings
+import com.tailscale.ipn.ui.localapi.Client
 import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.model.Ipn.State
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -152,6 +156,16 @@ class MainViewModel(private val vpnViewModel: VpnViewModel) : IpnViewModel() {
     viewModelScope.launch {
       App.get().healthNotifier?.currentIcon?.collect { icon -> healthIcon.set(icon) }
     }
+
+    Client(viewModelScope).tailnetLockStatus { result ->
```
agottardo, line 160: This code appears to be called whenever IpnViewModel is initialized. This will detect when the device is locked out upon launch. What happens once you sign the device and the app is already running? What will trigger the locked-out state to be removed upon signing?

On iOS, we have a timer running in the background. If you get locked-out, we start fetching the lock state from LocalAPI every 3 seconds. If you are not locked out anymore, we clear the locked-out state from the UI. Ugly, but there is no other way with the current API. See `HomeViewModel.swift`.

kari-ts: Closing this to instead send out a Notify when the locked out status changes.

PR #534, opened 2024-10-04
ui: hide commit hashes in user-facing version string

agottardo:
We currently show the full version number (with two commit hashes) everywhere. This pointlessly causes confusion for users, and is only really useful for Tailscale employees. Let's show the marketing version everywhere instead.

Users can still tap on the version number to copy the full version string. The extended version is also available in the Android settings, when inspecting Tailscale from the Apps list.


android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
```diff
@@ -71,7 +72,7 @@ fun AboutView(backToSettings: BackNavigation) {
                         Modifier.clickable {
                           localClipboardManager.setText(AnnotatedString(BuildConfig.VERSION_NAME))
```
kari-ts, line 73: can you add a comment here noting that the clipboard version should be the long form version and the displayed version is the short version please?

kari-ts (approved):

android/src/main/java/com/tailscale/ipn/ui/view/AboutView.kt
agottardo: done, good call

android/src/main/java/com/tailscale/ipn/ui/util/AppVersion.kt
```diff
@@ -0,0 +1,17 @@
+package com.tailscale.ipn.ui.util
```
kari-ts, line 1: // Copyright (c) Tailscale Inc & AUTHORS
// SPDX-License-Identifier: BSD-3-Clause

agottardo: We should find a way to make this a local check so it will prevent you from building. I always forget it 😅

kari-ts: if you will make a github issue for it, I'll create the gradle task checking for the heading :)

PR #539, opened 2024-10-09
android: hide disconnect action if force enabled

kari-ts:
In notification, don't show 'Disconnect' button if MDM force enable is on.

Fixes tailscale/corp#23764


agottardo (approved):
LGTM! Let's merge after 1.76 is cut.

barnstar (approved):

PR #542, opened 2024-10-15
android: bump OSS

kari-ts:
OSS and Version updated to 1.77.12-ta8f9c0d6e-g753b8d3fb4b


agottardo (approved):

bradfitz:
Btw, per http://go/commit we should use "bump" instead of "bumping"

PR #546, opened 2024-10-18
Makefile,*: use tailscale.com/cmd/mkversion

raggi:
We've suffered misalignment in versioning and toolchain usage due to the
shell invocations downstream of ./version/tailscale-version.sh, but also
the whole version data scheme in the Makefile was quite complicated, and
required synchronization in the build.grade.

- Makfile no longer needs to be version aware itself.
- A Makefile target tailscale.version refreshes a local cached output
  from tailscale.com/cmd/mkversion which is updated when go.mod / go.sum
  change.
- build.gradle loads tailscale.version to get the version string.
- ldflags are produced from tailscale.version via version-ldflags.sh

Updates https://github.com/tailscale/tailscale/issues/13850


.gitignore
```diff
@@ -38,3 +38,5 @@ tailscale.jks
 libtailscale.aar
 libtailscale-sources.jar
 .DS_Store
+
+tailscale.version
```
bradfitz, line 42: I thought we were going to check this file in to the repo? Or did you not `git add` it?

Makefile
```diff
@@ -141,6 +127,12 @@ tailscale-test.apk: gradle-dependencies
 	(cd android && ./gradlew assembleApplicationTestAndroidTest)
 	install -C ./android/build/outputs/apk/androidTest/applicationTest/android-applicationTest-androidTest.apk $@
 
+tailscale.version: go.mod go.sum
+    $(shell ./tool/go run tailscale.com/cmd/mkversion > tailscale.version)
+
```
bradfitz, line 132: oh, so we're still doing the full git clone pain, just hidden behind this tool

bradfitz (commented):
> Updates

You can use the F-Droid woes bug: https://github.com/tailscale/tailscale/issues/13850

Makefile
raggi: yeah, at least we now are down to one implementation of that behavior though!

Makefile
```diff
@@ -141,6 +127,12 @@ tailscale-test.apk: gradle-dependencies
 	(cd android && ./gradlew assembleApplicationTestAndroidTest)
 	install -C ./android/build/outputs/apk/androidTest/applicationTest/android-applicationTest-androidTest.apk $@
 
+tailscale.version: go.mod go.sum .git/HEAD
```
bradfitz, line 130: does having a git file as a dep here cause problems for people who download a release tarball from github without the `.git` directory?

Makefile
bradfitz: hah. lolsob.

android/build.gradle
```diff
@@ -38,7 +38,7 @@ android {
         minSdkVersion 26
         targetSdkVersion 34
         versionCode 241
-        versionName "1.77.12-ta8f9c0d6e-g753b8d3fb4b"
+        versionName getVersionProperty("VERSION_LONG")
```
bradfitz, line 41: ooooh hell yeah

android/build.gradle
```diff
@@ -181,3 +181,13 @@ def getLocalProperty(key, defaultValue) {
...
+
+
+def getVersionProperty(key) {
+    // tailscale.version is created / updated by the makefile, it is in a loosely
+    // Makfile/envfile format, which is also loosely a properties file format.
+    //      make tailscale.version
+    def versionProps = new Properties()
+    versionProps.load(project.file('../tailscale.version').newDataInputStream())
+    return versionProps.getProperty(key).replaceAll('^\"|\"$', '')
+}
```
bradfitz, line 193: groovy!

bradfitz (approved):

Makefile
andrew-d: I believe you can make this optional with `$(wildcard .git/HEAD)`, which will expand to the empty string if the file is not present. e.g.:
```
tailscale.version: go.mod go.sum $(wildcard .git/HEAD)
```

Makefile
```diff
@@ -141,6 +127,12 @@ tailscale-test.apk: gradle-dependencies
...
+    $(shell ./tool/go run tailscale.com/cmd/mkversion > tailscale.version)
+
+version: tailscale.version ## print the current version information
```
andrew-d, line 133: Slightly nitpick-y, but... make this a `PHONY` target?

PR #547, opened 2024-10-23
android: use native search

kari-ts:
-Add dynamic suggestions
-Use search bar with expanded view showing suggestions

Updates tailscale/corp#18973


oxtoacart:
Looking now ...

oxtoacart:
On an API level 30 emulated device, the layout jumps a bit when I select the Search bar.

[search_recording.webm](https://github.com/user-attachments/assets/eb432246-bfd9-4f73-afb7-6100f897dd7f)

oxtoacart:
Other than the above layout issue, it's working well for me.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -526,154 +534,117 @@ fun PeerList(
...
-                fontWeight = FontWeight.Light)
-          }
-        }
+    Spacer(modifier = Modifier.height(if (showNoResults) 0.dp else 8.dp))
 
-        var first = true
-        peerList.forEach { peerSet ->
-          if (!first) {
-            item(key = "user_divider_${peerSet.user?.ID ?: 0L}") { Lists.ItemDivider() }
+    // Peers display
```
barnstar, line 549: This isn't wrapped in the enableSearch flag... Now, this might actually work better on AndroidTV now!  but I tried it on a 720p simulator and you can't see the search results - there's just not enough vertical space.  I'm OK with just hiding the search bar on TV for now but ideally, we should make it work.  The hand-rolled search bar we were using before had several issues...

barnstar (changes requested):
AndroidTV is a bit borked on low res devices.  Everything else works great though.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
kari-ts: yeah, unfortunately the keyboard just takes up so much of the screen on 720p. I've readded that enableSearch check, I'll punt making the search bar work for TV for now.

kari-ts:
I used AnimatedVisibility to make the expansion animation a bit smoother. This keeps the initial search bar in place while expanding the content below it.

android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
kari-ts: I added the check back - the keyboard takes up so much space in 720p. Agreed we should have search on TV, but this might require a separate design for TV.

oxtoacart (approved):

agottardo (approved):

barnstar (commented):
I pulled and built the latest, and the list of nodes is not appearing.

kari-ts:
fixed! PTAL

barnstar (approved):

PR #548, opened 2024-10-24
android: don't show permissions for TV

kari-ts:
Android TV has limited support for notifications compared to mobile - notifications are not show in the system UI to provide a leanback experience. Remove 'Permissions' from Settings menu.

Fixes tailscale/corp/#21034


android/src/main/java/com/tailscale/ipn/ui/viewModel/LoginQRViewModel.kt
```diff
@@ -23,7 +23,7 @@ class LoginQRViewModel : IpnViewModel() {
   val numCode: StateFlow<String?> = MutableStateFlow(null)
   val qrCode: StateFlow<ImageBitmap?> = MutableStateFlow(null)
   // Remove this once changes to admin console allowing input code to be entered are made.
-  val codeEnabled = false
+  val codeEnabled = true
```
agottardo, line 26: debugging leftover?

kari-ts: oops, thanks for catching!

PR #550, opened 2024-10-30
android: fix MainActivityTest

kari-ts:
-Permissions are shown after 'Get Started' screen, fix ordering in test -Tap 'Authorize Tailscale'
-Re-add instrumentation test runner in build.gradle

Updates tailscale/corp#24242


oxtoacart (approved):
I confirmed that the test passes in my environment.

PR #551, opened 2024-10-31
android: use a coroutine for loadfiles

kari-ts:
contentResolver.query is attempting to perform a network query on the main thread. Move this to a coroutine to prevent blocking.

Fixes tailscale/corp#24293


agottardo (approved):
LGTM

PR #554, opened 2024-11-04
android: cherry pick fix for tailscale/corp#24293

kari-ts:


barnstar (changes requested):
This should be going to release/1.76 not main....

PR #556, opened 2024-11-04
android: bump OSS

kari-ts:
OSS and Version updated to 1.77.65-t698536947-gba306bf88


barnstar (changes requested):
This should be 1.76.5... Looking

PR #557, opened 2024-11-05
android: show hex code for TV log in

kari-ts:
Fixes tailscale/tailscale#13277


agottardo (approved):
Looks good!

Can you please add a descriptive commit message, so that a future reader will be able to know why this was the flag was dropped at this point in time? 🙏

PR #563, opened 2024-11-19
android: lazily init app in Client

kari-ts:
-Lazily init app in Client to ensure that we aren't trying to make any local API calls before app has been initialized. 
-Add @Volatile to ensure that isInitialized is always visible across threads and uses the updated value

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -90,6 +90,13 @@ class App : UninitializedApp(), libtailscale.AppContext, ViewModelStoreOwner {
     Log.d(s, s1)
   }
 
+  fun getLibtailscaleApp(): libtailscale.Application {
+    if (!isInitialized) {
+      initOnce() // Calls the synchronized initialization logic
+    }
+    return app
+  }
+
   override fun onCreate() {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/App.kt
```diff
@@ -119,15 +124,26 @@ class App : UninitializedApp(), libtailscale.AppContext, ViewModelStoreOwner {
...
+        initOnce() // Calls the synchronized initialization logic
+    }
+}
+
   @Synchronized
   private fun initOnce() {
     if (isInitialized) {
       return
     }
     isInitialized = true
```
barnstar, line 141: The method is synchronized so it's OK, but let's defer setting isInitialized to the end of the method since, technically, nothing isInitialzied at this point.

Does kotlin have a built in "run this block of code once and only once"  construct?  In Swift, we would do dispatch_once() here (or use a closure) instead of a volatile bool and the synchronization.

kari-ts: it doesn't afaik. I think the best we can do is the combination of the volatile isInitialized and the synchronization.

barnstar (approved):

PR #569, opened 2024-11-20
android: exclude Adaptive Connectivity Services

kari-ts:
Default to excluding Adaptive Connectivity Services to fix issue where it is erroneously classifying wifi as broken

Fixes tailscale/tailscale#14128


agottardo (approved):
An interesting one! Surprised that Google doesn't include this package by default if it causes issues with VPNs. 🤔

PR #580, opened 2024-12-05
android: fix android TV focus requester crash

kari-ts:
Properly attach FocusRequester to the root column of TaildropView so that there is a focusable UI element available to receive the focus

Fixes tailscale/corp#25007

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/SharedViews.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/SharedViews.kt
@@ -5,6 +5,7 @@ package com.tailscale.ipn.ui.view
 
 import androidx.annotation.StringRes
 import androidx.compose.foundation.clickable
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/SharedViews.kt
```diff
@@ -45,10 +46,14 @@ fun Header(
...
 ) {
-  val f = FocusRequester()
+  val focusRequester = remember { FocusRequester() }
 
   if (isAndroidTV()) {
-    LaunchedEffect(Unit) { f.requestFocus() }
+    LaunchedEffect(focusRequester) {
+      try {
+        focusRequester.requestFocus()
+      } catch (e: Exception) {}
```
agottardo, line 55: Log it perhaps?

agottardo (approved):

PR #585, opened 2024-12-09
android: synchronize ipn state and UI

kari-ts:
Pass in intended state to toggleVpn and keep track of progress to avoid redundant updates and ensure that the action taken reflects the user's intent. This fixes a possible recomposition loop caused by the ipn state and the vpn toggle state getting out of sync.


agottardo (approved):
LGTM. I suggest testing by exercising interactions with intents such as toggling via quick toggle / persistent notification.

PR #586, opened 2024-12-09
android: add logging to see what's spamming editPrefs

kari-ts:

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -190,6 +190,7 @@ class App : UninitializedApp(), libtailscale.AppContext, ViewModelStoreOwner {
     applicationScope.launch {
       val hideDisconnectAction = MDMSettings.forceEnabled.flow.first()
     }
+    TSLog.init(this)
   }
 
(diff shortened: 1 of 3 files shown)


agottardo (approved):
Can we add a check to ensure this doesn't accidentally ship in a stable release?

agottardo:
> > Can we add a check to ensure this doesn't accidentally ship in a stable release?
> 
> We're using debug logs now, so to differentiate, I made this a verbose log and then added a CI check to detect verbose logs. Is this along the lines of what you were thinking?

I simply meant something super-easy along the lines of 

```
if Version.IsUnstable {
   // Log.d(stacktrace)
}
```

where `Version.IsUnstable` looks at the current version number to return true/false. Alternatively, you can do what we do on Apple platforms, where our `TSLog.verbose` (the equivalent) checks for a specific UserDefaults flag to be enabled before printing anything to the syslog.

There's value in preventing `Log.v` calls from entering a release branch, but I don't think that CI check as-is will work because CI checks out a commit rather than a branch. But I'm not a CI expert. Feel free to merge everything as it was, I was just making a nit suggestion.

PR #587, opened 2024-12-09
android: put new search behind flag

kari-ts:
Until the WIP feature is ready

Updates tailscale/corp#18973

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
@@ -27,7 +27,10 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Search
 import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.Clear
+import androidx.compose.material.icons.outlined.Close
 import androidx.compose.material.icons.outlined.Lock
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -518,24 +525,70 @@ fun ConnectView(
...
       remember { derivedStateOf { searchTermStr.isNotEmpty() && peerList.isEmpty() } }.value
 
   val netmap = viewModel.netmap.collectAsState()
+  val focusManager = LocalFocusManager.current
+  var isSearchFocussed by remember { mutableStateOf(false) }
   var isListFocussed by remember { mutableStateOf(false) }
   val expandedPeer = viewModel.expandedMenuPeer.collectAsState()
   val localClipboardManager = LocalClipboardManager.current
   val enableSearch = !isAndroidTV()
+  val enableNewSearch = false // This enables new search page with suggestions.
```
barnstar, line 543: Let's add a featureFlags.kt file and class and have feature flag constants available to all parts of the code instead of embedding this where it's used (to match the other code bases)...   That'll come in handy in the future.

Something like (sorry for the pseudo-code, I'm in swift mode):
```
class FeatureFlags {
   static const enabled_full_page_search: bool = false
}

...
if (FeatureFlags.enable_full_page_search) {

}
```

barnstar (changes requested):

barnstar (approved):

PR #588, opened 2024-12-10
android: bump OSS

kari-ts:
OSS and Version updated to 1.77.166-t7f9ebc0a8-gebbc1b2f4


barnstar (approved):
Commit message is incorrect, please fix before merging.  Should be 1.79.something-something-something

PR #594, opened 2025-01-05
android: mark Quick Settings tile as toggleable

jarnedemeulemeester:
Mark the Quick Settings tile as toggleable.

The primary function of the Quick Settings tile is to turn on or of the VPN connection so it should be marked as toggleable as per the Android guidelines: https://developer.android.com/develop/ui/views/quicksettings-tiles#mark-tile

This also removes the little chevron on the tile which should only be shown on tiles that open a dialog or activity.


agottardo (approved):

agottardo:
Thank you!

PR #598, opened 2025-01-17
mdm: define Hostname syspolicy on Android

agottardo:
Updates tailscale/corp#25936

Adds a definition for the Hostname syspolicy which we are adding in OSS: https://github.com/tailscale/tailscale/pull/14676

Also adds it to the app_restrictions.xml file so that MDM solutions can pick it up.

### DIFF
--- a/android/src/main/res/values/strings.xml
+++ b/android/src/main/res/values/strings.xml
@@ -319,5 +319,7 @@
     <string name="run_as_subnet_router">Run as subnet router</string>
     <string name="use_tailscale_subnets_subtitle">Route traffic according to your network\'s rules. Some networks require this to access IP addresses that don\'t start with 100.x.y.z.</string>
     <string name="subnet_routing">Subnet routing</string>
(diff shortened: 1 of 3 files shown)


android/src/main/res/values/strings.xml
```diff
@@ -319,5 +319,7 @@
     <string name="run_as_subnet_router">Run as subnet router</string>
     <string name="use_tailscale_subnets_subtitle">Route traffic according to your network\'s rules. Some networks require this to access IP addresses that don\'t start with 100.x.y.z.</string>
     <string name="subnet_routing">Subnet routing</string>
+    <string name="specifies_a_device_name_to_be_used_instead_of_the_automatic_default">Specifies a device name to be used instead of the automatic default.</string>
+    <string name="hostname">Hostname</string>
```
nickkhyl, line 323: It_appears_we_use_extremely_long_string_names_for_other_policy_settings_as_well, so it's probably fine, but... 😭

nickkhyl (approved):

PR #612, opened 2025-02-05
android: bump OSS

nickkhyl:
OSS and Version updated to 1.81.24-te19c01f5b-gfee71b410

### DIFF
--- a/libtailscale/backend.go
+++ b/libtailscale/backend.go
@@ -20,6 +20,7 @@ import (
 	_ "tailscale.com/feature/condregister"
 	"tailscale.com/hostinfo"
 	"tailscale.com/ipn"
+	"tailscale.com/ipn/ipnauth"
 	"tailscale.com/ipn/ipnlocal"
 	"tailscale.com/ipn/localapi"
 	"tailscale.com/logtail"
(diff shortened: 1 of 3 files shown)


libtailscale/backend.go
```diff
@@ -134,7 +135,7 @@ func (a *App) runBackend(ctx context.Context) error {
 	a.backend = b.backend
 	defer b.CloseTUNs()
 
-	h := localapi.NewHandler(b.backend, log.Printf, *a.logIDPublicAtomic.Load())
+	h := localapi.NewHandler(ipnauth.Self, b.backend, log.Printf, *a.logIDPublicAtomic.Load())
```
nickkhyl, line 138: In tailscale/tailscale#14910, we updated `localapi.NewHandler` to include a new `ipnauth.Actor` parameter.
Currently, it is only used for the new backend-enforced `AlwaysOn.Enabled` / `AlwaysOn.OverrideWithReason` policy settings, which are not yet fully supported on Android, so we can use `ipnauth.Self` here.

We'll need to do something similar to tailscale/corp#26360 -- or better yet, have a unified oss implementation in `tailscale.com/ipn/ipnauth` -- when we need a proper `ipnauth.Actor` for Android.

kari-ts (approved):

PR #617, opened 2025-02-27
Makefile: update Docker image every three months

kari-ts:
Update Docker image every three months to force periodic rebuilds.

Fixes tailscale/corp#26816


barnstar (approved):

Makefile
```diff
@@ -9,8 +9,13 @@
...
 #
-# The convention here is tailscale-android-build-amd64-<date>
-DOCKER_IMAGE=tailscale-android-build-amd64-191124
+# The convention here is tailscale-android-build-amd64-<go.toolchain.rev>
+#
+# Read the current Go toolchain revision.
+GO_TOOLCHAIN_REV := $(shell cat go.toolchain.rev)
+
+# Define the Docker image tag using the current revision.
+DOCKER_IMAGE := tailscale-android-build-amd64-$(GO_TOOLCHAIN_REV)
```
barnstar, line 18: This actually won't work the way it is because we have no way of refreshing the builder when the android toolchain updates (which won't get refreshed automatically).  Can we append the android toolchain here as well? That should be sufficient.  Maybe toss in a constant as well, just in case we ever have to manually trigger a refresh for other reasons.

barnstar (changes requested):

PR #619, opened 2025-03-05
Makefile: make unstripped libtailscale and debug symbols

kari-ts:
- Added Makefile targets to build an unstripped version of the libtailscale AAR.
- Extracted the native .so (libgojni.so) from the unstripped AAR.
- Generated a debug symbols file (libgojni.so.debug) using llvm-objcopy with --only-keep-debug.
- Stripped the native library (libgojni.so.stripped) with --strip-debug and repackaged the final AAR.

This allows the build chain to produce a stripped AAR for the app and a separate debug symbols file that can be uploaded to Google Play for crash deobfuscation.

Updates tailscale/tailscale#15210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,13 @@ tailscale.jks
 .vscode
 .idea
 
+# Native libraries
+*.stripped
+*.unstripped
+
+# Debug symbols
+*.debug
+
 libtailscale.aar
 libtailscale-sources.jar
 .DS_Store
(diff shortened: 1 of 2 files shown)


barnstar (approved):
Looks good, but it looks like we're just swallowing the unstripped aar during the build right now?   The build pipeline is in corp.   Perhaps, there's a way to add the symbols to the upload step via the Playstore API.

PR #631, opened 2025-04-03
.github/workflows: pin GitHub action versions

mpminardi:
Pin versions of GitHub actions that are used in our workflows.

Bumps an instance of actions/checkout from v3.x to v4.x.

Bumps actions/setup-java from v3.x to v4.x.

Updates #cleanup

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -16,12 +16,12 @@ jobs:
 
     steps:
       - name: Check out code
-        uses: actions/checkout@v3
-      - uses: actions/setup-go@v5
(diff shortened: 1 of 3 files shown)


.github/workflows/android.yml
```diff
@@ -16,12 +16,12 @@ jobs:
 
     steps:
       - name: Check out code
-        uses: actions/checkout@v3
-      - uses: actions/setup-go@v5
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
mpminardi, line 19: We're using checkout `v4` elsewhere already, updated this to be in line with other usage.

zehauser (approved):

PR #636, opened 2025-04-14
docker: refactor and update docker image build

smikes:
See https://github.com/smikes/tailscale-android/tree/smikes/simplify-docker-expanded for step-by-step refactoring of dockerfile

update docker image datestamp
make docker-shell target use the build image with an explicit /bin/bash command retain the docker-remove-shell-image target in case something depends on it, but clarify that it is now a no-op update gradle version to 8.11.1
hoist all ENV declarations
hoist all mkdir commands
combine mkdir commands
use env vars in mkdir
combine apt-get commands into single RUN
make argument order consistent in apt-get commands collect packages into fewer apt-get commands
add clean up of temporary files after apt-get
expand apt-get commands to one package per line
sort packages in apt-get commands
combine go installation commands into single RUN
update golang version to 1.24.1
also copy build.gradle file
ensure gradlew is executable before running
remove trailing newline


docker/DockerFile.amd64-build
```diff
@@ -3,31 +3,51 @@
...
-RUN mkdir -p build
 ENV HOME /build
-
-# Make android sdk location, the later make step will populate it.
-RUN mkdir android-sdk
 ENV ANDROID_HOME $HOME/android-sdk
 ENV ANDROID_SDK_ROOT $ANDROID_HOME
 ENV PATH $PATH:$HOME/bin:$ANDROID_HOME/platform-tools
 
+RUN mkdir -p \
```
danderson, line 11: (mild suggestion, no requirement to act on it; can also do this in a followup change)

When Dockerfile RUN statements get long enough to need copious newline escaping, I tend to extract the build steps into a build.sh or similar, and compact the Dockerfile to copying that build script into the image and running it. That makes editors happier to edit the build script, and also makes it visible to tools like shellcheck (though I don't think we have it set up in this repo).

danderson (approved):

danderson:
Issuebot wants the issue mentioned in the git commit message, for posterity. The syntax it recognizes is `Updates tailscale/tailscale#15674`, or `Fixes tailscale/tailscale#15674` to close the bug on merge (though not sure if that works across repos, but I can fix it in post if necessary)

PR #637, opened 2025-04-14
dependencies: update crypto to fix CVE

smikes:
Fixes issue #15680

Fix https://github.com/golang/vulndb/issues/3328
Github https://github.com/advisories/GHSA-v778-237x-gjrc

NOTE: there is no known exploit of this code via tailscale and I do not believe tailscale-android even used the at-risk code path


raggi:
fixed by https://github.com/tailscale/tailscale-android/pull/639

PR #638, opened 2025-04-14
libtailscale: drop inet.af dependency, use net/netip

raggi:
Move to the stdlib version, inet.af is bitrotting as a domain and implementation.

Updates #cleanup


WalterHub (approved):

libtailscale/net.go
```diff
@@ -102,9 +101,12 @@ func (a *App) getInterfaces() ([]netmon.Interface, error) {
 
 		addrs := strings.Trim(fields[1], " \n")
 		for _, addr := range strings.Split(addrs, " ") {
-			ip, err := netaddr.ParseIPPrefix(addr)
+			ip, err := netip.ParsePrefix(addr)
```
jwhited, line 104: nit: We could also `net.ParseCIDR()` since the goal is to get a `*net.IPNet`.

jwhited (approved):

PR #648, opened 2025-05-12
mdm: define TailscaleOnboardingSeen syspolicy on Android

zbuchheit:
Adds an MDM setting `TailscaleOnboardingSeen` which allows for the intro screen to be skipped when set to true.

Updates https://github.com/tailscale/corp/issues/27966

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/MainActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/MainActivity.kt
@@ -292,7 +292,7 @@ class MainActivity : ComponentActivity() {
                 }
 
             // Show the intro screen one time
-            if (!introScreenViewed()) {
+            if (!introScreenViewed() && !mdmTailscaleOnboardingSeen()) {
               navController.navigate("intro")
               setIntroScreenViewed(true)
             }
@@ -455,6 +455,10 @@ class MainActivity : ComponentActivity() {
         .edit()
         .putBoolean("seen", seen)
         .apply()
+  }
+
+  private fun mdmTailscaleOnboardingSeen(): Boolean {
+    return MDMSettings.tailscaleOnboardingSeen.flow.value.value
   }
 }
 
--- a/android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
+++ b/android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
@@ -97,6 +97,9 @@ object MDMSettings {
   // Overrides the value provided by os.Hostname() in Go
   val hostname = StringMDMSetting("Hostname", "Device Hostname")
 
+  // Allows admins to skip the get started intro screen
+  val tailscaleOnboardingSeen = BooleanMDMSetting("TailscaleOnboardingSeen", "Suppress the intro screen")
+
   val allSettings by lazy {
     MDMSettings::class
         .declaredMemberProperties
--- a/android/src/main/res/values/strings.xml
+++ b/android/src/main/res/values/strings.xml
@@ -214,6 +214,8 @@
     <string name="run_as_exit_node_visibility">Run as exit node visibility</string>
     <string name="defines_an_auth_key_that_will_be_used_for_login">Defines an auth key that will be used for login.</string>
     <string name="auth_key">Auth Key</string>
+    <string name="skips_the_intro_page_shown_to_users_that_open_the_app_for_the_first_time">Skips the intro page shown to users that open the app for the first time</string>
+    <string name="tailscale_onboarding_seen">Mark Onboarding Page as Seen</string>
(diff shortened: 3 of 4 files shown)


android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
```diff
@@ -97,6 +97,9 @@ object MDMSettings {
   // Overrides the value provided by os.Hostname() in Go
   val hostname = StringMDMSetting("Hostname", "Device Hostname")
 
+  // Allows admins to skip the get started intro screen
+  val tailscaleOnboardingSeen = BooleanMDMSetting("TailscaleOnboardingSeen", "Suppress the intro screen")
```
nickkhyl, line 101: Can you please use `OnboardingFlow` instead of `TailscaleOnboardingSeen`?

See also: https://github.com/tailscale/tailscale/pull/13987#issuecomment-2452363333

nickkhyl: TLDR: We have `TailscaleOnboardingSeen` on macOS. It's not a great policy setting name, so we went with `OnboardingFlow` on Windows. We'll update the macOS client sometime, but let's stick with `OnboardingFlow` and `show`/`hide` as the values for any new platforms.

android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
```diff
@@ -97,6 +97,9 @@ object MDMSettings {
   // Overrides the value provided by os.Hostname() in Go
   val hostname = StringMDMSetting("Hostname", "Device Hostname")
 
+  // Allows admins to skip the get started intro screen
+  val onboardingFlow = ShowHideMDMSetting("OnboardingFlow", "Suppress the intro screen")
```
barnstar, line 101: There is an existing MDM key for this we use on other platforms which should be replicated here:

```
TailscaleOnboardingSeen
```

https://tailscale.com/kb/1315/mdm-keys#suppress-the-first-launch-onboarding-flow

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -291,8 +297,7 @@ class MainActivity : ComponentActivity() {
                   }
                 }
 
-            // Show the intro screen one time
-            if (!introScreenViewed()) {
+            if (onboardingFlowShowHide == ShowHide.Show && !introScreenViewed()) {
```
barnstar, line 300: Let's put all the logic into introScreenViewed().  Maybe reverse it to return true if we should show the onboarding ie: shouldDisplayOnboarding()...

android/src/main/res/values/strings.xml
```diff
@@ -214,6 +214,8 @@
     <string name="run_as_exit_node_visibility">Run as exit node visibility</string>
     <string name="defines_an_auth_key_that_will_be_used_for_login">Defines an auth key that will be used for login.</string>
     <string name="auth_key">Auth Key</string>
+    <string name="skips_the_intro_page_shown_to_users_that_open_the_app_for_the_first_time">Skips the intro page shown to users that open the app for the first time</string>
+    <string name="onboarding_flow">Set Onboarding Flow</string>
```
barnstar, line 218:
```
<string name="onboarding_flow">Skip the Onboarding Flow</string>
```

android/src/main/res/xml/app_restrictions.xml
```diff
@@ -134,4 +134,12 @@
         android:key="Hostname"
         android:restrictionType="string"
         android:title="@string/hostname" />
+
+    <restriction
+        android:description="@string/skips_the_intro_page_shown_to_users_that_open_the_app_for_the_first_time"
+        android:entries="@array/show_hide_labels"
+        android:entryValues="@array/show_hide"
+        android:key="OnboardingFlow"
```
barnstar, line 142:
```
android:key="TailscaleOnboardingSeen"
```

barnstar (changes requested):
Thank you @zbuchheit !

Just a few small changes to use the existing policy name for consistency with other platforms.

android/src/main/java/com/tailscale/ipn/mdm/MDMSettings.kt
zbuchheit: @barnstar we had some discussion here to change it to OnboardingFlow

zbuchheit:
@barnstar I made some of the changes you suggested. I am also happy to change it to TailscaleOnboardingSeen if that's what you all decide to finalize it as. @nickkhyl suggested changing it to OnboardingFlow.

PR #659, opened 2025-05-29
android: do not stop running on login, and edit prefs after startLogi…

kari-ts:
…nInteractive

Previously: start, edit prefs with wantRunning=false, then startLoginInteractive
Now: editPrefs, start, startLoginInteractive. Do not call wantRunning=false; the route clearing issue requiring that is resolved.

Updates tailscale/corp#24002

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
@@ -11,6 +11,7 @@ import com.tailscale.ipn.ui.localapi.Client
 import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.model.IpnLocal
 import com.tailscale.ipn.ui.model.UserID
+import com.tailscale.ipn.ui.model.deepCopy
 import com.tailscale.ipn.ui.notifier.Notifier
 import com.tailscale.ipn.ui.util.AdvertisedRoutesHelper
 import com.tailscale.ipn.ui.util.LoadingIndicator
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -144,52 +144,75 @@ open class IpnViewModel : ViewModel() {
 
   // Login/Logout
 
+  /**
+   * Order of operations:
+   * 1. start() -> boots tailscaled with UpdatePrefs + AuthKey
+   * 2. startLoginInteractive()
+   * 3. if AuthKey != null => WantRunning = true
```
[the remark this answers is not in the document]
kari-ts: Thanks for the explanation! Fixed to do this.

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -144,52 +144,60 @@ open class IpnViewModel : ViewModel() {
...
+          .onSuccess {
+            val opts = Ipn.Options(UpdatePrefs = editResult.getOrThrow(), AuthKey = authKey)
+            client.start(opts) { startResult ->
+              startResult
+                  .onFailure {
+                    TSLog.e(TAG, "start() failed: ${it.message}")
+                    completionHandler(Result.failure(it))
+                  }
+                  .onSuccess {
+                    if (authKey.isNullOrEmpty()) {
```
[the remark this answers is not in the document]
nickkhyl: (I haven't tested it recently, so I might be wrong. But if I am, please let me know so I can fix it)

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -144,52 +145,55 @@ open class IpnViewModel : ViewModel() {
 
   // Login/Logout
 
+  /**
+   * Order of operations:
+   * 1. editPrefs() with WantRunning=true, LoggedOut=false if AuthKey != null
```
nickkhyl, line 150: nit: maybe something like
```suggestion
   * 1. editPrefs() with maskedPrefs (to allow ControlURL override), WantRunning=true, LoggedOut=false if AuthKey != null
```

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -144,52 +145,55 @@ open class IpnViewModel : ViewModel() {
...
+   * 2. start() -> boots tailscaled; if an AuthKey was provided and WantRunning/LoggedOut were set
+   *    to true/false in step 1, backend will attempt to log in immediately.
```
nickkhyl, line 152: Sadly, this part turned out to be wrong at this time:
> if an AuthKey was provided and WantRunning/LoggedOut were set to true/false in step 1, backend will attempt to log in immediately.

Maybe remove and change to `start() starts the LocalBackend state machine.` or similar?

android/src/main/java/com/tailscale/ipn/ui/viewModel/IpnViewModel.kt
```diff
@@ -144,52 +145,55 @@ open class IpnViewModel : ViewModel() {
...
+   * 3. startLoginInteractive()
```
nickkhyl, line 153: Maybe mention that it's currently required for both interactive and non-interactive (with an auth key) logins?

nickkhyl (approved):
LGTM with the updated comment!

barnstar (approved):

PR #664, opened 2025-06-04
android: detect amazon fire stick as a AndroidTV

barnstar:
fixes tailscale/tailscale#16164

We weren't detecting fire stick devices as TV devices.


kari-ts (approved):
Agree with the bot about making the val private; otherwise LGTM

PR #665, opened 2025-06-04
android: use updated exit node detection logic in exit node categorizer

barnstar:
updates tailscale/corp#29045

The exit node sorting logic wasn't using node.isMullvadNode and still had a variant of the original issue (listing mullvad nodes as self-nodes).

This also switches the logic to ignore the name fields entirely and queue only off the Location field matching Windows and Apple.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
@@ -107,13 +107,9 @@ class Tailcfg {
     val isExitNode: Boolean =
         (AllowedIPs?.contains("0.0.0.0/0") ?: false) && (AllowedIPs?.contains("::/0") ?: false)
 
-    // mullvad nodes are exit nodes with a mullvad.ts.net domain *or* Location Info.
-    // These checks are intentionally redundant to avoid false negatives.
+    // mullvad nodes are exit nodes with non-nil Location info.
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
```diff
@@ -110,10 +110,7 @@ class Tailcfg {
     // mullvad nodes are exit nodes with a mullvad.ts.net domain *or* Location Info.
```
kari-ts, line 110: please update comment

kari-ts (approved):

android/src/main/java/com/tailscale/ipn/ui/model/TailCfg.kt
```diff
@@ -107,13 +107,9 @@ class Tailcfg {
     val isExitNode: Boolean =
         (AllowedIPs?.contains("0.0.0.0/0") ?: false) && (AllowedIPs?.contains("::/0") ?: false)
 
-    // mullvad nodes are exit nodes with a mullvad.ts.net domain *or* Location Info.
-    // These checks are intentionally redundant to avoid false negatives.
+    // mullvad nodes are exit nodes with non-nil Location info.
```
bradfitz, line 110: This isn't a future-proof heuristic. 

We plan to add Location fields to other types of nodes.

PR #667, opened 2025-06-09
android: touchless auth key login

barnstar:
updates tailscale/corp#29482

If an authKey is detected in the mdm payload, we will now skip the onboarding flows and several of the other non-mandatory permission prompts.


android/src/main/java/com/tailscale/ipn/ui/view/MainView.kt
```diff
@@ -215,13 +215,16 @@ fun MainView(
...
               Ipn.State.Running -> {
-
-                PromptPermissionsIfNecessary()
                 viewModel.maybeRequestVpnPermission()
                 LaunchVpnPermissionIfNeeded(viewModel)
-                LaunchedEffect(state) {
-                  if (state == Ipn.State.Running && !isAndroidTV()) {
-                    viewModel.checkIfTaildropDirectorySelected()
+
+                if (!viewModel.skipPermissionsPrompts()) {
```
kari-ts, line 221: can you bake the check into `PromptPermissionsIfNecessary`, or rename it so it doesn't imply it's doing this check?

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -219,6 +225,10 @@ class MainViewModel(private val vpnViewModel: VpnViewModel) : IpnViewModel() {
   }
 
   fun checkIfTaildropDirectorySelected() {
```
kari-ts, line 227: what is the right taildrop behavior in the touchless case? select no directory because they will never use taildrop, or select one for them?

PR #690, opened 2025-08-12
android: Taildrop/share plain text (i.e. clipboard)

### LINKED ISSUE #16850: Taildrop sharing should share text if stream is unavailable
### What is the issue?

Not sure whether to call this a bug report or a feature request. Related to #6057 and my request [here](https://androiddev.social/@dgmltn/115012145641445538) ... currently when trying to Taildrop text (e.g. from the clipboard), the request is accepted but ignored.

Image

### Steps to reproduce

1. Android: Copy text to the clipboard
2. Click share button
3. Click Tailscale
4. "No files to share"

_No response_

### OS

Android

### OS version

Android 16

### Tailscale version

1.86.4

### Other software

_No response_

### Bug report

_No response_

dgmltn:
Fixes tailscale/tailscale#16850

This enables the ability to share text (Taildrop) from Android, via the clipboard or any other text sharing intent. The relevant intent is already handled by the app, but plain text is just ignored.

This is a partial solution to [my request here](https://androiddev.social/@dgmltn/115012145641445538)

image

An open question: what do you think of the generated filename `shared_text_[epoch].txt`? I see the existing fallback filename when the filename is unknown is simply `[epoch].txt` but it seemed appropriate to add a bit of context when it's known that the text is coming from a share intent instead of directly from a file on disk.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -131,6 +131,16 @@
                 android:value="true" />
         </service>
 
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="com.tailscale.ipn.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/file_paths" />
+        </provider>
+
         <meta-data
             android:name="android.content.APP_RESTRICTIONS"
             android:resource="@xml/app_restrictions" />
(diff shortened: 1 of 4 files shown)


harryVGY:
any comment about the project closing? Its a very sought-after feature from what i understand. Or am I just a newbie about how projects works and its done and to be pushed to main bransch?

dgmltn:
> any comment about the project closing? Its a very sought-after feature from what i understand. Or am I just a newbie about how projects works and its done and to be pushed to main bransch?

I closed because this was sitting on a branch for 10mo with no comments. I'd still like the feature too, but blip also does the same thing.

PR #694, opened 2025-08-29
android,libtailscale: implement key.HardwareAttestationKey

awly:
Use a KeyStore-backed key to store a hardware-bound private key.

Updates https://github.com/tailscale/tailscale/issues/15830


dblohm7 (changes requested):
I'll defer to @patrickod for the rest, but I do have an ask:

dblohm7 (approved):

PR #695, opened 2025-08-31
feat: support disabling remote log uploads

mnahkies:
relates: https://github.com/tailscale/tailscale/issues/13174

- adds a new switch to the settings page for enabling/disabling remote log uploads

- calls the `Disable` function from the `logtail` package during init when the setting is turned off

ref: https://pkg.go.dev/tailscale.com/logtail#Disable

<details><summary>Expand for Screenshot</summary>
<p>

image

</p>
</details>


android/src/main/res/values/strings.xml
```diff
@@ -346,6 +346,8 @@
     <string name="run_as_subnet_router">Run as subnet router</string>
     <string name="use_tailscale_subnets_subtitle">Route traffic according to your network\'s rules. Some networks require this to access IP addresses that don\'t start with 100.x.y.z.</string>
     <string name="subnet_routing">Subnet routing</string>
+    <string name="client_remote_logging_enabled">Remote client logging</string>
+    <string name="client_remote_logging_enabled_subtitle">Equivalent to --no-logs-no-support on Linux.\nChanges require restarting the app to take effect.</string>
```
bradfitz, line 350: This is pretty nerdy. We don't need to reference the flag name or Linux here and can just explain the implications:

No debug logs, no support, prevents using Network Flow logs, etc. 

Like "Whether debug logs are uploaded to Tailscale support."

It's admittedly hard to fit all the nuance in a couple lines of text.

mnahkies: It's a bit wordy, but perhaps something like this?
image

Couple other thoughts:
- Should the "Bug report" button be hidden when logs are disabled?
- Is this fine as a top-level settings item, or would it be better under "Permissions"?
- How does / should this interact with MDM settings? (I fear I might be opening a can of worms with this one, as it might require backend changes to closed source components? :sweat_smile:)

mnahkies: Another thought on MDM: assuming I'm correct that this would need to be controllable through MDM (to enforce flow logs are produced for enterprise) - is there a stepping stone where we could force this setting on when any MDM is employed, and then later make it an individually configurable MDM setting?

mnahkies: I've taken a stab at implementing what I think respecting MDM _could_ look like in https://github.com/tailscale/tailscale-android/pull/695/commits/04fd66c55f7e29b10ebab1b64890adf7356cfb2e (has an outstanding todo to handle async config updates correctly, and I haven't been able to enroll a device to test e2e properly yet)

TBH for my purposes, I can probably just run a custom build with this patch applied, though I do think it would be nice to land in `main` - particularly in the headscale use-case, sending logs to tailscale doesn't make a lot of sense to me, which as far as I can tell would currently occur.

generic-internet-user:
any progress on this?

fartoonice:
Eagerly awaiting this implementation so I can finally start using what I otherwise deem to be a pretty cool project.

Mynacol:
Hey, we are again two months in without any reaction. Can someone from Tailscale bring this forward? @Erisa maybe?

generic-internet-user:
I'd be interested in seeing an official response to this as well by now

bradfitz:
Looks good to us.

I rebased this locally and fixed some of the English up and it looks like the MDM check is backwards, so fixing + verifying that too. But first I sent out https://github.com/tailscale/tailscale/pull/19446 to remove the "you must restart the process" thing which is awkward.

I'll update this PR and merge it once that other change is in and I've verified it works.

bradfitz (approved):

PR #697, opened 2025-08-31
chore: reformat code

mnahkies:
stacked on https://github.com/tailscale/tailscale-android/pull/696 this change simply runs the new `make lint` command to reformat the code.

I wasn't sure how annoying to review this would be, and so figured there is a decent chance a maintainer might prefer to run themselves (essentially replacing this PR).


mnahkies:
> It should probably also have CI coverage to prevent regressions in the future.
> 
> **Edit**: I should've read #696 first :)

Given the diff isn't as bad to review as I'd guessed it might be, I'd be happy to fold this into #696 now that it's been demonstrated to do its job in CI correctly.

PR #706, opened 2025-09-26
libtailscale: fix regression in interface address enumeration

raggi:
Fix regression introduced in 9c933a08a2ce531ed56dd3008de3cbc1b14b6d22.

Fixes tailscale/tailscale#16836


sfllaw (approved):
Approved.

I have one suggestion to simplify things.

I was also going to ask for a test case to prevent future breakage, but then I realized that there aren’t any Go tests in this repo?!?

PR #735, opened 2026-02-06
Fix missing Allow LAN access toggle 

### LINKED ISSUE #16187: Android 11 Allow LAN access toggle missing when using exit node
### What is the issue?

I connected to exit node, but I can't enable LAN access, the option is just missing on my phone, but visible on other android devices.

Image

### Steps to reproduce

I have no idea. I tried clean install of the latest 1.84.1 from [https://pkgs.tailscale.com/stable/tailscale-android-universal-1.84.1.apk](https://pkgs.tailscale.com/stable/tailscale-android-universal-1.84.1.apk) and the issue persists.

It is an issue from a couple of releases back, I can't pin point which exact version broke it.

### OS

Android

### OS version

11

### Tailscale version

1.84.1

### Other software

None

### Bug report

BUG-600dc93769eb12dfc00e1ef195f43919ec6af4e7990cc75e7eebc74c6ebba82a-20250604104030Z-0d2cbbba2a591042

Pawloland:
Fix missing Allow LAN access toggle when exit node is activated for all devices older than android 13 - API 33.
Instead of relying on android SDK, manually calculate required ranges.
Fixes https://github.com/tailscale/tailscale/issues/16187

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/App.kt
+++ b/android/src/main/java/com/tailscale/ipn/App.kt
@@ -314,6 +314,8 @@ class App : UninitializedApp(), libtailscale.AppContext, ViewModelStoreOwner {
 
   override fun getOSVersion(): String = Build.VERSION.RELEASE
 
+  override fun getSDKInt(): Long = Build.VERSION.SDK_INT.toLong()
+
   override fun isChromeOS(): Boolean {
     return packageManager.hasSystemFeature("android.hardware.type.pc")
   }
--- a/android/src/main/java/com/tailscale/ipn/ui/view/ExitNodePicker.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/view/ExitNodePicker.kt
@@ -100,9 +100,7 @@ fun ExitNodePicker(
           }
         }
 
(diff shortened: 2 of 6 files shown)


Pawloland:
@kari-ts what do you say?

libtailscale/ranges_calc.go
```diff
@@ -0,0 +1,253 @@
...
+		cur = cur.Add(cur, new(big.Int).Lsh(one, maxSize))
+	}
+
+	return result
+}
+
+// ---------- CIDR -> range ----------
+// prefixToRange converts a netip.Prefix to an ipRange with Start and End addresses.
+// Start is the network address and End is the broadcast address.
+func (s space) prefixToRange(p netip.Prefix) ipRange {
```
kari-ts: `prefixToRange` should probably call p = `p.Masked()` first. Right now it assumes `p.Addr() `is already the network address, which may not always be true for a `netip.Prefix.`

kari-ts:
Thanks! This mostly looks good. 

Since we squash anyway, please squash your branch into a single commit and include the issue reference in the commit message so issuebot passes.

Pawloland:
Should be good now

kari-ts (approved):

PR #781, opened 2026-04-20
Makefile: add docker-tailscale-debug target with persistent debug keystore

bradfitz:
The existing docker-run-build target is geared at release builds and
requires JKS signing env vars. There was no ergonomic path for the
common "iterate on a debug APK inside the build container" workflow.
Worse, every docker run produces an APK signed with a fresh ephemeral
debug keystore, so `adb install -r` refuses to update a prior install.

Add a docker-tailscale-debug target that builds the debug APK inside
the container, and bind-mount $(CURDIR)/.android-docker (gitignored)
at /root/.android in every docker-* target so the Gradle-generated
debug.keystore survives across invocations. Signer stays stable across
builds in the same checkout.

Note: mount target is /root/.android rather than the Dockerfile's
HOME=/build because the JVM's user.home resolves from /etc/passwd
(i.e. /root for uid 0), not from the HOME env var.

Made while working on tailscale/tailscale-android#695

Updates #13174
Updates tailscale/tailscale-android#695

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,10 @@ tailscale.jks
 # android sdk dir
 ./android-sdk
 
+# Persistent $HOME/.android for `make docker-*` (keeps debug.keystore so
+# the debug signer is stable across container runs).
+.android-docker
+
 # Java profiling output
 *.hprof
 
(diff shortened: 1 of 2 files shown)


barnstar (approved):
Thanks.  This tripped me up last week too.  I just gave it a BS key and password and it at least builds but this is much nicer.

PR #787, opened 2026-05-05
android: coalesce peer routes to skip rebuild storm on large tailnets

### LINKED ISSUE #19591: Android client rebuilds tun0 every 1-10s on large tailnet (~hundreds of peers), breaking SSH/TCP sessions
## Summary

On Android with a large corporate tailnet (~hundreds of peers + many subnet routers), the Tailscale Android client repeatedly calls `VpnService.Builder.establish()`, rebuilding `tun0` every 1-10 seconds. Each rebuild tears down all TCP connections through the tunnel, making SSH (and any long-lived TCP) unusable.

**Tailscale Bug Report ID** (already submitted from the app): `BUG-ae0ab5fe6a06c5a6a0f9f87f17ba4544702d591697a391b6d002cc1ceb03aa50-20260430101537Z-2ba0e11b7528edd9`

## Environment

- **Device**: Google Pixel 10 Pro Fold
- **OS**: Android 16
- **Tailscale Android**: 1.96.4-t8cf541dfd-g980f12083 (latest in Play Store as of 2026-04-30)
- **Tailscale (peer Mac, for reference)**: 1.96.5-t4ee448d3a-g74ffbefc2 (`io.tailscale.ipn.macsys`)
- **Tailnet**: large corporate (~hundreds of users + many tagged-devices acting as subnet routers advertising internal `10.x.x.x` /15-/16 ranges)
- **Network conditions**: stable home Wi-Fi, no roaming, no captive portal. Underlying Wi-Fi is not flapping (verified via `dumpsys wifi` and connectivity logs).

## Symptoms

- Status-bar VPN-key icon visibly flickers (appears/disappears every few seconds).
- TCP connections through `tun0` are torn down every few seconds.
- SSH from phone (Termius) to Mac peer (`100.100.x.x`) succeeds at the auth layer (server-side `Accepted publickey` log line confirmed via `log show` on the Mac), but the Termius client immediately reports `software caused connection abort` because `tun0` was rebuilt and the underlying socket died.
- Brief HTTP fetches to a Mac-side test listener occasionally succeed during the short windows when `tun0` is alive.

## Repro

1. Phone joins a large corporate tailnet. Verify peer count and subnet-router count is high (hundreds of peers; many advertised subnets).
2. Tailscale establishes `tun0`.
3. `adb shell logcat | grep "D/Vpn"` shows continuous state cycling.

## Evidence — VPN agent state from logcat

```
04-30 12:13:24.967 D/Vpn  setting state=DISCONNECTED, reason=agentDisconnect
04-30 12:13:25.282 D/Vpn  setting state=CONNECTING,   reason=establish
04-30 12:13:26.247 D/Vpn  setting state=CONNECTING,   reason=agentConnect
04-30 12:13:26.267 D/Vpn  setting state=CONNECTED,    reason=agentConnect
04-30 12:13:26.452 D/Vpn  setting state=DISCONNECTED, reason=agentDisconnect   # 185ms uptime
04-30 12:13:26.729 D/Vpn  setting state=CONNECTING,   reason=establish
04-30 12:13:27.688 D/Vpn  setting state=CONNECTING,   reason=agentConnect
04-30 12:13:27.702 D/Vpn  setting state=CONNECTED,    reason=agentConnect
04-30 12:13:29.758 D/Vpn  setting state=DISCONNECTED, reason=agentDisconnect   # 2s uptime
04-30 12:13:30.041 D/Vpn  setting state=CONNECTING,   reason=establish
04-30 12:13:31.017 D/Vpn  setting state=CONNECTED,    reason=agentConnect
04-30 12:13:39.064 D/Vpn  setting state=DISCONNECTED                           # 8s uptime
04-30 12:13:39.428 D/Vpn  setting state=CONNECTING,   reason=establish
04-30 12:13:40.428 D/Vpn  setting state=CONNECTED,    reason=agentConnect
04-30 12:13:50.982 D/Vpn  setting state=DISCONNECTED                           # 10s uptime
04-30 12:13:51.287 D/Vpn  setting state=CONNECTING,   reason=establish
04-30 12:13:52.296 D/Vpn  setting state=CONNECTED,    reason=agentConnect
```

The `reason=establish` lines indicate the client itself is calling `Builder.establish()` to rebuild — these are not OS-driven disconnects.

## Evidence — Route push size on each rebuild

Each `ConnectivityService.registerNetworkAgent` call from Tailscale pushes a very large route list:

- ~80 IPv4 subnet routes (advertised internal `10.x.x.x` ranges from corporate subnet routers)
- Hundreds of per-peer `/32` routes (`100.68.x.x/32`, `100.64.x.x/32`)
- IPv6 equivalents

`ip route show table all | grep -c 'dev tun0'` returned **588** initially.

After toggling "Use Tailscale subnets" OFF in the app, route count rose to **12052** across all routing tables — because Tailscale removed the broad `100.64.0.0/14` aggregate route and replaced it with explicit `100.64.x.x/32` per-peer routes for every tailnet member. This worth flagging as a separate concern: with subnets off, the route list still inflates linearly with peer count, the opposite of the user's intent in toggling subnets off.

## Mitigations attempted (none successful)

- `com.tailscale.ipn` is whitelisted in `dumpsys deviceidle whitelist` (battery optimization off) — flap continues
- "Use Tailscale subnets" toggled off — flap continues, route count actually grows
- Tailscale FGS is system allowlisted (`tempAllowListReason: SYSTEM_ALLOW_LISTED`) — process is not being killed
- No app crashes or ANRs in logcat for `com.tailscale.ipn`

## Hypothesis

`VpnService.Builder.establish()` on Android requires re-applying the entire interface config (routes, DNS, MTU) on any change — Android VPN API doesn't support incremental route updates. When netmap updates arrive frequently (normal in a tailnet of this size where peers come/go), the Android client appears to call `establish()` for each update. Each rebuild kills every TCP socket on `tun0`.

We suspect:
1. **No batching/debouncing of netmap-driven `establish()` calls** in the Android client. Multiple netmap updates within a short window should coalesce into a single rebuild.
2. **Per-peer /32 route inflation** when subnets are off. Routes that fall within the existing `100.64.0.0/10` Tailscale CGNAT space probably don't need explicit /32 entries.

## Impact

For users on large corporate tailnets, the Android client is effectively unusable for any long-lived TCP workload (SSH, persistent HTTP, mosh handshake, anything not built for second-order reconnect logic). The client appears connected (`tun0` exists, IP assigned) but nothing through it lives more than a few seconds.

## Asks

1. Batch/debounce netmap-driven `Builder.establish()` calls — coalesce updates into a single rebuild per N seconds.
2. Avoid pushing per-peer `/32` routes that are already covered by the broader `100.64.0.0/10` CGNAT space.
3. Consider exposing a metric / debug log for "rebuilds-per-minute" so users can self-diagnose this class of issue.

Happy to provide more logcat, the full BUG-ID bundle is already submitted. Reach out if you need adb shell access to repro steps.

ZebanNikolay:
## Summary

On a tailnet with hundreds of peers and `subnets-off`, `wgengine` emits a `router.Config` update on every peer up/down — typically every few seconds on busy tailnets — but the only field that changes is one peer's `/32` route inside Tailscale's CGNAT range (`100.64.0.0/10`). Each update flowed through `ReconfigureVPN` → `SetBoth` → `updateTUN`, calling `VpnService.Builder.establish()` and tearing down every TCP socket on `tun0`.

This PR coalesces per-peer `/32` (and IPv6 `/128`) routes into the well-known parent prefixes (`tsaddr.CGNATRange()` and `tsaddr.TailscaleULARange()`) before passing them to `VpnService.Builder`. Per-peer routing is already handled inside `wgengine`, so the parent prefix is sufficient at the Android VPN layer.

With routes coalesced, we snapshot the `VpnService.Builder`-relevant subset of the config (`vpnEffectiveCfg`) and short-circuit `updateTUN` when nothing in that subset changed. Peer churn that was previously triggering a tun rebuild every few seconds is now a no-op.

Side effect on a ~6600-peer tailnet: `AddRoute` calls per rebuild drop from ~6600 to ~10, and rebuilds become rare.

Fixes tailscale/tailscale#19591.

## Validation

Tested on a Pixel 10 Pro Fold (Android 16, API 36+) attached to a tailnet with ~6600 peers and subnets disabled.

**Before this patch** — `D/Vpn  setting state=CONNECTING, reason=establish` every 2–10 seconds; SSH/TCP sessions through the VPN broken.

**After this patch** — over a 2.5-minute observation window: 20 `router.Config` updates received (route count flapping between 6654 and 6665 due to peer up/down), all 20 hit the early-return path:

```
updateTUN: VPN-effective cfg unchanged; skipping rebuild (routes 6665, coalesced 56)
updateTUN: VPN-effective cfg unchanged; skipping rebuild (routes 6662, coalesced 56)
updateTUN: VPN-effective cfg unchanged; skipping rebuild (routes 6663, coalesced 56)
...
```

Zero `agentDisconnect` / `establish` events. SSH sessions through the VPN remain stable.

## Test plan

- [x] Builds locally (`make tailscale-debug`)
- [x] Sideloaded onto Pixel 10 Pro Fold; tailscale connects normally, exit nodes / subnets / DNS work as before
- [x] Verified rebuild storm is suppressed under sustained peer churn on a real ~6600-peer tailnet (logs above)
- [x] Unit tests for `routesutil.CoalescePeerRoutes` — added in a new `libtailscale/routesutil` sibling package so they run under `make go-test` (the `libtailscale` package itself is excluded because it needs the Android NDK to link). Covers passthrough, single/many peer collapse, mixed v4+v6, subnet routers + exit nodes, dedup when the parent prefix is already in input, and a regression test asserting peer-up/peer-down does not perturb the coalesced output.

## Notes for reviewers

- The CGNAT and Tailscale ULA ranges come from `tailscale.com/net/tsaddr` (`CGNATRange()`, `TailscaleULARange()`), not hardcoded.
- Subnet routers, exit nodes (`0.0.0.0/0`, `::/0`), and any advertised routes outside the CGNAT/ULA ranges pass through `CoalescePeerRoutes` unchanged.
- The change does not affect peer reachability — `wgengine` still receives the full per-peer route list internally; only `VpnService.Builder` sees the coalesced view.
- The early-return at the top of `updateTUN` precedes `b.CloseTUNs()` and the deferred `b.devices.Down()`, so the existing tun (and `multiTUN`) stay up and the `#18679` deadlock invariant is preserved.
- The new `vpnEffectiveCfg` gate co-exists with the existing `VPNFacade.Set`/`SetDNS` `Equal` short-circuits in `vpnfacade.go`. The latter filters most no-ops at the `Set` boundary; this new gate catches the case where `wgengine` sees a "real" `Set`-level change (e.g. one peer /32 added/removed) but the `VpnService.Builder`-relevant subset is unchanged. Happy to consolidate the two gates if reviewers prefer.

### DIFF
--- a/libtailscale/backend.go
+++ b/libtailscale/backend.go
@@ -101,7 +101,12 @@ type backend struct {
 	settings   settingsFunc
 	lastCfg    *router.Config
 	lastDNSCfg *dns.OSConfig
-	netMon     *netmon.Monitor
+	// lastEffective is the snapshot of the VpnService.Builder-relevant
+	// subset of the last-applied config. updateTUN uses it to skip a
+	// rebuild when only fields outside that subset (e.g. peer /32 churn
+	// that we coalesce) have changed. See tailscale/tailscale#19591.
+	lastEffective vpnEffectiveCfg
+	netMon        *netmon.Monitor
 
 	logIDPublic logid.PublicID
 	logger      *logtail.Logger
--- a/libtailscale/net.go
+++ b/libtailscale/net.go
@@ -9,11 +9,13 @@ import (
 	"log"
 	"net/netip"
 	"runtime/debug"
+	"slices"
 	"strings"
 	"syscall"
 
 	"github.com/tailscale/tailscale-android/libtailscale/ifaceparse"
 	rangescalc "github.com/tailscale/tailscale-android/libtailscale/ranges_calc"
+	routesutil "github.com/tailscale/tailscale-android/libtailscale/routesutil"
 	"github.com/tailscale/wireguard-go/tun"
 	"tailscale.com/net/dns"
 	"tailscale.com/net/netmon"
@@ -77,6 +79,20 @@ var googleDNSServers = []netip.Addr{
 }
 
 func (b *backend) updateTUN(rcfg *router.Config, dcfg *dns.OSConfig) (err error) {
+	// On large tailnets, wgengine fires a router.Config update on every peer
+	// up/down — but the change is just a per-peer /32 inside the tailnet's
+	// CGNAT range, which the Android VPN layer doesn't need (peer-level
+	// routing is handled inside wgengine). Coalescing those /32s into a
+	// single parent prefix lets us no-op when only peer membership changed
+	// and skip the VpnService.Builder.establish() rebuild that was tearing
+	// down every TCP socket on tun0. See tailscale/tailscale#19591.
+	eff := computeEffective(rcfg, dcfg)
+	if rcfg != nil && len(rcfg.LocalAddrs) > 0 && b.lastEffective.Equal(eff) {
+		b.logger.Logf("updateTUN: VPN-effective cfg unchanged; skipping rebuild (routes %d, coalesced %d)",
+			len(rcfg.Routes), len(eff.routes))
+		return nil
+	}
+
 	b.logger.Logf("updateTUN: changed")
 	defer b.logger.Logf("updateTUN: finished")
 
@@ -144,7 +160,8 @@ func (b *backend) updateTUN(rcfg *router.Config, dcfg *dns.OSConfig) (err error)
 
 	if useExclude {
 		// For API 33+, use ExcludeRoute for LocalRoutes and AddRoute for Routes.
-		for _, route := range rcfg.Routes {
+		// eff.routes is the coalesced set — see computeEffective.
+		for _, route := range eff.routes {
 			// Normalize route address; Builder.addRoute does not accept non-zero masked bits.
 			route = route.Masked()
 			if err := builder.AddRoute(route.Addr().String(), int32(route.Bits())); err != nil {
@@ -163,10 +180,10 @@ func (b *backend) updateTUN(rcfg *router.Config, dcfg *dns.OSConfig) (err error)
 			}
 		}
 
-		b.logger.Logf("updateTUN: added %d routes (exclude-mode), localRoutes=%d", len(rcfg.Routes), len(rcfg.LocalRoutes))
+		b.logger.Logf("updateTUN: added %d routes (exclude-mode), localRoutes=%d", len(eff.routes), len(rcfg.LocalRoutes))
 	} else {
 		// Older APIs: compute allowed-minus-disallowed prefixes and AddRoute them.
-		prefixesV4, prefixesV6, err := rangescalc.Calculate(rcfg.Routes, rcfg.LocalRoutes)
+		prefixesV4, prefixesV6, err := rangescalc.Calculate(eff.routes, rcfg.LocalRoutes)
 		if err != nil {
 			b.logger.Logf("updateTUN: route calculation error: %v", err)
 			return err
@@ -253,6 +270,7 @@ func (b *backend) updateTUN(rcfg *router.Config, dcfg *dns.OSConfig) (err error)
 
 	b.lastCfg = rcfg
 	b.lastDNSCfg = dcfg
+	b.lastEffective = eff
 	return nil
 }
 
@@ -269,6 +287,7 @@ func closeFileDescriptor() error {
 // CloseVPN closes any active TUN devices.
 func (b *backend) CloseTUNs() {
 	b.lastCfg = nil
+	b.lastEffective = vpnEffectiveCfg{}
 	b.devices.Shutdown()
 }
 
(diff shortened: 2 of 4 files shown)


ZebanNikolay:
Closing this — superseded by tailscale/tailscale#19652, which enables `OneCGNATRoute` on Android automatically (matching the macOS behavior). That fix lives in core (`ipn/ipnlocal`) rather than in the Android client, which is a strictly better landing spot than this patch and resolves tailscale/tailscale#19591 for every Android user, not just those who configure their tailnet for it.

For anyone hitting the rebuild storm before the upstream fix ships in a `tailscale-android` release: setting the `one-cgnat?v=true` node attribute on the tailnet ACL applies the same behavior to existing clients with no app update needed. We confirmed on our tailnet that the storm disappeared on the unmodified F-Droid build immediately after the ACL change.

Thanks @kari-ts for the quick turnaround.

PR #789, opened 2026-05-12
android: fix health warnings sort order

kari-ts:
This changes the comparator to sort by severity first (descending), ImpactsConnectivity as a secondary sort, and title as a final tiebreaker. This fixes the issue where the comparator was sorting ascending, and matches the new WinUI sorting behavior.

Fixes tailscale/corp#41733

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/model/Health.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/model/Health.kt
@@ -36,12 +36,19 @@ class Health {
 
     override fun compareTo(other: UnhealthyState): Int {
       // Compare by severity first
-      val severityComparison = Severity.compareTo(other.Severity)
+      val severityComparison = other.Severity.compareTo(other.Severity)
       if (severityComparison != 0) {
         return severityComparison
       }
 
-      // If severities are equal, compare by warnableCode
+      // If severities are equal, connectivity impacting issues first
+      val connectivityComparison =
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tailscale/ipn/ui/model/Health.kt
```diff
@@ -36,13 +36,20 @@ class Health {
...
-      return WarnableCode.compareTo(other.WarnableCode)
+      // If severities are equal, connectivity impacting issues first
+      val connectivityComparison =
+          (other.ImpactsConnectivity ?: false).compareTo(ImpactsConnectivity ?: false)
+      if (connectivityComparison != 0) {
+        return connectivityComparison
+      }
+
+      // Otherwise, alphabetical by title
+      return Title.compareTo(other.Title)
```
nickkhyl: I'm not sure whether this should be `Title` or `WarnableCode`.

I think `WarnableCode` makes a bit more sense, since the `Title` can theoretically change during the issue's lifecycle.

That said, in practice, we don't currently do that. So I guess either is fine?

nickkhyl (approved):
Thanks!!

android/src/main/java/com/tailscale/ipn/ui/model/Health.kt
kari-ts: Yeah, I wasn't sure either. Let's go with WarnableCode

PR #790, opened 2026-05-13
libtailscale: do not close old TUNs during reconfig

nickkhyl:
[2 commits to preserve a bit more context around these changes; see commit messages for details]

 - `libtailscale`: do not close old TUNs during reconfig, except on ChromeOS 
 - `libtailscale`: do not close old TUNs during reconfig on ChromeOS

Updates tailscale/tailscale#19591

### DIFF
--- a/libtailscale/net.go
+++ b/libtailscale/net.go
@@ -80,36 +80,41 @@ func (b *backend) updateTUN(rcfg *router.Config, dcfg *dns.OSConfig) (err error)
 	b.logger.Logf("updateTUN: changed")
 	defer b.logger.Logf("updateTUN: finished")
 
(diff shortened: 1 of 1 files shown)


nickkhyl:
I tested this manually on both Android and ChromeOS -- seems fine.

But I'd give it a bit more time on a few devices before merging, mainly because the `multiTUN` implementation as a whole is a bit fragile.

/cc @croakerbcts -- this change might be a good candidate for more scrupulous testing in 1.99.x / 1.100 when it lands.

kari-ts:
Awesome!! Thanks for the fix!

kari-ts (approved):

PR #796, opened 2026-05-26
Fix search crash on API 32 and older

Pawloland:
It should fix https://github.com/tailscale/corp/issues/27375, but I don't have access to it, so I didn't mention it in the commit message. This commit basically fixes a crash that prevented the search function from working on older Android versions, like Android 11.

```
16:35:00.336  E  FATAL EXCEPTION: main
Process: com.tailscale.ipn, PID: 22578
java.lang.NoClassDefFoundError: Failed resolution of: Landroid/window/OnBackInvokedCallback
```


Pawloland:
@kari-ts I hate to be that guy, but is there any hope of merging it?

Pawloland:
@bradfitz Also fixed merge conflicts, so it can also be merged

bradfitz (approved):

kari-ts:
Thanks for the fix! Sorry for the delay in reviewing. Could you please rebase onto current main and keep just the compatibility fix instead of merging this as is? It looks like there is merge conflict resolution that is no longer relevant (eg bringing in older `SearchView` implementation). Thanks again for helping us improve `tailscale-android`

PR #797, opened 2026-05-26
android: stop watching legacy netmap notifications

bradfitz:
Follow up to tailscale/tailscale#19607, which introduced the
constant-time node add/remove path and temporarily kept Android on
legacy Notify.NetMap emission.

Move the Android app notification watcher to InitialStatus plus
peer-change notifications, and opt both Android watchers out of
runtime NetMap delivery.

Updates tailscale/tailscale#12542


android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -114,15 +120,103 @@ object Notifier {
...
+        AllowedIPs = allowedIPs,
+        Hostinfo =
+            Tailcfg.Hostinfo(
+                OS = OS, Hostname = HostName, ShareeNode = ShareeNode, Location = Location),
+        LastSeen = LastSeen,
+        Online = Online,
+        Capabilities = Capabilities,
+        CapMap = CapMap,
+        ComputedName = computedName,
+        ComputedNameWithHost = computedName)
```
kari-ts: It looks like we're not using `ComputedNameWithHost` today, though it's something we could do. Could you make this mirror `InitDisplayName` and show `computedName (HostName) `https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L615 ?

kari-ts (approved):

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
bradfitz: done

bradfitz:
just fixed some problems around account switching back & forth between accounts where the previous version was retaining nodes in the display list (not in Go in wireguard) from the old account. seems to all work in the latest version now

bradfitz:
(merging per Slack chat w/ @kari-ts)

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -114,15 +143,218 @@ object Notifier {
...
+      }
+      return
+    }
+
+    if (next == null) {
+      return
+    }
+
+    val self = notify.SelfChange
+    if (initial == null && self != null && !next.isSameSelf(self)) {
```
kari-ts, line 183: do we need to do `userProfiles.clear()` here too?

bradfitz: https://github.com/tailscale/tailscale-android/pull/800

PR #798, opened 2026-05-27
android: add emulator integration test

bradfitz:
Updates tailscale/tailscale#13038


bradfitz:
Note the new magic label `run-integration-test` to make it run the VM-based integration tests. Let's see how long it takes to run in GitHub Actions 😬

bradfitz:
(I haven't yet worked on optimizing the test/layer caching here at all. Step 1 is just getting Hello Test working)

bradfitz:
15 minutes! Not as bad as I'd feared.

integration/androidvmtest/testdata/NetprobeReceiver.java
```diff
@@ -0,0 +1,74 @@
...
+    new Thread(new ProbeRunnable(context.getApplicationContext(), pending, path, url)).start();
+  }
+
+  private static final class ProbeRunnable implements Runnable {
+    private final Context context;
+    private final PendingResult pending;
+    private final String path;
+    private final String url;
+
+    ProbeRunnable(Context context, PendingResult pending, String path, String url) {
```
barnstar: Let's keep new things Kotlin.  I assume that's possible here.

bradfitz: Done. PTAL

barnstar (approved):
Doesn't run on arm/macOS for me - but we can fix that separately.

bradfitz:
Hmm, interesting. Yeah we can fix that. But my main goal was CI here.

PR #805, opened 2026-06-05
build.gradle: update versionCode to match playstore

kari-ts:
updates #cleanup

Manual bump to fix Fdroid

PR #806, opened 2026-06-05
build.gradle: update versionCode to match playstore (#805)

croakerbcts:
updates #cleanup

Manual bump to fix Fdroid


bradfitz:
We should have issue to automate+test this, instead of doing it manually with a `#cleanup`. (http://go/cleanup)

kari-ts (approved):

PR #807, opened 2026-06-05
android: derive versionCode from tailscale version

kari-ts:
Previously, versionCode was only set by the Play Store builder and never committed back to the repo, so fdroid wouldn't detect updates.
Now, we compute versionCode deterministically using the formula major*100M + minor*100K + patch*10 + platform, so that fdroid and Play Store builds always produce the same value.

Updates https://github.com/tailscale/tailscale/issues/18969

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ $(RELEASE_AAB): version gradle-dependencies
 
 $(RELEASE_TV_AAB): version gradle-dependencies
 	@echo "Building TV release AAB"
-	(cd android && ./gradlew test bundleRelease_tv)
+	(cd android && ./gradlew test bundleTvRelease -PPLATFORM=1) 
 	install -C ./android/build/outputs/bundle/release_tv/android-release_tv.aab $@
 
 tailscale-test.apk: version gradle-dependencies
@@ -278,10 +278,6 @@ tag_release: debug-symbols tailscale.version ## Tag the current commit with the 
 bumposs: update-oss tailscale.version
 	source tailscale.version && git commit -sm "android: bump OSS" -m "OSS and Version updated to $${VERSION_LONG}" go.toolchain.rev android/build.gradle go.mod go.sum
 	source tailscale.version && git tag -a "$${VERSION_LONG}" -m "OSS and Version updated to $${VERSION_LONG}"
-
-.PHONY: bump_version_code ## Bump the version code in build.gradle
-bump_version_code:
-	sed -i'.bak' "s/versionCode .*/versionCode $$(expr $$(awk '/versionCode ([0-9]+)/{print $$2}' android/build.gradle) + 1)/" android/build.gradle && rm android/build.gradle.bak
 
 .PHONY: update-oss ## Update the tailscale.com go module
 update-oss:
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -37,7 +37,7 @@ android {
     defaultConfig {
         minSdkVersion 26
         targetSdkVersion 35
-        versionCode 615
+        versionCode computeVersionCode()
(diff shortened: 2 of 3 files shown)


SuggonM:
Out of curiosity, is this the reason Tailscale hasn't been updated on F-Droid in 3 months?

Makefile
```diff
@@ -270,13 +270,16 @@ androidpath:
...
 
-.PHONY: tag_release
-tag_release: debug-symbols tailscale.version ## Tag the current commit with the current version
-	source tailscale.version && git tag -a "$${VERSION_LONG}" -m "OSS and Version updated to $${VERSION_LONG}"
-
-.PHONY: bumposs ## Bump to the latest oss and update the versions.
-bumposs: update-oss tailscale.version
-	source tailscale.version && git commit -sm "android: bump OSS" -m "OSS and Version updated to $${VERSION_LONG}" go.toolchain.rev android/build.gradle go.mod go.sum
+## Bump versionCode, amend the current commit, and tag it
+tag_release: debug-symbols tailscale.version bump_version_code
```
barnstar: Just bumping here will cause f-droid and playstore versions to diverge.. I don't know if we *care* since the version source-of-truth is the tailscale version, but Android cares and this makes things hard to reason about.

The build scripts will check the playstore version and increment this during the build.  Perhaps a better approach is to do something deterministic so that an f-droid build always ends up with same version as the corresponding playstore build and that the playstore version can be derived from the tailscale version and vice-versa.  That way, we're not changing anything at build time (except perhaps a modifier to denote atv)

Something like: major * 10000000 + minor * 1000000 + patch *10 + (platform)
1.99.123 -> 1_099_0123_0 and 1.100.123 - > 1_100_0123_0 and add 1 for the atv build and remove the check in the build pipeline that looks up the version and increments it and just change this on very bumposs or tagrelease.  With enough precision to allow for minor version up to 999, and patch versions  up to 9999....

barnstar (changes requested):

barnstar (approved):
Let's not merge until we get the builder updated... It appears that it would still work the "old way" but safety first...

sfllaw:
@barnstar, @kari-ts : It looks like go/build still needs this?

https://builds.corp.ts.net/logtail/2026-07-02T22-48-33-07-00-unstable-main-sfllaw@tailscale.com.log
```
make: *** No rule to make target 'bump_version_code'.  Stop.
make: *** [Makefile:381: docker-run-build] Error 2
android: 2026/07/03 06:02:06 Build failed: exit status 2
```

Related: https://github.com/tailscale/tailscale-android/pull/464

linsui:
F-Droid can't get auto update with this setup. It can be updated manually though. For auto update, F-Droid needs to be able to find the version code with regex.

barnstar:
@linsui Ack.  We'll modify this so that we write down the version explicitly in build.gradle when we change the OSS version.

linsui:
If you don't want to bump the version code manually, you can also release a text file for every version, in github release or your website.

kari-ts:
Thanks @linsui for the suggestion! Working on implementing this now.

kari-ts:
> If you don't want to bump the version code manually, you can also release a text file for every version, in github release or your website.

@linsui we are working on implementing this, and running into failures here: https://gitlab.com/tailscale-group/fdroiddata/-/jobs/16017264011. iiuc, the failure happens when go is verifying dependencies against `sum.golang.org`? I retried this a few times and saw failures on different modules (`docker-credential-helpers`, `prometheus`, and `chromedp/sysutil`), which makes me think there's some issue between the fdroid build environment and `sum.golang.org`?

linsui:
I can't see your pipelines. Your repo is not public. I never see such error though.

PR #809, opened 2026-06-13
android: add Spanish translations

oscfdezdz:


bradfitz:
Thanks for this, but we're still currently figuring out our policy (internal discussion: https://github.com/tailscale/corp/issues/43407) for what we do with translations to our various apps & websites.

We probably won't be able to merge this for some time until we figure out our overall translation plan.

PR #810, opened 2026-06-17
android: add launcher shortcuts to connect and disconnect the VPN

brettjenkins:
## Summary

Adds two static launcher shortcuts — **Connect** and **Disconnect** — backed by a transparent, no-UI `AutomationActivity` trampoline on the existing `CONNECT_VPN` / `DISCONNECT_VPN` intents. The actions can also be invoked from Samsung Modes & Routines and Tasker (Send Intent → Activity).

## Why an activity rather than the existing broadcast

I wrote the original `IPNReceiver` broadcast support, and it worked reliably when it landed. Over successive Android releases the background-execution and power-saving restrictions have tightened (Doze, app standby, and OEM battery managers like Samsung's deep sleep), and the broadcast path has become unreliable at waking an app that's been stopped or slept — it often can't start the foreground VPN service from that state.

Starting an *activity* avoids this: it wakes the app and lets connect start the foreground VPN service. The existing broadcast actions are unchanged.

## On-device validation

On a Galaxy S25 (SM-S948B, Android 16), with Tailscale put into deep sleep (hibernated + background-execution denied + restricted standby bucket + force-stopped) and triggered by **real Tasker intents**:
- Broadcast → `IPNReceiver`: **failed to connect**
- Activity → `AutomationActivity`: **connected** — both screen-on and screen-off

## Notes for reviewers

- **Exported without a permission** is deliberate: the same CONNECT/DISCONNECT actions are already reachable unguarded via the existing exported `IPNReceiver`. This adds an activity entry point for the reliable foreground-wake path, not new capability.
- **First run / not set up:** if the user isn't logged in or VPN consent hasn't been granted, the shortcut opens the app (which owns login and the consent prompt) rather than one-tap connecting.
- **Exit-node automation is unchanged** and remains available via the existing `IPNReceiver` `USE_EXIT_NODE` broadcast once connected.

## Related issues

Implements the launcher-shortcut / automation requests in tailscale/tailscale#9531, tailscale/tailscale#9497, tailscale/tailscale#16415, tailscale/tailscale#17855, and tailscale/tailscale#17738, and improves the broadcast-intent reliability reported in tailscale/tailscale#10831, tailscale/tailscale#14148, tailscale/tailscale#13623, and tailscale/tailscale#18847.

## Also fixes a cold-start crash

Found while testing the shortcuts against a force-stopped app. A localAPI call from a cold process (e.g. `UseExitNodeWorker`, started by `IPNReceiver`) reads `Request`'s `lateinit app` before the backend is initialized, so it crashes with `UninitializedPropertyAccessException`. `execute()` now resolves the app lazily via `App.get()`, the same way `Client` already does.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -57,6 +57,30 @@
             <intent-filter>
                 <action android:name="android.service.quicksettings.action.QS_TILE_PREFERENCES" />
             </intent-filter>
+            <!-- Static app shortcuts must be attached to the LAUNCHER activity. -->
+            <meta-data
+                android:name="android.app.shortcuts"
+                android:resource="@xml/shortcuts" />
+        </activity>
+
+        <!--
+            Transparent, no-UI trampoline that connects or disconnects the VPN. Exported so the
+            actions show up in the launcher long-press shortcut menu, Samsung Modes & Routines, and
+            can be sent from Tasker. Starting an activity (rather than IPNReceiver's broadcast)
+            reliably wakes a stopped app so connect can start the foreground VPN service.
+        -->
+        <activity
+            android:name="AutomationActivity"
+            android:exported="true"
+            android:excludeFromRecents="true"
+            android:noHistory="true"
+            android:taskAffinity=""
+            android:theme="@android:style/Theme.NoDisplay">
+            <intent-filter>
+                <action android:name="com.tailscale.ipn.CONNECT_VPN" />
+                <action android:name="com.tailscale.ipn.DISCONNECT_VPN" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
         </activity>
         <activity
             android:name="ShareActivity"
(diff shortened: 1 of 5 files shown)


surajverma:
Thanks @brettjenkins , hope they approve and merge it soon. This is much needed. Thanks again.

barnstar (approved):
@kari-ts LGTM but I'd like a second opinion.

GalenOfTheShadows:
I definitely want to +1 the urgency on this. I haven't looked too heavily at the code, but the workarounds to use the intent method causes some very irritating side-effects on all the devices I manage using this. As long as there isn't a security hole opened by this, it would be a much needed relief for administrators like myself.

jtruland:
+1 — would love to see this merged. Just looked into automating Tailscale connect/disconnect via Samsung Routines/Automate on a Galaxy S25 and ran into exactly the problem this PR describes. Native shortcut support like this would remove the need for third-party automation apps entirely — hope the second review comes through soon.

PR #814, opened 2026-06-26
android: support text and link sharing

### LINKED ISSUE #16850: Taildrop sharing should share text if stream is unavailable

willh-ts:
This enhances the Android share feature as follows:
- Text and URLs will be recognized and shared via the .txt and .url files
- Files, text, and URLs will now show a small banner atop the devices list, mirroring iOS behavior.
- If multiple taildrops arrive, tapping the banner will open a bottom sheet to take action on them individually.
- Files can be opened or the enclosing taildrop folder opened.
- Text can be copied.
- URLs can be opened in the default browser.
- "Recently Used" taildrop targets will be locally cached and presented atop the list. Presentation & max length matches iOS behavior.
- Simple polling for online status is added to avoid the need to reload the share sheet to find devices that connect while it is visible.
- Fixes a bug that can cause the share sheet to be non-functional if you share; don't exit the view; share again

The iOS and macOS share extensions are currently receiving minor
Taildrop enhancements to support sharing of Text and URLs from
the OS-level share menu. To allow older clients to seamlessly receive
shares from newer clients, the .txt and .url suffixes were chosen.

iOS & macOS (in https://github.com/tailscale/corp/pull/44001), Android
(with these changes), and soon Windows and Linux will recognize these
file types and respond accordingly upon receipt of them:
- .txt: Notification presented to "copy" the text to the clipboard.
- .url: Notification presented to "open" the link

For platforms that lack in-app notifications, the files will additionally
be saved to the Taildrop folder as "Text [timestamp].txt",
"URL [timestamp].inetloc", or "URL [timestamp].url". This addresses
a limitation for users who explicitly opt-out of app notifications. Those
users would otherwise receive no transferred content whatsoever.

Upon use or dismissal of URL shares, the .url file will be deleted. This
is done to avoid piling up generally unusable .url files on Android in the
Taildrop folder.

All other taildrop file transfer functionality remain unchanged.

updates https://github.com/tailscale/tailscale/issues/4896
updates https://github.com/tailscale/tailscale/issues/4996
fixes https://github.com/tailscale/tailscale/issues/16850

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/InlineShareActionReceiver.kt
+++ b/android/src/main/java/com/tailscale/ipn/InlineShareActionReceiver.kt
@@ -1,0 +1,66 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package com.tailscale.ipn
+
+import android.content.BroadcastReceiver
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.widget.Toast
+import com.tailscale.ipn.util.BrowserOpener
+import com.tailscale.ipn.util.InlineShare
+import com.tailscale.ipn.util.TSLog
+
+// Handles taps on Taildrop inline-share notifications: URL → browser, text → clipboard.
+class InlineShareActionReceiver : BroadcastReceiver() {
+  companion object {
+    private const val TAG = "InlineShareActionReceiver"
+    const val ACTION_CONSUME = "com.tailscale.ipn.INLINE_SHARE_CONSUME"
+    const val EXTRA_KIND = "kind"
+    const val EXTRA_CONTENT = "content"
+    const val EXTRA_ID = "id"
+  }
+
+  override fun onReceive(context: Context, intent: Intent) {
+    val kindRaw = intent.getStringExtra(EXTRA_KIND) ?: return
+    val content = intent.getStringExtra(EXTRA_CONTENT) ?: return
+    val id = intent.getStringExtra(EXTRA_ID)
+
+    val kind =
+        runCatching { InlineShare.Kind.valueOf(kindRaw.uppercase()) }.getOrNull()
+            ?: run {
+              TSLog.w(TAG, "unknown inline share kind: $kindRaw")
+              return
+            }
+
+    when (kind) {
+      InlineShare.Kind.URL -> openUrl(context, content)
+      InlineShare.Kind.TEXT -> copyToClipboard(context, content)
+    }
+
+    if (id != null) {
+      com.tailscale.ipn.ui.notifier.Notifier.removeInlineShare(id)
+    }
+  }
+
+  private fun openUrl(context: Context, content: String) {
+    val uri = runCatching { Uri.parse(content) }.getOrNull()?.takeIf { !it.scheme.isNullOrEmpty() }
+    if (uri == null) {
+      copyToClipboard(context, content)
+      return
+    }
+    if (!BrowserOpener.openInDefaultBrowser(context, uri)) {
+      TSLog.w(TAG, "failed to open URL $content")
+      copyToClipboard(context, content)
+    }
+  }
+
+  private fun copyToClipboard(context: Context, content: String) {
+    val cm = context.getSystemService(Context.CLIPBOARD_SERVICE) as? ClipboardManager ?: return
+    cm.setPrimaryClip(ClipData.newPlainText("Tailscale", content))
+    Toast.makeText(context, R.string.taildrop_copied_to_clipboard, Toast.LENGTH_SHORT).show()
+  }
+}
--- a/android/src/main/java/com/tailscale/ipn/InlineShareActionReceiver.kt
+++ b/android/src/main/java/com/tailscale/ipn/InlineShareActionReceiver.kt
@@ -1,0 +1,40 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package com.tailscale.ipn
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import com.tailscale.ipn.ui.notifier.Notifier
+import com.tailscale.ipn.ui.notifier.TaildropNotifier
+import com.tailscale.ipn.util.InlineShareConsumer
+
+// Handles taps on Taildrop inline-share notifications: URL → browser, text → clipboard.
+class InlineShareActionReceiver : BroadcastReceiver() {
+  companion object {
+    const val ACTION_CONSUME = "com.tailscale.ipn.INLINE_SHARE_CONSUME"
+    const val ACTION_DISMISS = "com.tailscale.ipn.INLINE_SHARE_DISMISS"
+    const val EXTRA_ID = "id"
+  }
+
+  // Looking the share up by id makes this idempotent: whichever of tap or swipe lands
+  // second finds no entry and does nothing.
+  override fun onReceive(context: Context, intent: Intent) {
+    val id = intent.getStringExtra(EXTRA_ID) ?: return
+    val share = Notifier.inlineShareInbox.value.find { it.id == id } ?: return
+    when (intent.action) {
+      ACTION_CONSUME -> {
+        // Leave the entry alone if nothing happened, so the user can retry.
+        if (!InlineShareConsumer.consume(context, share)) return
+        Notifier.removeInlineShare(id)
+        TaildropNotifier.cancel(context, id)
+      }
+      // Swiped away: the notification is already gone, so there's nothing to cancel.
+      ACTION_DISMISS -> {
+        InlineShareConsumer.dismiss(context, share)
+        Notifier.removeInlineShare(id)
+      }
+    }
+  }
+}
--- a/android/src/main/java/com/tailscale/ipn/ShareActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/ShareActivity.kt
@@ -8,6 +8,7 @@ import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.provider.OpenableColumns
+import android.util.Patterns
 import android.webkit.MimeTypeMap
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
@@ -30,6 +31,7 @@ import com.tailscale.ipn.ui.theme.AppTheme
 import com.tailscale.ipn.ui.util.set
 import com.tailscale.ipn.ui.util.universalFit
 import com.tailscale.ipn.ui.view.TaildropView
+import com.tailscale.ipn.util.InlineShare
 import com.tailscale.ipn.util.TSLog
 import kotlin.random.Random
 import kotlinx.coroutines.Dispatchers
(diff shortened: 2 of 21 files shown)


android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
kari-ts, line 1: `MainViewModel` has become pretty bloated. Could we move the new logic to its own self contained `PendingTaildropViewModel` or `TaildropInboxViewModel` or controller/service that `MainViewModel` delegates to?

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
```diff
@@ -242,6 +306,110 @@ class MainViewModel(private val appViewModel: AppViewModel) : IpnViewModel() {
...
+      is PendingTaildropItem.Payload -> {
+        when (item.payload.kind) {
+          TdPayload.Kind.URL -> openUrl(context, item.payload.content)
+          TdPayload.Kind.TEXT -> copyToClipboard(context, item.payload.content)
+        }
+        Notifier.removeTdPayload(item.payload.id)
+        TaildropNotifier.cancel(context, item.payload.id)
+      }
+    }
+    if (pendingItems.value.isEmpty()) isPresentingPendingItemsList.value = false
```
kari-ts: maybe make this react to changes to guard against acting on stale info?

android/src/main/java/com/tailscale/ipn/ShareActivity.kt
```diff
@@ -144,6 +150,47 @@ class ShareActivity : ComponentActivity() {
...
+  private fun tdPayloadFromIntent(intent: Intent): Ipn.OutgoingFile? {
+    val raw = intent.getCharSequenceExtra(Intent.EXTRA_TEXT)?.toString()?.trim().orEmpty()
+    if (raw.isEmpty()) return null
+
+    val text = stripChromeHighlightedTextShare(raw)
+
+    val title =
+        intent.getCharSequenceExtra(Intent.EXTRA_SUBJECT)?.toString()?.takeIf { it.isNotBlank() }
+
+    val kind =
```
kari-ts: maybe require `http`/`https` when classifying?

android/src/main/java/com/tailscale/ipn/ui/view/TaildropBannerView.kt
```diff
@@ -0,0 +1,161 @@
...
+              .padding(horizontal = 16.dp, vertical = 8.dp)
+              .offset { IntOffset(animatedOffset.toInt(), 0) }
+              .alpha(opacity)
+              .pointerInput(swipeDismissable, onlyItem?.id) {
+                if (!swipeDismissable || onlyItem == null) return@pointerInput
+                detectHorizontalDragGestures(
+                    onDragEnd = {
+                      dragOffsetPx =
+                          when {
+                            abs(dragOffsetPx) > 180f ->
```
kari-ts: since these are raw pixels, this could lead to strange behavior on some devices. maybe use something like `180.dp.toPx()`?

android/src/main/java/com/tailscale/ipn/util/BrowserOpener.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.net.Uri
+
+object BrowserOpener {
+  // Pin the ACTION_VIEW to the user's default browser package so non-browser
+  // apps that also claim http(s) (e.g. the Google app) don't trigger the chooser.
+  fun openInDefaultBrowser(context: Context, uri: Uri): Boolean {
```
kari-ts, line 17: should we restrict to `http`/`https` uris?

android/src/main/java/com/tailscale/ipn/util/TaildropUsageTracker.kt
```diff
@@ -0,0 +1,78 @@
...
+    return sorted.subList(0, pivot).toList() to sorted.subList(pivot, sorted.size).toList()
+  }
+
+  fun sortPeers(
+      context: Context,
+      userID: Long,
+      peers: List<Tailcfg.Node>,
+  ): List<Tailcfg.Node> =
+      peers.sortedWith { a, b ->
+        val ta = getLastUsed(context, userID, a.StableID)
```
kari-ts: can we load the per-user map once instead of calling `getLastUsed` inside the comparator? right now sorting repeatedly reloads/decodes prefs

android/src/main/java/com/tailscale/ipn/ShareActivity.kt
willh-ts: I'd like to keep this permissive for any valid protocol. Mac and ([hopefully soon](https://github.com/tailscale/corp/pull/42388)) iOS, for example, support `tailscale:` deep-link navigation. e.g., "share a link to open `device or service XYZ` details", "open `Settings > Accounts`", etc.

Users could legitimately want to Taildrop `tel:`, `sms:`, `mailto:`, `ftp:`, `ssh:` or other app-defined links between their devices, and the Android share sheet hands us all of those. The classifier just decides between "URL" and "text" presentation (tap to open vs. tap to copy). If the scheme isn't browsable, `BrowserOpener.openInDefaultBrowser` will fail and we should fall back to the text / clipboard copy path. Restricting here would force valid links down the text path and lose the "tap to open" affordance.

android/src/main/java/com/tailscale/ipn/ui/viewModel/MainViewModel.kt
willh-ts: I relocated the new feature code into a separate PendingTaildropViewModel.

android/src/main/java/com/tailscale/ipn/util/BrowserOpener.kt
willh-ts: I updated slightly to let this handle both http/https and any supported protocol: 

Tested, functional
- http:, https:, tel:, sms:, mailto:.

Tested, not yet supported, falls through to "copy" handler on tap:
- `tailscale:`

Tested, fails & falls through to "copy" handler:
- `invalid:`

The latter 2 could be improved slightly (i.e., we could show "copy" instead of "open" for the subtitle) if we used PackageManager.resolveActivity to preemptively check whether a URL can be handled, but as I understand it, older Android versions and some OEM flavors restrict access to that.

Screenshot_20260629_120306

android/src/main/java/com/tailscale/ipn/ui/notifier/TaildropNotifier.kt
```diff
@@ -0,0 +1,80 @@
...
+        body = pending.content.take(120).replace("\n", " ")
+        subtitle = context.getString(R.string.taildrop_tap_to_copy)
+      }
+    }
+
+    val tapIntent =
+        Intent(context, InlineShareActionReceiver::class.java).apply {
+          action = InlineShareActionReceiver.ACTION_CONSUME
+          putExtra(InlineShareActionReceiver.EXTRA_KIND, pending.kind.name.lowercase())
+          putExtra(InlineShareActionReceiver.EXTRA_CONTENT, pending.content)
```
kari-ts: what if this hits the size limit? https://developer.android.com/guide/components/activities/parcelables-and-bundles#sdbp looks like this is 1MB

android/src/main/java/com/tailscale/ipn/util/ShareFileHelper.kt
```diff
@@ -120,8 +123,26 @@ object ShareFileHelper : libtailscale.ShareFileHelper {
...
+
   @Throws(IOException::class)
   override fun openFileWriter(fileName: String, offset: Long): libtailscale.OutputStream {
+    if (InlineShare.matchesAnyStage(fileName)) {
+      val ctx = appContext ?: throw IOException("App context not initialized")
+      val cached =
+          inlineShareCacheFiles.computeIfAbsent(fileName) {
+            File(inlineShareCacheRoot(ctx), fileName)
+          }
+      val fos = FileOutputStream(cached, offset != 0L)
```
kari-ts: I think this means append, not write starting at offset. for example, see https://github.com/tailscale/tailscale-android/blob/0867f01687a3955f7c0b5c6c62b236b997d68601/android/src/main/java/com/tailscale/ipn/util/ShareFileHelper.kt#L116-L117

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
```diff
@@ -57,6 +59,17 @@ object Notifier {
   val incomingFiles: StateFlow<List<Ipn.PartialFile>?> = MutableStateFlow(null)
   val filesWaiting: StateFlow<Empty.Message?> = MutableStateFlow(null)
 
+  private val _inlineShareInbox = MutableStateFlow<List<PendingInlineShare>>(emptyList())
```
kari-ts: if the user hasn't granted notification permissions doesn't the content disappear if the app is killed?

android/src/main/java/com/tailscale/ipn/ShareActivity.kt
kari-ts: thanks for the explanation!!

android/src/main/java/com/tailscale/ipn/util/ShareFileHelper.kt
willh-ts: Updated & retested a couple files from iOS to android.

android/src/main/java/com/tailscale/ipn/ui/notifier/TaildropNotifier.kt
willh-ts: Good catch, I wasn't familiar with that. Updated to send and ref by ID.

android/src/main/java/com/tailscale/ipn/ui/notifier/Notifier.kt
willh-ts: Another good catch. I added a cache object that survives the process being killed. Tested with these steps & I'll check if we'll need to adjust the iOS side accordingly:

- launch
- iOS sends a URL to Android
- Leave banner untouched
- `adb shell am force-stop com.tailscale.ipn`
- relaunch, banner reappears

android/src/main/java/com/tailscale/ipn/ui/viewModel/PendingTaildropViewModel.kt
```diff
@@ -0,0 +1,190 @@
...
+      is PendingTaildropItem.File -> {
+        openFile(context, item.partial)
+        consumedFileIds.update { it + item.id }
+      }
+      is PendingTaildropItem.InlineShareItem -> {
+        when (item.share.kind) {
+          InlineShare.Kind.URL -> openUrl(context, item.share.content)
+          InlineShare.Kind.TEXT -> copyToClipboard(context, item.share.content)
+        }
+        Notifier.removeInlineShare(item.share.id)
```
kari-ts, line 107: it looks like `copyToClipboard` returns if there is a failure. should we still call `removeInlineShare` in this case?

android/src/main/java/com/tailscale/ipn/util/InlineShareInboxStore.kt
```diff
@@ -0,0 +1,42 @@
...
+// field initializer, which runs synchronously inside App's own initOnce()/initializeApp(),
+// before isInitialized is set. Calling App.get() here would re-enter initOnce() and start
+// Libtailscale a second time.
+object InlineShareInboxStore {
+  private const val PREF_KEY_INBOX = "inline_share_inbox"
+  private val json = Json { ignoreUnknownKeys = true }
+
+  private fun encryptedPrefs() = (UninitializedApp.get() as App).getEncryptedPrefs()
+
+  fun save(inbox: List<PendingInlineShare>) {
```
kari-ts, line 23: do we have a size bound on `inbox` before persisting? I wonder if a large taildrop or a bunch of pending saves might make this synchronous save problematic (cause lags, performance issues)

android/src/main/java/com/tailscale/ipn/InlineShareActionReceiver.kt
```diff
@@ -0,0 +1,64 @@
...
+              TSLog.w(TAG, "unknown inline share kind: $kindRaw")
+              return
+            }
+
+    when (kind) {
+      InlineShare.Kind.URL -> openUrl(context, content)
+      InlineShare.Kind.TEXT -> copyToClipboard(context, content)
+    }
+
+    Notifier.removeInlineShare(id)
```
kari-ts: same here with `removeInlineShare` being called even if `copyToClipboard` early returns with failure

android/src/main/java/com/tailscale/ipn/util/ShareFileHelper.kt
```diff
@@ -207,14 +240,83 @@ object ShareFileHelper : libtailscale.ShareFileHelper {
...
+        Notifier.appendInlineShare(pending)
+        TaildropNotifier.notify(ctx, pending)
+      } else {
+        TSLog.w("ShareFileHelper", "consumeInlineShare: decode failed for $targetName")
+      }
+    } else {
+      TSLog.w("ShareFileHelper", "consumeInlineShare: no bytes for $targetName")
+    }
+
+    cachedPartial.delete()
```
kari-ts: looks like this deletes regardless of success/failure

kari-ts (commented):
thanks for bearing with me!

android/src/main/java/com/tailscale/ipn/ui/viewModel/PendingTaildropViewModel.kt
willh-ts: I'll make `copyToClipboard` return Boolean and gate removeInlineShare +
TaildropNotifier.cancel on it. Same for `openUrl`, since it falls back to the clipboard.
Also wrapping `setPrimaryClip` in runCatching. It can throw out of consume().

android/src/main/java/com/tailscale/ipn/util/InlineShareInboxStore.kt
willh-ts: No bound today, afaik, and thinking about why led somewhere bigger.

I'd been assuming this path only sees modest text selections and URLs, since that's what the share sheet produces. But routing is by filename, not origin, so anything named `md5.txt` lands here. Someone taildropping a hash-named log dump would hit this path and get it silently converted into a clipboard snippet instead of a saved file. That's a real (if specific) use case we'd be breaking and it's a better argument than any size cap. So rather than bound the ephemeral path, I've removed it. Details below.

1. Share sheet text/link sharing is unchanged, here and in the open iOS PR.
2. `<md5>.txt` and `<md5>.url` save to the Taildrop folder like any other file. This does block text/link sharing until a folder is picked, but that's consistent with every other share and it deletes a lot of code.
3. `.url` files are deleted on either "open link" or dismiss. Android has no handler for the format and I legitimately can't think of a reason to keep one lingering on your phone.
4. `.txt` files are never deleted. Copy is offered when the file is small enough to be worth reading back; otherwise it's just a file. Anyone already sending themselves `<md5>.txt` log dumps is unaffected.

android/src/main/java/com/tailscale/ipn/util/ShareFileHelper.kt
willh-ts: You were right, and the fix was to stop keeping a separate copy at all. The cache staging path is gone & as mentioned above, these now take the normal SAF path, with copy/open layered on top as a best-effort nicety.

The upshot for this thread: the failure paths don't delete anything now. Unreadable, too large, or undecodable all just leave a normal file in the Taildrop folder.

android/src/main/java/com/tailscale/ipn/util/InlineShareInboxStore.kt
kari-ts: hmmm this is probably even more niche but if we get a `.url` file and the user dismisses the banner then the file gets deleted right? probably obscure enough that it is very unlikely to happen but just wanted to clarify

kari-ts (approved):

android/src/main/java/com/tailscale/ipn/util/InlineShareInboxStore.kt
willh-ts: Yes, in the latest revision, that dismiss action removes the .url file & users would need to reshare it. On mobile, this seems like a reasonable trade off to avoid a pile of unusable .url files. 

iOS actually does support .webloc, but even there, it seems awkward to persist those files. I updated its PR to behave the same as Android does here.

PR #815, opened 2026-06-29
android: extend support for protocol navigation to Android

willh-ts:
Extends support for the `navigate` path of the `tailscale:` protocol to Android,
mirroring the existing macOS routes & pending iOS routes.

All settings tabs without existing routes simply open the main settings view
for now.

```
tailscale://navigate/main/devices
tailscale://navigate/main/devices/<computedName or stableID>
tailscale://navigate/main/exit-nodes
tailscale://navigate/main/exit-nodes/<stableID>
tailscale://navigate/main/exit-nodes/location/<country>
tailscale://navigate/settings[/<anyTab>]
```

The `scripts/deeplink-probe.go` script can be used to trigger the application to
navigate to a given route by passing that to it.
e.g., `./tool/go run ./scripts/deeplink-probe.go main/devices`

updates tailscale/corp#41056

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/MainActivity.kt
+++ b/android/src/main/java/com/tailscale/ipn/MainActivity.kt
@@ -62,6 +62,7 @@ import com.tailscale.ipn.ui.model.Ipn
 import com.tailscale.ipn.ui.notifier.Notifier
 import com.tailscale.ipn.ui.theme.AppTheme
 import com.tailscale.ipn.ui.util.AndroidTVUtil
+import com.tailscale.ipn.ui.util.DeepLinkNavigator
 import com.tailscale.ipn.ui.util.set
 import com.tailscale.ipn.ui.util.universalFit
 import com.tailscale.ipn.ui.view.AboutView
@@ -114,6 +115,8 @@ class MainActivity : ComponentActivity() {
   private lateinit var vpnPermissionLauncher: ActivityResultLauncher<Intent>
   private lateinit var appViewModel: AppViewModel
   private lateinit var viewModel: MainViewModel
+  private var deepLinkNavigator: DeepLinkNavigator? = null
+  private var pendingDeepLink: Uri? = null
 
   val permissionsViewModel: PermissionsViewModel by viewModels()
 
@@ -135,6 +138,8 @@ class MainActivity : ComponentActivity() {
   @SuppressLint("SourceLockedOrientationActivity")
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+
+    pendingDeepLink = deepLinkUri(intent)
 
     // grab app to make sure it initializes
     App.get()
@@ -252,6 +257,20 @@ class MainActivity : ComponentActivity() {
       }
 
       navController = rememberNavController()
+
+      val introNotYetViewed = remember { isIntroScreenViewedSet() }
+
+      LaunchedEffect(navController) {
+        val nav = DeepLinkNavigator(navController)
+        deepLinkNavigator = nav
+        pendingDeepLink?.let {
+          pendingDeepLink = null
+          when {
+            introNotYetViewed -> TSLog.d(TAG, "Deep link dropped, intro not viewed: $it")
+            !nav.handle(it) -> TSLog.d(TAG, "Deep link handler returned false for $it")
+          }
+        }
+      }
 
       AppTheme {
         Surface(color = MaterialTheme.colorScheme.inverseSurface) { // Background for the letterbox
@@ -483,6 +502,23 @@ class MainActivity : ComponentActivity() {
         }
       }
     }
+    deepLinkUri(intent)?.let { uri ->
+      val nav = deepLinkNavigator
+      if (nav != null) {
+        if (!nav.handle(uri)) {
+          TSLog.d(TAG, "Deep link handler returned false for $uri")
+        }
+      } else {
+        pendingDeepLink = uri
+      }
+    }
+  }
+
+  private fun deepLinkUri(intent: Intent?): Uri? {
+    if (intent?.action != Intent.ACTION_VIEW) return null
+    val uri = intent.data ?: return null
+    if (uri.scheme != "tailscale" || uri.host != "navigate") return null
+    return uri
   }
 
   private fun login(urlString: String) {
--- a/android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
@@ -1,0 +1,105 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package com.tailscale.ipn.ui.util
+
+import android.net.Uri
+import androidx.navigation.NavHostController
+import com.tailscale.ipn.ui.model.Tailcfg
+import com.tailscale.ipn.ui.notifier.Notifier
+import com.tailscale.ipn.util.TSLog
+
+// Deep links must only navigate — never perform an action on the user's
+// behalf. Tails whose leaf would imply an action (e.g., a specific exit
+// node) are accepted for URL compatibility but resolve to the parent
+// view; the user picks and confirms in-app.
+class DeepLinkNavigator(private val navController: NavHostController) {
+  companion object {
+    private const val TAG = "DeepLinkNavigator"
+  }
+
+  fun handle(uri: Uri): Boolean {
+    if (uri.host != "navigate") return false
+    val segments = uri.pathSegments
+    val window = segments.firstOrNull() ?: return false
+    val tail = segments.drop(1)
+
+    return when (window) {
+      "main" -> handleMain(tail)
+      "settings" -> {
+        showSettings()
+        true
+      }
+      else -> false
+    }
+  }
+
+  private fun handleMain(tail: List<String>): Boolean {
+    val tab = tail.firstOrNull() ?: return false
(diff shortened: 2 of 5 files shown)


android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,153 @@
...
+import android.net.Uri
+import androidx.navigation.NavHostController
+import com.tailscale.ipn.ui.localapi.Client
+import com.tailscale.ipn.ui.model.Ipn
+import com.tailscale.ipn.ui.model.Tailcfg
+import com.tailscale.ipn.ui.notifier.Notifier
+import com.tailscale.ipn.util.TSLog
+import kotlinx.coroutines.CoroutineScope
+
+// URL shape (mirrors iOS / macOS):
```
barnstar: Let's link this to our documentation instead.

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,153 @@
...
+    val node = findNode(identifier)
+    if (node == null) {
+      TSLog.d(TAG, "Deep link: device not found for '$identifier'")
+      return
+    }
+    popToMain()
+    navController.navigate("peerDetails/${node.StableID}")
+  }
+
+  // Matches macOS: ComputedName (case-insensitive) first, then StableID.
```
barnstar: Remove the macOS ref... Let's just document the scheme (somewhere).

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,153 @@
...
+              pushDeviceDetail(rest[0])
+              true
+            }
+            else -> false
+          }
+      "exit-nodes" ->
+          when {
+            rest.isEmpty() -> {
+              presentExitNodePicker()
+              true
```
barnstar, line 55: If we do this elsewhere, I think we shouldn't.   It's one thing to present the UI - it's another thing to allow changing something like an exit node via a deeplink.

scripts/deeplink-probe.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
```
barnstar: nit:  go, not bash when we can...  Maybe toss this into an llm and have it port it.

barnstar (dismissed):
Mainly, I think we need to be really conservative on performing actions with deeplinks (and perhaps audit iOS/macOS if we allow that there).  Navigation is perfectly fine, but the idea that clicking a link can change what exit node I'm using has some implications we need to consider before rolling it out.

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
willh-ts: I submitted a docs request just now for Mac & iOS. I'll update it after this merges to layer in the Android routes.

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
willh-ts: Good call. I'll reduce any that take action to instead only show the root view. This differs slightly from desktop where you can view exit nodes w/o toggling them. Updating the iOS changes similarly.

scripts/deeplink-probe.sh
willh-ts: Converted

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,100 @@
...
+        ?: all.firstOrNull { it.StableID == identifier }
+  }
+
+  private fun presentExitNodePicker() {
+    popToMain()
+    navController.navigate("exitNodes")
+  }
+
+  private fun presentSettings() {
+    popToMain()
```
kari-ts: I think this won't work if main isn't in the backstack, eg if a settings deep link is handled on cold start

kari-ts: maybe something liek this? 
```
private fun presentSettings() {
  navController.navigate("settings") {
    popUpTo("main") { inclusive = false }
    launchSingleTop = true
  }
}
```
https://developer.android.com/guide/navigation/backstack#pop

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,100 @@
...
+    navController.navigate("settings")
+  }
+
+  private fun popToMain() {
+    navController.popBackStack(route = "main", inclusive = false)
```
kari-ts: maybe check the return value and log if this fails?

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,100 @@
...
+    return when (tab) {
+      "devices" ->
+          when (rest.size) {
+            0 -> {
+              showDeviceList()
+              true
+            }
+            1 -> {
+              pushDeviceDetail(rest[0])
+              true
```
kari-ts: this returns true even if the node is missing, is this intended?

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  private fun findNode(identifier: String): Tailcfg.Node? {
+    val netmap = Notifier.netmap.value ?: return null
+    val all = netmap.Peers.orEmpty() + netmap.SelfNode
+    return all.firstOrNull { identifier.equals(it.ComputedName, ignoreCase = true) }
+        ?: all.firstOrNull { it.StableID == identifier }
+  }
+
+  private fun presentExitNodePicker() {
```
kari-ts: nit: this sounds a little Apple-y. I think Android tends to use "show"

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -483,6 +497,21 @@ class MainActivity : ComponentActivity() {
...
+      val nav = deepLinkNavigator
+      if (nav != null) {
+        nav.handle(uri)
+      } else {
+        pendingDeepLink = uri
+      }
+    }
+  }
+
+  private fun navigateUri(intent: Intent?): Uri? {
```
kari-ts: maybe rename, since this doesn't actually navigate? this just extracts a uri from the intent, so maybe something liek `deepLinkUri` for clarity?

android/src/main/java/com/tailscale/ipn/MainActivity.kt
```diff
@@ -253,6 +258,15 @@ class MainActivity : ComponentActivity() {
 
       navController = rememberNavController()
 
+      LaunchedEffect(navController) {
+        val nav = DeepLinkNavigator(navController)
+        deepLinkNavigator = nav
+        pendingDeepLink?.let {
+          pendingDeepLink = null
+          nav.handle(it)
```
kari-ts: maybe check here and also log failure too?

android/src/main/java/com/tailscale/ipn/MainActivity.kt
kari-ts, line 425: how does/should this interact with deep linking? eg, if deep linking happens without having viewed the intro screen?

kari-ts (commented):
Thanks for doing this!

android/src/main/java/com/tailscale/ipn/MainActivity.kt
willh-ts: That's a good call out and I suspect we have the same issue on both iOS and mac. It could make sense in all three cases to gate the DeepLink router on our intro/onboarding properties, ignoring the navigation request if the intro screens haven't yet been marked as `seen`. I updated MainActivity's pendingDeepLink handler to do that here.

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
willh-ts: The docs request got denied, leaving this in a bit of a nebulous state. I suppose we can leave it as is for now and re-enumerate the routes at some point.

android/src/main/java/com/tailscale/ipn/ui/util/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,105 @@
...
+    navController.navigate(route) {
+      popUpTo("main") { inclusive = false }
+      launchSingleTop = true
+    }
+  }
+
+  private fun popToMain() {
+    val popped = navController.popBackStack(route = "main", inclusive = false)
+    if (!popped) {
+      TSLog.d(TAG, "Deep link: popBackStack to 'main' returned false")
```
kari-ts, line 102: I think we want something like 
```
if (navController.currentDestination?.route == "main") {
  return
}

if (!navController.popBackStack(route = "main", inclusive = false)) {
  TSLog.d(TAG, "Deep link: failed to pop back stack to 'main'")
}
```
so that we don't log if we're alreayd on main

kari-ts:
LGTM once conflicts are resolved

kari-ts (approved):

PR #818, opened 2026-07-06
android: add French (fr) translation

foXaCe:
## What

Adds a complete French (`fr`) localization of the Android app UI:
`android/src/main/res/values-fr/strings.xml` — 307 translatable strings,
mirroring the default `values/strings.xml`.

## Details

- **Register:** formal (vouvoiement), consistent throughout.
- **Terminology:** Tailscale product terms kept in English (Tailscale,
  tailnet, Taildrop, WireGuard, Mullvad, Headscale, DNS, VPN, MDM,
  split tunneling). UI concepts translated consistently, e.g.
  `exit node` → *nœud de sortie*, `admin console` → *console d'administration*,
  `auth key` → *clé d'authentification*, `subnet router` → *routeur de sous-réseau*,
  `Tailnet lock` → *verrou du tailnet*, `peer` → *pair*.
- **Formatting preserved:** all placeholders (`%s`, `%d`, `%1$s`, `%.1f`),
  XML entities (`&#160;`, `&amp;`), `\n`, and escaped apostrophes (`\'`).
- `translatable="false"` strings are intentionally omitted (Android convention).
- The trademark footnote (`about_view_footnotes`) is deliberately left in English.
- French typography applied (non-breaking spaces before `? : ;`, guillemets).

## Verification

- `xmllint` valid; EN↔FR key parity 307/307 (no missing, no orphan keys).
- Placeholder parity checked per key; terminology consistency checked
  (no EN string mapped to divergent FR values).
- No unescaped apostrophes (build-safe), UTF-8, no mojibake.

PR #819, opened 2026-07-06
android: externalize hardcoded UI strings

foXaCe:
## What

Externalizes user-facing strings that were hardcoded in Kotlin/Java into
`res/values/strings.xml`, so they can be localized and pass the
`HardcodedText` lint check.

## Changes

- **10 new string resources** (mostly accessibility `contentDescription`s):
  `MainView`, `TaildropView`, `PeerDetails`, `HealthView`, `LoginQRView`,
  `SharedViews`, and the split-tunnel switch confirmation dialog title.
- **Reuse existing resources** where they already exist:
  - search icons → `R.string.search` / `R.string.clear_search`
  - Quick Settings tile label → `R.string.tile_name`
- French translations for the new keys added to `res/values-fr/strings.xml`.

No behavior change — the English strings are byte-for-byte identical to before.

## Notes

- Stacked on #818 (French translation). This PR's diff also contains that
  commit; once #818 is merged/closed it will show only the extraction changes.
- `./gradlew ktfmtCheck` was not run in this environment (no network for the
  Gradle/ktfmt plugin download); the one wrapped line was formatted to match
  the surrounding ktfmt (Meta) style by hand.

## Verification

- No hardcoded UI strings remain (re-scan of Compose `Text`/`contentDescription`,
  `setLabel`/`setTitle`).
- `xmllint` valid; EN↔FR key parity 317/317.
- Imports added where needed (`com.tailscale.ipn.R` in `SharedViews.kt`).

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/QuickToggleService.java
+++ b/android/src/main/java/com/tailscale/ipn/QuickToggleService.java
@@ -30,7 +30,7 @@ public class QuickToggleService extends TileService {
         if (t == null) {
             return;
         }
-        t.setLabel("Tailscale");
+        t.setLabel(app.getString(R.string.tile_name));
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
             t.setSubtitle(act ? app.getString(R.string.connected) : app.getString(R.string.not_connected));
         }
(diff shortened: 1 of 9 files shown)


barnstar (approved):

barnstar (changes requested):
Let's remove the French translations from this and deal with those as a separate concern.  The other bits are good.

foXaCe:
Done — removed the French translations from this PR. It now only externalizes the hardcoded strings into `res/values/strings.xml` (plus reusing `R.string.search` / `R.string.clear_search` / `R.string.tile_name` where they already existed). Also rebased on latest `main`.

The translations are handled separately in #818.

willh-ts (approved):

PR #826, opened 2026-07-22
Makefile: bump API level to 36

willh-ts:
Sets gradle.properties for API level that is shared across various tooling steps.

updates tailscale/corp#40476


Makefile
barnstar: Bump this to today's date too to force the builder to create a fresh build image.

barnstar (approved):

PR #835, opened 2026-08-05
android: fix init order in HealthNotifier

barnstar:
fixes tailscale/corp#46191

The init order in in HealthNotifier is incorrect and can cause an immediate runtime NPE.


kari-ts (approved):
Thanks for the fix!

PR #837, opened 2026-08-10
.github: add bump-oss GH action

willh-ts:
Adds a manual-dispatch GitHub Action to run `make update-oss` and `make bump-version-code`. GITHUB_TOKEN usage prevents CI checks, so this additionally, preemptively runs make go-test before creating the update OSS PR.

Example run w/ the workflow temporarily set to pull_request trigger:

image

updates #cleanup

### DIFF
--- a/.github/workflows/bumposs.yml
+++ b/.github/workflows/bumposs.yml
@@ -1,0 +1,95 @@
+# Manual-dispatch "bump OSS + versionCode" action. 
+# `make update-oss` (pull tailscale.com@main) followed by
+# `make bump-version-code` (stamp android/build.gradle's versionCode from the
+# wall clock), then opens a PR.
+#
+# Uses the built-in GITHUB_TOKEN (no GitHub App required). Two consequences:
+#   1. Requires repo Settings -> Actions -> General ->
+#      "Allow GitHub Actions to create and approve pull requests" = enabled,
+#      or the create-pull-request step 403s.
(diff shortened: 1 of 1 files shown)


.github/workflows/bumposs.yml
```diff
@@ -0,0 +1,97 @@
+# Manual-dispatch "bump OSS + versionCode" action. 
+# `make update-oss` (pull tailscale.com@main) followed by
+# `make bump-version-code` (stamp android/build.gradle's versionCode from the
+# wall clock), then opens a PR.
+#
+# Uses the built-in GITHUB_TOKEN (no GitHub App required). Two consequences:
+#   1. Requires repo Settings -> Actions -> General ->
+#      "Allow GitHub Actions to create and approve pull requests" = enabled,
+#      or the create-pull-request step 403s.
+#   2. GitHub will NOT run push/pull_request workflows on a PR opened with
```
kari-ts: According to https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow, `pull_request` workflow runs in approval-required state are created for PRs opened with GITHUB_TOKEN

willh-ts: Good catch. I misunderstood the setup to mean all checks would be bypassed with this. It seems this PR will land in a state that will require us to click the button to manually run the checks. Updated this comment & removed the redundant go-test step.

willh-ts:
Closing due to the versionCode revert.

PR #839, opened 2026-08-15
android: force Client.app to initialize before executing local API requests

pinoybear:
### What this fixes

Fixes the `lateinit property app has not been initialized` crash at `Request.execute()`, most recently reported in tailscale/tailscale#20884, and previously in #14125 / #14314.

### Root cause

`Client` has an `app` property intended to force `App.get()` (and therefore `Request.setApp()`) to run before any local API call:

```kotlin
private val app: libtailscale.Application by lazy { App.get().getLibtailscaleApp() }
```

This property was added in #563 specifically to prevent this crash ("Lazily init app in Client to ensure that we aren't trying to make any local API calls before app has been initialized"). However, `by lazy` only evaluates on first *read* of the property, and nothing in `Client` ever reads `this.app` — none of `start()`, `status()`, `prefs()`, etc. reference it. So the property has never actually forced initialization since it was introduced; it's been dead code from the start (confirmed by checking every commit that has touched this file since #563 — none add or remove a reference to `app`).

If a `Client` is constructed and used before anything else has called `App.get()` — e.g. a WorkManager background task running on a cold-started process — `Request.execute()` reads the companion object's separate `lateinit var app` (set only via `Request.setApp()`, called from `App.startLibtailscale()`) before it's been set, and crashes.

### Fix

Add an `init` block that reads `app`, forcing the lazy initializer to run at `Client` construction time rather than never. `App.initOnce()` is `@Synchronized`, so this blocks the constructing thread until `App.get()` (and `Request.setApp()`) has fully completed, closing the race.

`Client.kt` is the only place `Request` is constructed in this codebase, so this covers every path that can reach this crash today.

This isn't a novel pattern — `Notifier.start()` already guards the equivalent case correctly:

```kotlin
if (!::app.isInitialized) {
  App.get()
}
```

This PR brings `Client` in line with that existing, correct approach used elsewhere in this codebase for the same class of problem.

### Verification

Static/manual verification only — I don't have a full Android build environment to run `make fmt-check` / `make tailscale-debug.apk` locally, so please treat CI as the source of truth for formatting and build correctness. Happy to address any CI failures.

Confirmed via `adb shell dumpsys dropbox` that this exact crash signature is currently reproducing on a shipped 1.102.2 build (Galaxy Z Fold8, and previously a Fold7 on a different Android major version), despite the crash being reported "resolved" after #563/#14125. Full crash details and device history in tailscale/tailscale#20884.

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/localapi/Client.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/localapi/Client.kt
@@ -71,6 +71,17 @@ class Client(private val scope: CoroutineScope) {
   // Access libtailscale.Application lazily
   private val app: libtailscale.Application by lazy { App.get().getLibtailscaleApp() }
 
+  init {
+    // Force the lazy `app` property to evaluate now, at construction time, instead of
+    // waiting for something to read it. This guarantees App.get() (and therefore
+    // Request.setApp()) has run before this Client can be used to execute a Request,
+    // even if the very first caller is a background WorkManager task on a cold-started
+    // process that never otherwise touches App.get(). Without this, `app` was declared
+    // but never read anywhere in this class, so it never actually forced initialization
+    // (see #563, which added this property with that intent but never wired it up).
+    app
+  }
+
   fun start(options: Ipn.Options, responseHandler: (Result<Unit>) -> Unit) {
     val body = Json.encodeToString(options).toByteArray()
     return post(Endpoint.START, body, responseHandler = responseHandler)


bradfitz:
> I don't have a full Android build environment

Please elaborate. Did you compile & test this change on your own device or not?

pinoybear:
Correct — I have not compiled or run this on a device. This was static analysis: full rename-aware commit history of `Client.kt` (via local `git log --follow`, not just the GitHub API, which under-reports by one commit), a full-repo grep confirming `Client.kt` is the only place `Request` is constructed, and a byte-for-byte diff against the exact commit that produced the shipped `1.102.2-...-g0b3c1bdb2` build to make sure I was reasoning about the actual code in question.

The fix itself is a one-line addition (`init { app }`) using a pattern that already exists and works correctly elsewhere in this codebase (`Notifier.start()`'s `if (!::app.isInitialized) { App.get() }` guard), so I'm confident in the mechanism, but I have not verified it builds or passes `make fmt-check` locally — happy to have CI be the judge, or if you'd rather I build and test it on-device first before this goes further, let me know and I'll do that.

pinoybear:
Following up with an actual on-device test, per your question above.

Built `tailscale-debug.apk` from this branch via the project's own Docker build (`make docker-build-image` + `make docker-tailscale-debug`), installed it on the same device that produced the original crash reports in tailscale/tailscale#20884 (Galaxy Z Fold8), uninstalling the Play Store release first since debug and release builds share `applicationId` but not a signing key.

```
versionName=1.103.90-t25877455e-gab045c406
firstInstallTime=2026-08-15 12:23:28
```
(the `g` hash matches this branch's fix commit)

```
$ adb shell dumpsys dropbox --print > dropbox_dump.txt
$ grep -c "lateinit property app" dropbox_dump.txt
0
$ grep -ic "tailscale" dropbox_dump.txt
0
$ head -1 dropbox_dump.txt
2026-08-14 18:20:09 ...
$ tail -1 dropbox_dump.txt
2026-08-17 09:27:25 ...
```
Buffer covers 2026-08-14 18:20 through 2026-08-17 09:27, comfortably spanning the full period since install through now (~45 continuous hours), including normal background use, doze cycles, and a device migration between two different host machines over USB. Zero occurrences of the crash signature.

For comparison, I went back through the original crash timestamps from #20884 and computed the actual gaps between consecutive crashes rather than eyeballing it: 12 crashes over ~2 days, with the longest gap between any two crashes being 21h41m — and even that gap ended right when the device rebooted, which is close to a worst-case trigger condition (guaranteed cold process start). The current clean window (~45h) is now more than double that longest pre-fix gap, including at least one device reboot and OS migration of its own.

pinoybear:
Update with a cleaner test: cherry-picked this fix onto `release-branch/1.102` (the actual shipping stable line, not `main`) and rebuilt, so this is closer to "1.102.3 plus just this one fix" rather than the mixed `main`-tip build from my last comment.

```
versionName=1.102.3-t9329c3677-g7004610d1-fix839
firstInstallTime=2026-08-30 18:09:23
```
(base version + commit hash match `release-branch/1.102` tip + this fix; suffix is just a local marker, not part of the actual patch)

```
$ adb shell dumpsys dropbox --print > dropbox_dump.txt
$ grep -c "lateinit property app" dropbox_dump.txt
0
$ grep -ic "tailscale" dropbox_dump.txt
0
$ head -1 dropbox_dump.txt
2026-09-08 09:43:07 ...
$ tail -1 dropbox_dump.txt
2026-09-11 08:40:40 ...
```

This build has been installed and running continuously on my phone for 11.5 days now, and I can confirm zero occurrences of the crash signature (or any other Tailscale crash) in the last 3-day window (2026-09-08 09:43 through 2026-09-11 08:40). I see release 1.102.4 has shipped since I've been running this patched build and the issue looks to remain after confirming the tag directly.

PR #832, opened 2026-08-03
android: fix LazyColumn crash from duplicate StableNodeIDs

fryorcraken:
Fixes tailscale/tailscale#20725

The Android app can crash with `IllegalArgumentException: Key "X" was already used` when a netmap contains duplicate or otherwise non-unique StableIDs and those IDs are used as LazyColumn item keys.

Changes:
- Deduplicate peers by StableID in `PeerCategorizer`, processing `SelfNode` first so the local node wins if a peer somehow shares its StableID.
- Deduplicate tailnet exit nodes by StableID in `ExitNodePickerViewModel`.
- Namespace `MainView` peer keys per user group as defense in depth.
- Add unit tests for `PeerCategorizer` duplicate StableID handling.

Tests: added `PeerCategorizerTest`; a local Android SDK was not available in this environment so full `./gradlew testDebugUnitTest` was not executed here.

PR #840, opened 2026-08-21
PeerHelper: deduplicate self node in peer list to fix LazyColumn crash

D-Fed11:
## Problem

The Android app crashes immediately on launch with headscale servers:

```
java.lang.IllegalArgumentException: Key "27" was already used. If you are using LazyColumn/Row please make sure you provide a unique key for each item.
```

## Root cause

`PeerHelper.regenerateGroupedPeers()` iterates over `(peers + selfNode)`. With headscale, `selfNode` is already present in `netmap.Peers`, so it gets added twice. Both entries share the same `StableID`, which becomes a duplicate key in `MainView`'s `LazyColumn`.

## Fix

`distinctBy { it.StableID }` on the combined list before iterating.

## Testing

Verified working on Raspberry Pi 4 running LineageOS TV, authenticated against a headscale server. Before fix: crash on every launch. After fix: UI loads and functions correctly.

## Refs

- Closes #20725
- Ref #832 (same fix, closed by author before merge)

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/ui/util/PeerHelper.kt
+++ b/android/src/main/java/com/tailscale/ipn/ui/util/PeerHelper.kt
@@ -32,7 +32,7 @@ class PeerCategorizer {
 
     val me = netmap.currentUserProfile()
 
-    for (peer in (peers + selfNode)) {
+    for (peer in (peers + selfNode).distinctBy { it.StableID }) {
 
       val userId = peer.User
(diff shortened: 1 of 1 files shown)


bradfitz:
`Closes #20725` ... that doesn't link anything. You mean `tailscale/tailscale#20725`: https://github.com/tailscale/tailscale/issues/20725

But why is Headscale doing that?

PR #849, opened 2026-09-02
android: include tailscale in split tunnel allowlist

### LINKED ISSUE #20978: Allowlist for split tunneling on Android
### What is the issue?

I was trying to get a new Android device added to my Tailnet and there were only five apps that I needed to grant Tailnet access to on the device. I noticed that there was a feature to change the split tunneling blocklist to an allowlist. When I did that and enabled the five apps that I wanted to grant Tailnet access, subnet routing was working fine, but I could not get DNS resolution for any local or internet addresses. I spent a few hours trying to figure out why it wasn't working and then I eventually decided to start from scratch and switch back to the blocklist from the allowlist. I blocked everything but the same five apps and everything began working normally.

Afterwards, I noticed your support doc states that the split tunneling allowlist requires an MDM on Android devices. If this _**is not**_ accurate, then the documentation needs to be updated and then maybe I should troubleshoot some more. If it _**is**_ accurate, then the TS app should add that bit to the warning message when Android users try to change to the allowlist.

### Steps to reproduce

_No response_

_No response_

### OS

Android

### OS version

17

### Tailscale version

1.98.8

### Other software

My network is running a pi-hole for DNS that my tailnet leverages indirectly via subnet routing.

### Bug report

_No response_

kari-ts:
This adds tailscale to the allowlist when split tunneling is configured in include mode. Without this, included apps lose access to tailscale DNS because the tailscale process is excluded from the vpn.

Tested manually with an included app against a dns record that is onlly resolvable through a tailnet dns servier, and verified the record resolves in allowlist mode and confirmed the query reaches the tailnet resolver over tailscale

Fixes tailscale/tailscale#20978

### DIFF
--- a/android/src/main/java/com/tailscale/ipn/IPNService.kt
+++ b/android/src/main/java/com/tailscale/ipn/IPNService.kt
@@ -209,15 +209,19 @@ open class IPNService : VpnService(), libtailscale.IPNService {
       TSLog.d(TAG, "Application packages were set by user: $packagesList")
     }
 
+    packagesList =
+        packagesForVpnBuilder(
+            packagesList = packagesList,
+            allowPackages = allowPackages,
+            tailscalePackageName = UninitializedApp.get().packageName,
+            builtInDisallowedPackages = UninitializedApp.get().builtInDisallowedPackageNames)
+
     if (allowPackages) {
       for (packageName in packagesList) {
         TSLog.d(TAG, "Including app: $packageName")
         allowApp(b, packageName)
       }
     } else {
-      // Make sure to also exclude hard-coded apps that are known to cause issues
-      packagesList += UninitializedApp.get().builtInDisallowedPackageNames
-
       for (packageName in packagesList) {
         TSLog.d(TAG, "Disallowing app: $packageName")
         disallowApp(b, packageName)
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tailscale/ipn/IPNService.kt
```diff
@@ -234,3 +238,15 @@ open class IPNService : VpnService(), libtailscale.IPNService {
...
+    packagesList: List<String>,
+    allowPackages: Boolean,
+    tailscalePackageName: String,
+    builtInDisallowedPackages: List<String>,
+): List<String> =
+    if (allowPackages) {
+      packagesList + tailscalePackageName
+    } else {
+      packagesList + builtInDisallowedPackages
+    }
```
willh-ts, line 252: It's an edge case, but if `allowPackages` is true, but `packagesList` is empty (however that manifests ... e.g., deselecting the last included app), this would set an otherwise empty allowlist (meaning no split-tunnel?) to a one-item allowlist. We might want a third condition for that before the one at L248.

Not sure if it matters to have duplicates, or whether to worry about potentially invalid MDM entries, but (packagesList + tailscalePackageName).distinct() could deduplicate.

barnstar (approved):

android/src/test/kotlin/com/tailcale/ipn/IPNServiceTest.kt
```diff
@@ -0,0 +1,33 @@
...
+  fun excludedPackagesIncludeBuiltInDisallowedPackages() {
+    val packages =
+        packagesForVpnBuilder(
+            packagesList = listOf("com.example.excluded"),
+            allowPackages = false,
+            tailscalePackageName = "com.tailscale.ipn",
+            builtInDisallowedPackages = listOf("com.example.builtin"))
+
+    assertEquals(listOf("com.example.excluded", "com.example.builtin"), packages)
+  }
```
willh-ts, line 32: nit: It might be worth a third test case for the empty-allowlist path & maybe another for deduplication.

android/src/main/java/com/tailscale/ipn/IPNService.kt
kari-ts: thanks!! will address in follow-up

PR #856, opened 2026-09-04
android: device pins/favorites and match iOS long push menu

hiimtmac:
Adds support for device pinning - from device list and from peer detail
Match iOS long push menu - ipv4/ipv6/magic DNS, update icon and reorder

Fixes https://github.com/tailscale/corp/issues/47581
Fixes https://github.com/tailscale/corp/issues/47582

https://github.com/user-attachments/assets/8fb70908-d52b-4a7b-a758-20ac97325eb7


android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
```diff
@@ -0,0 +1,136 @@
...
+) {
+  private val TAG = "FavoritesManager"
+
+  private val _favorites = MutableStateFlow<Favorites?>(null)
+  val favorites: StateFlow<Favorites?> = _favorites
+
+  private val _writing = MutableStateFlow(false)
+  val writing: StateFlow<Boolean> = _writing
+
+  private val userFlow = netmapFlow.mapNotNull { it?.User() }
```
kari-ts: pins are scoped per profile login, and iiuc since the active store is being tracked by `ipn.ProfileID` https://github.com/tailscale/tailscale/blob/023255e8a27ec9f6a21d24e3eda21c052ff72af3/feature/favorites/favorites.go#L68` we should use `LoginProfile.ID` here?

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
```diff
@@ -0,0 +1,136 @@
...
+      }
+    }
+  }
+
+  private fun send(request: FavoritesRequest) {
+    val snapshot = revert
+    val user = currentUser
+    revert = null
+    _writing.value = true
+    client.setFavorites(request) { result ->
```
kari-ts: could multiple `setFavorites` be in flight at the same time? is it possible that an older request might complete after a newer one and overwrite the newer state?

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
hiimtmac: Good call. It's now keyed on the profile ID. Did some other cleanup so lookup issued before a switch can't land after and install stale profile and logout clears pins.

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
hiimtmac: Yes also good catch. Fixed by only having one POST in flight. A write that fires while one is outstanding replaces pending and is sent after the current one settles. Older response landing on newer state isn't reachable now.

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+
+  private suspend fun write(request: FavoritesRequest) {
+    val snapshot = revert
+    val profile = currentProfile
+    revert = null
+    _writing.value = true
+
+    val result = suspendCancellableCoroutine { cont ->
```
barnstar, line 142: Tangential to this PR, but the client API functions not being suspend fun's is maybe something we should consider changing to avoid the use of continuations here.

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
```diff
@@ -0,0 +1,171 @@
...
+                  load(profile, isRetry = true)
+                }
+              }
+            }
+      }
+    }
+  }
+
+  private suspend fun write(request: FavoritesRequest) {
+    val snapshot = revert
```
barnstar, line 137: This is missing a try-catch in case the the suspendCancellableCoroutine is cancelled...  But it may make more sense to track the job and cancel that.   Or just wrap the contents of write() in a mutex.

```
private var writeJob: Job? = null

fun enqueueWrite(request: FavoritesRequest) {
    writeJob?.cancel() 
    writeJob = viewModelScope.launch {
        write(request)
    }
}

private suspend fun write(request: FavoritesRequest) {
    val snapshot = _favorites.value 
    val profile = currentProfile
    
    _writing.value = true

    try {
        val result = suspendCancellableCoroutine { cont -> 
            val call = client.setFavorites(request) { cont.resume(it) }
        }

        if (currentProfile != profile) return

        result.onFailure {
            TSLog.e(TAG, "Error writing favorites: ${it.message}")
            _favorites.value = snapshot
        }
    } finally {
        _writing.value = false 
    }
}
```

barnstar (changes requested):
Looks good - just one small issue with the write serialization.  I think we should also add the long press gesture the the search results if that's possible - though we can do that in a follow-up.

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
hiimtmac: Yeah this is something id like to do when I look into the openapi stuff for localapi/client 👍

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
hiimtmac: Good catch - updated 👍

android/src/main/java/com/tailscale/ipn/ui/notifier/FavoritesManager.kt
```diff
@@ -0,0 +1,174 @@
...
+    try {
+      val result = suspendCancellableCoroutine { cont ->
+        client.setFavorites(request) { cont.resume(it) }
+      }
+
+      if (currentProfile != profile) return
+      if (revert != null) return // newer burst opened while in flight
+      result.onFailure {
+        TSLog.e(TAG, "Error writing favorites: ${it.message}")
+        _favorites.value = snapshot
```
kari-ts, line 151: I think there could potentially be an issue with this line, because it's updating `_favorites` optimistically, but if you have two consecutive failed writes, then you might get something odd: 
toggle A -> `_favorites`: `[A]`, `snapshot`: `[ ]` 
toggle B while write is in flight -> `_favorites`: `[A, B]`, `revert`: `[A]`
the write fails and the rollback is sksipped because of the newer burst, then the second write fails, and we get `snapshot`: `[A]` so `favorites` is `[A]`

one approach would be to track the last successfully loaded/written favprites separately and use that for rollback

hiimtmac: Agreed, I changed rollback to use the last successfully loaded/written favourites and added some tests.

kari-ts:
thanks! this mostly looks good. I'm curious about what happens if a profile switch happens if there is an in-flight write. iiuc, the `LocalAPI` write doesn't get canceled. in a profile switch, could `e.store` be switched to that of the new profile before an older pins request under the previous profile is completed?

hiimtmac:
> thanks! this mostly looks good. I'm curious about what happens if a profile switch happens if there is an in-flight write. iiuc, the `LocalAPI` write doesn't get canceled. in a profile switch, could `e.store` be switched to that of the new profile before an older pins request under the previous profile is completed?

I couldn't find any other coordination between inflight LocalAPI writes and switching profiles in android. I agree that this is a valid concern but I think it should probably be addressed at the project level (rather than a custom implementation for favourites alone). AFAICT the iOS/macOS impl also lack the coordination so potentially this could be solved at the shared go/LocalAPI level instead? Id propose a project-wide follow up for shoring up this pattern everywhere it can be applied to - wdyt?

kari-ts:
> > thanks! this mostly looks good. I'm curious about what happens if a profile switch happens if there is an in-flight write. iiuc, the `LocalAPI` write doesn't get canceled. in a profile switch, could `e.store` be switched to that of the new profile before an older pins request under the previous profile is completed?
> 
> I couldn't find any other coordination between inflight LocalAPI writes and switching profiles in android. I agree that this is a valid concern but I think it should probably be addressed at the project level (rather than a custom implementation for favourites alone). AFAICT the iOS/macOS impl also lack the coordination so potentially this could be solved at the shared go/LocalAPI level instead? Id propose a project-wide follow up for shoring up this pattern everywhere it can be applied to - wdyt?

yep, I agree with this!

kari-ts (approved):

PR #857, opened 2026-09-04
android: update DNS when default network is changed

kari-ts:
When a new network is detected, we get onAvailable, onCapabilitiesChanged, onLinkPropertiesChanged, and these trigger recomputing and possible updating of the default network. Only onLinkPropertiesChanged updates the dns config, and if the network validation happens after onLinkPropertiesChanged fires, then the cached default network is still on a different network. This refreshes the dns config when a capabilities change causes the selected network to change.

Updates tailscale/corp#47631


cmol (approved):
Looks good on the surface, though I don't think I am super authoritative on android ATM.

barnstar (approved):

PR #859, opened 2026-09-08
android: update multiTUN for new tun.Device Read API and OSS bump

kari-ts:
This adapts multiTUN to the slab-based tun.Device.Read API introduced by https://github.com/tailscale/tailscale/commit/c21e854657a6896ec5ce32ace11b55ab014cb80b.

Updates https://github.com/tailscale/corp/issues/46716


libtailscale/multitun.go
```diff
@@ -58,10 +60,11 @@ type tunDevice struct {
 }
 
 type ioRequest struct {
```
jwhited: nit: a number of fields in this struct used to be shared between `tun.Device.Read` and `tun.Device.Write`, but now they are direction specific. 

Might be worth some comments and/or updated names. Or, separate read and write types, assuming that fits into the model used here.

jwhited (approved):
One nit/suggestion. Just needs a squash. LGTM.

PR #865, opened 2026-09-11
android: migrate agp to 9.4

hiimtmac:
Fixes tailscale/corp#48155

## What changed

- Updated AGP from 8.13.0 to 9.4.0
- Updated the Gradle wrapper to 9.6.0
- Updated the required Kotlin toolchain to 2.2.10
- Added the Kotlin Compose compiler plugin
- Migrated from the external Kotlin Android plugin to AGP built-in Kotlin
- Migrated to the public AGP DSL
- Replaced the deprecated Kotlin `jvmTarget` configuration
- Removed generated compatibility flags that were demonstrably irrelevant
- Retained behavior-sensitive compatibility settings
- Preserved `minSdk`, `targetSdk`, and Java 17 compilation

## Verification

- `./gradlew test`
- `./gradlew assembleDebug`
- `./gradlew assembleRelease`
- `./gradlew assembleApplicationTestAndroidTest`


Makefile
barnstar: Change the date stamp on this one too.  That will trigger the builder to generate a new image with the new toolchains.

barnstar (dismissed):

android/gradle.properties
```diff
@@ -4,3 +4,8 @@ android.nonFinalResIds=false
 android.nonTransitiveRClass=true
 android.useAndroidX=true
 org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m
```
kari-ts, line 6: Iiuc, these are added to preserve existing behavior. do we want to follow up and remove these so we eventually adopt agp 9 defaults wehre possible?

hiimtmac: Yep thats correct. These preserve the previous behaviour where defaults changed in AGP 9. I think we should follow up and remove them where possible to adopt the AGP 9 default values. The followups should validate the flags independently especially the two R8 settings against a minified release build. R8 changes are scary 😅

kari-ts: agreed, sounds good to me!

kari-ts (approved):

PR #873, opened 2026-09-16
android: fix ktfmt for agp 9.4

hiimtmac:
Fixes tailscale/corp#48331

Upgrade the ktfmt Gradle plugin from 0.17.0 to 0.27.0 for compatibility with AGP 9.4. With 0.17.0, `ktfmtCheck` had no actions because the plugin did not discover AGP 9.4’s Kotlin source. Version 0.27.0 correctly registers and runs the source-set-specific ktfmt tasks.

The plugin upgrade also moves the bundled ktfmt engine from 0.47 to 0.64, so this PR includes the resulting one-time formatting changes. CI continues to run `ktfmtCheck` across all Kotlin source sets (not just what is changed in the PR).


barnstar (approved):
@hiimtmac Does ktfmt have rules like SwiftLint where we might be able to override some if it's more colourful choices like not liking single line if statements or trailing brackets?  At some point, those were acceptable formatting to it. I dislike that this touched everything (and I didn't look carefully - we use ktfmt so I will defer to its new choices).   

If we have to to, we might as well rip off the bandaid.    I didn't check every change - but if it still compiles - merge it.

hiimtmac:
> @hiimtmac Does ktfmt have rules like SwiftLint where we might be able to override some if it's more colourful choices like not liking single line if statements or trailing brackets? At some point, those were acceptable formatting to it. I dislike that this touched everything (and I didn't look carefully - we use ktfmt so I will defer to its new choices).
> 
> If we have to to, we might as well rip off the bandaid. I didn't check every change - but if it still compiles - merge it.

`ktfmt` has a few broad options, but nothing like SwiftLint, and unfortunately many choices are controlled by the formatter. This diff comes from upgrading the ktfmt plugin 0.17.0->0.27.0 (for AGP comp). And this brings along a formatter change for ktfmt 0.47->0.64. So I think we are out of luck for trailing brackets/single line statements unfortunately 👎.

PR #875, opened 2026-09-16
android: notify users before device key expiry

kari-ts:
We were surfacing expiry warnings in-app when a netmap update caused the warning state to be recalculated. This change schedules a system notification using WorkManager, and reschedules when the expiry or configured warning window changes.

Updates tailscale/tailscale#20989


hiimtmac (approved):
Should an expiry change defensively cancel any notification that has already been posted? I think this handles canceling the "work to schedule a notification" but if you had a notification already posted, did the re-auth, the notification could still be visible even though its no longer relevant? (I could be wrong about this tho!)

kari-ts:
> Should an expiry change defensively cancel any notification that has already been posted? I think this handles canceling the "work to schedule a notification" but if you had a notification already posted, did the re-auth, the notification could still be visible even though its no longer relevant? (I could be wrong about this tho!)

yes, good catch! added a defensive cancelation

hiimtmac (approved):

